PR #71, opened 2020-12-10
Auth test

Qlexio:
Issue #51
# Testing existing code

* Add function to settings.py to test captcha
* Change path redirection in authentication.views.py to make it work
* Testing login and redirections
  * Test addition of new user with new mail
  * On database with existing users, test addition of user with:
    * different mail of existing users
    * same mail of existing users
  * Add 2 users with same mail and test result
* Testing token sent by mail and good redirections with:
  * Valid token
  * Invalid token
  * Already used token
  * Timed out token

### DIFF
--- a/transport_nantes/authentication/tests.py
+++ b/transport_nantes/authentication/tests.py
@@ -7,6 +7,7 @@ import authentication.views as views
 from asso_tn.utils import make_timed_token
 from django.utils.crypto import get_random_string
 from time import sleep
+from captcha.models import CaptchaStore
 
 # Create your tests here.
 
@@ -23,10 +24,12 @@ class UserTest(TestCase):
         site.name = "localhost"
         site.save()
 
+        # Set captcha
+        captcha = CaptchaStore.objects.get(hashkey=CaptchaStore.generate_key())
         # Set parameters to similate POST
         context = { "email": "test1@truc.com",
-                     'captcha_0': "captcha",
-                     'captcha_1': "PASSED" }
+                     'captcha_0': captcha.hashkey,
+                     'captcha_1': captcha.response }
         request = RequestFactory().post(reverse("authentication:login"), context, follow=True)
 
         # Call .views.login()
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
PR #71 and #72 refer to the same branch.  This will be overly confusing.

https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-requests

transport_nantes/authentication/tests.py
```diff
@@ -1,3 +1,163 @@
...
+
+# TODO Test good printing of pages (first print, post, redirect) and also page needing token from mail
+# TODO Test adding of new user if account doesn't exist and if account already exists
+class UserTest(TestCase):
+
+    # Add new user and test site page shown
+    def test_add_new_user(self):
+
+        # Set site into Site table
+        site = Site.objects.get(id=1)
```
JeffAbrahamson, line 22: Where was this set?

JeffAbrahamson:
At commit b1df174 : 
* `python manage.py test` fails: the site record is not staged to the testing db.
* The change to ` transport_nantes/transport_nantes/settings.py` isn't appropriate -- all development should not be subject to that constraint, only auth tests.  This should be set in the test's `setup()` or `setUpTestData()` methods.

JeffAbrahamson:
At 6024f71:
* `python manage.py test` fails - site still isn't set, but there appear to be some other errors (EObjectDoesNotExist).
* The settings file is reverted, which is good, but it leaves the flash of change in its history.

JeffAbrahamson:
I'm now closing this PR in favour of #72 .

transport_nantes/authentication/tests.py
Qlexio: This changes the value "exemple.com" to "localhost:8000" in Site model. If not done get_current_site in send_activation function fails and all the test fails also

Qlexio:
> At commit [b1df174](https://github.com/transport-nantes/tn_web/commit/b1df17494efc5d379c9e4d4149b96921b04f1744) :
> 
>     * `python manage.py test` fails: the site record is not staged to the testing db.

The test fails probably because I use 'py manage.py test authentication --settings=transport_nantes.settings_tn' command instead of 'python manage.py test'

PR #72, opened 2020-12-10
Fix auth test

Qlexio:
# Fix test issues

* Change call to "account_activation_sent" by call to "account_activation_sent.html" because is_new variable not properly passed
* Remove not unsed anymore "account_activation_sent" function in views.py
* Remove not used anymore "account_activation_sent" path in urls.py
* Correct and translate sentences in authentication/templates

issue #51

### DIFF
--- a/transport_nantes/authentication/templates/authentication/account_activation_invalid.html
+++ b/transport_nantes/authentication/templates/authentication/account_activation_invalid.html
@@ -2,5 +2,5 @@
 {% load static %}
 
 {% block content %}
-  <p>The confirmation link was invalid, possibly because it has already been used.</p>
+  <p>Le lien de confirmation est invalide. Peut-être qu'il a déjà été utilisé ou qu'il a expiré.</p>
 {% endblock %}
--- a/transport_nantes/authentication/templates/authentication/account_activation_sent.html
+++ b/transport_nantes/authentication/templates/authentication/account_activation_sent.html
@@ -7,13 +7,13 @@
 	<div class="col-lg-12">
 	    {% if is_new %}
 	    <h1>Merci !</h1>
-	    <p>Un mél est un route pour que vous puissiez confirmer la
+	    <p>Un mail est un route pour que vous puissiez confirmer la
 		création de votre compte.
 	    </p>
 	    {% else %}
 	    <h1>C'est parti</h1>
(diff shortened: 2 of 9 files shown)


JeffAbrahamson:
PR #71 and #72 refer to the same branch.  This will be overly confusing.

https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-requests

JeffAbrahamson:
Even at ac60d76, tests still fail.  Tests passing is a minimum condition for merging a commit node to master.  This is the whole point of tests.

I'm going to close this PR, because the rebase I'm asking you to do below will corrupt it.  Please create a new PR the meets these conditions:
* settings is not modified
* each commit has a clear comment that explains what it does
* tests pass and are clear at each commit node
* diffs are understandable once one reads the commit comments

To fix this, you'll need to do the following:
* at ac60d76 (the head of `origin/fix_auth_test`), do an interactive rebase against the base of the branch -- `git rebase -i`.  Reorder the commits so that the two settings changes can be squashed together.  Do squash them together.  The history on the settings file will then show that no change ever happened.
* potentially in a second interactive rebase (better for sanity), modify ("edit" action) each commit so that tests pass
* make sure each commit has a clear comment that explains what you've done ("reword, probably also available via "edit" but I don't remember).  Your commits are getting better, but they still have long lines and sometimes aren't clear.
* in a final rebase (non-interactive), move the branch base to the current origin/master.

The only reason to do this in multiple rebases is for sanity.  Doing it all at once is too overwhelming for even most experienced devs.

Remember that you have the reflog to help you remember where you've been.  (You can also note commit numbers (in a file, on paper, ...) to help keep you oriented.)  After each rebase, you'll have diverged from a fast-forward update of origin.  You'll either need to force push (`git push --force`, which is something you should never do on a shared branch) or simply create a new branch (`git checkout -b my-new-branch-name` and work with that.

Remember occasionally to clean up your unused branches: `git branch -d` for local tracking copies and `git push origin --delete` for upstream.  This isn't something that's urgent, it's just polite not to leave messes for too long.

It's one thing to have a commit message like "Fix typo" when visual inspection really shows you just changed something tiny like spelling.  But if there are any logic changes, you need to explain it clearly -- not the functionality but the reasoning, the why.

PR #73, opened 2020-12-11
Password reset

Qlexio:
# Add password reset form

* Add templates to authentication/templates/registration
* Add paths to transport_nantes/urls.py instead of authentication/urls.py because paths don't work in authentication/urls.py probably because of django architecture
* Add condition in settings.py for sending of mail to console in dev mode because of Amazon SES doesn't on local environment and make password reset form bug on local machine
* Add password reset link in authentication/login.html

Issue #51

### DIFF
--- a/transport_nantes/authentication/templates/registration/password_reset_complete.html
+++ b/transport_nantes/authentication/templates/registration/password_reset_complete.html
@@ -1,0 +1,12 @@
+{% extends 'asso_tn/base_transport_nantes.html' %}
+{% load static %}
+
+{% block app_content %}
+
+<div class="container">
+	<div class="d-flex flex-column">
+		<p class="m-auto">Votre mot de passe a été réinitialisé. Vous pouvez vous <a href="{% url 'authentication:login' %}">connecter</a>.</p>
+	</div>
+</div>
+
+{% endblock %}
--- a/transport_nantes/authentication/templates/registration/password_reset_form.html
+++ b/transport_nantes/authentication/templates/registration/password_reset_form.html
@@ -1,0 +1,45 @@
+{% extends 'asso_tn/base_transport_nantes.html' %}
+{% load static %}
+
+
+{% block app_content %}
+
+<style type="text/css">
+  .form-signin {
+    width: 100%;
+    max-width: 330px;
+    padding: 15px;
+    margin: auto;
+  }
+  .form-signin .checkbox {
+    font-weight: 400;
+  }
+  .form-signin .form-control {
+    position: relative;
+    box-sizing: border-box;
+    height: auto;
+    padding: 10px;
+    font-size: 16px;
+  }
+  .form-signin .form-control:focus {
+    z-index: 2;
+  }
+  .form-signin input[type="email"] {
+    margin-bottom: 10px;
+    border-bottom-right-radius: 0;
+    border-bottom-left-radius: 0;
+  }
+
+  .h3{
+    text-align: center;
+  }
+</style>
+
+
+<form method="POST" class="form-signin">{% csrf_token %}
+	<h1 class="h3 mb-3 font-weight-normal">Réinitialisation du mot de passe</h1>
+    <input name="email" class="form-control" placeholder="Adresse mail" type="email" id="id_email" required="true">
+	<button class="btn btn-lg btn-primary btn-block" type="submit">Envoyer le mail</button>  
+</form>
+
+{% endblock %}
--- a/transport_nantes/transport_nantes/settings.py
+++ b/transport_nantes/transport_nantes/settings.py
@@ -190,7 +190,3 @@ if 'STATIC_ROOT' in dir(settings_local):
 CAPTCHA_NOISE_FUNCTIONS = ('captcha.helpers.noise_dots',)
 CAPTCHA_CHALLENGE_FUNCT = 'captcha.helpers.random_char_challenge'
 CAPTCHA_LENGTH = 5
-
-# Captcha condition for testing in dev mod
-if ROLE == "dev":
-    CAPTCHA_TEST_MODE = True
--- a/transport_nantes/transport_nantes/settings.py
+++ b/transport_nantes/transport_nantes/settings.py
@@ -190,3 +190,7 @@ if 'STATIC_ROOT' in dir(settings_local):
 CAPTCHA_NOISE_FUNCTIONS = ('captcha.helpers.noise_dots',)
 CAPTCHA_CHALLENGE_FUNCT = 'captcha.helpers.random_char_challenge'
 CAPTCHA_LENGTH = 5
+
+# Captcha condition for testing in dev mod
+if ROLE == "dev":
+    CAPTCHA_TEST_MODE = True
--- a/transport_nantes/transport_nantes/settings.py
+++ b/transport_nantes/transport_nantes/settings.py
@@ -102,11 +102,15 @@ AUTH_PASSWORD_VALIDATORS = [
 ]
 PASSWORD_RESET_TIMEOUT_DAYS = 1
 
-EMAIL_BACKEND = 'django_amazon_ses.EmailBackend'
-AWS_ACCESS_KEY_ID = settings_local.AWS_ACCESS_KEY_ID
-AWS_SECRET_ACCESS_KEY = settings_local.AWS_SECRET_ACCESS_KEY
-AWS_DEFAULT_REGION = settings_local.AWS_DEFAULT_REGION
-DEFAULT_FROM_EMAIL = settings_local.DEFAULT_FROM_EMAIL
+if ROLE == "dev":
+    EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+    DEFAULT_FROM_EMAIL = 'webmaster@localhost'
+else:
+    EMAIL_BACKEND = 'django_amazon_ses.EmailBackend'
+    AWS_ACCESS_KEY_ID = settings_local.AWS_ACCESS_KEY_ID
+    AWS_SECRET_ACCESS_KEY = settings_local.AWS_SECRET_ACCESS_KEY
+    AWS_DEFAULT_REGION = settings_local.AWS_DEFAULT_REGION
+    DEFAULT_FROM_EMAIL = settings_local.DEFAULT_FROM_EMAIL
 
 LOGGING = {
     'version': 1,
--- a/transport_nantes/transport_nantes/urls.py
+++ b/transport_nantes/transport_nantes/urls.py
@@ -16,6 +16,7 @@ Including another URLconf
 from django.contrib import admin
 from django.urls import include, path
 from asso_tn.views import MainTransportNantes
+from django.contrib.auth import views as auth_views
 
 urlpatterns = [
     path('', MainTransportNantes.as_view(), name='index'),
@@ -34,4 +35,20 @@ urlpatterns = [
     path('tn/', include('asso_tn.urls')),
     path('v/', include('velopolitain.urls')),
     path('vo/', include('velopolitain_observatoire.urls')),
+
+    # Paths below must be here because of django architecture
+    path('password_reset/done/', auth_views.PasswordResetDoneView.as_view(
+        template_name='registration/password_reset_done.html'),
+        name='password_reset_done'),
+    path('reset/<uidb64>/<token>/', auth_views.PasswordResetConfirmView.as_view(
+        template_name='registration/password_reset_confirm.html'
+    ), name='password_reset_confirm'),
+    path('password_reset/', auth_views.PasswordResetView.as_view(
+        template_name='registration/password_reset_form.html',
+        email_template_name='registration/password_reset_email.html',
+        subject_template_name='registration/password_reset_subject.txt'),
+        name='password_reset'),
+    path('reset/done/', auth_views.PasswordResetCompleteView.as_view(
+        template_name='registration/password_reset_complete.html'),
+        name='password_reset_complete'),
 ]
--- a/transport_nantes/authentication/templates/authentication/account_activation_invalid.html
+++ b/transport_nantes/authentication/templates/authentication/account_activation_invalid.html
@@ -2,5 +2,5 @@
 {% load static %}
 
 {% block content %}
-  <p>The confirmation link was invalid, possibly because it has already been used.</p>
+  <p>Le lien de confirmation est invalide. Peut-être qu'il a déjà été utilisé ou qu'il a expiré.</p>
 {% endblock %}
--- a/transport_nantes/authentication/templates/authentication/account_activation_sent.html
+++ b/transport_nantes/authentication/templates/authentication/account_activation_sent.html
@@ -7,13 +7,13 @@
 	<div class="col-lg-12">
 	    {% if is_new %}
 	    <h1>Merci !</h1>
-	    <p>Un mél est un route pour que vous puissiez confirmer la
+	    <p>Un mail est un route pour que vous puissiez confirmer la
 		création de votre compte.
 	    </p>
 	    {% else %}
 	    <h1>C'est parti</h1>
-	    <p>Un mél est un route avec un lien magique qui vous permettra
-		de connecter.
+	    <p>Un mail est un route avec un lien magique qui vous permettra
+		de vous connecter.
 	    </p>
 	    {% endif %}
 	</div>
--- a/transport_nantes/authentication/templates/authentication/home.html
+++ b/transport_nantes/authentication/templates/authentication/home.html
@@ -2,6 +2,6 @@
 {% load static %}
 
 {% block content %}
-  <h2>Welcome, {{ user.get_full_name }} <small>{{ user.username }}</small>!</h2>
-  <p>Thanks for confirming your email address!</p>
+  <h2>Bienvenue, {{ user.get_full_name }} <small>{{ user.username }}</small>!</h2>
+  <p>Merci d'avoir confirmer votre adresse mail!</p>
 {% endblock %}
--- a/transport_nantes/authentication/templates/authentication/login.html
+++ b/transport_nantes/authentication/templates/authentication/login.html
@@ -66,6 +66,7 @@
 		    mot de passe, tapez-le une deuxième fois.</p>
 		<button type="submit" class="btn btn-outline-secondary">Inscription</button>
 	    </form>
+		<p><a href="{% url 'password_reset' %}">Mot de passe oublié?</a></p>
 	</div>
     </div>
 </div>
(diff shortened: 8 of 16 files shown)


transport_nantes/authentication/templates/registration/password_reset_complete.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'asso_tn/base_transport_nantes.html' %}
+{% load static %}
+
+{% block app_content %}
+
+<div class="container">
+	<div class="d-flex flex-column">
+		<p class="m-auto">Votre mot de passe a été réinitialisé. Vous pouvez vous <a href="{% url 'authentication:login' %}">connecter</a>.</p>
```
JeffAbrahamson, line 8: This line is too long.

transport_nantes/authentication/templates/registration/password_reset_form.html
```diff
@@ -0,0 +1,45 @@
+{% extends 'asso_tn/base_transport_nantes.html' %}
+{% load static %}
+
+
+{% block app_content %}
+
+<style type="text/css">
```
JeffAbrahamson, line 7: These changes deserve separate comment.  Why do they belong here and not in `transport_nantes.css`, for example?  Why do you provide a maximum width of 330 pixels on the sign-in form?  Does that work in a responsive context (on different screen sizes)?

transport_nantes/authentication/templates/registration/password_reset_form.html
```diff
@@ -0,0 +1,45 @@
...
+  .form-signin .form-control:focus {
+    z-index: 2;
+  }
+  .form-signin input[type="email"] {
+    margin-bottom: 10px;
+    border-bottom-right-radius: 0;
+    border-bottom-left-radius: 0;
+  }
+
+  .h3{
```
JeffAbrahamson, line 33: Why is this a style on all h3's rather than just centering some specific h3?

transport_nantes/authentication/templates/registration/password_reset_form.html
```diff
@@ -0,0 +1,45 @@
...
+    text-align: center;
+  }
+</style>
+
+
+<form method="POST" class="form-signin">{% csrf_token %}
+	<h1 class="h3 mb-3 font-weight-normal">Réinitialisation du mot de passe</h1>
+    <input name="email" class="form-control" placeholder="Adresse mail" type="email" id="id_email" required="true">
```
JeffAbrahamson, line 41: long line

Perhaps consider using pylint?  (I've not checked if it works well for django, but there must be similar tools if it doesn't.)

transport_nantes/transport_nantes/settings.py
```diff
@@ -102,11 +102,15 @@
 ]
 PASSWORD_RESET_TIMEOUT_DAYS = 1
 
-EMAIL_BACKEND = 'django_amazon_ses.EmailBackend'
-AWS_ACCESS_KEY_ID = settings_local.AWS_ACCESS_KEY_ID
-AWS_SECRET_ACCESS_KEY = settings_local.AWS_SECRET_ACCESS_KEY
-AWS_DEFAULT_REGION = settings_local.AWS_DEFAULT_REGION
-DEFAULT_FROM_EMAIL = settings_local.DEFAULT_FROM_EMAIL
+if ROLE == "dev":
```
JeffAbrahamson, line 105: This is quite questionable and so needs extensive comment.  What exactly depends on this change?  Why did something local (like testing auth) require accommodation at global scope?

Is it not possible to modify the settings you need in the tests setup function?

transport_nantes/transport_nantes/urls.py
```diff
@@ -34,4 +35,20 @@
     path('tn/', include('asso_tn.urls')),
     path('v/', include('velopolitain.urls')),
     path('vo/', include('velopolitain_observatoire.urls')),
+
+    # Paths below must be here because of django architecture
```
JeffAbrahamson, line 39: Why do these need to happen at global scope?  Have you read the contrib.auth package to understand what exactly creates this dependency and that there's really no work-around?  It seems quite surprising that library developers would impose this sort of root-of-the-tree obligation on their client developers.  It breaks encapsulation, which makes me think that it's not necessary.  (How could something that so breaks encapsulation have made its way through code review and into a shipping version of a mature tool like django?)

JeffAbrahamson:
* tests fail
* you don't need to repeat in a PR text what you've already written in commits -- indeed, you shouldn't -- rather, just explain briefly what the PR provides -- a PR is a discussion, and a diligent reviewer will begin by reading your commits

This branch is based on fix_auth_test, which is based on auth_test.  So the rebase requested in #71 / #72 will corrupt this PR.

You probably want to make each PR its own branch so that PR's can advance independently of each other.  How would you advance this PR if remember_me were rejected, for example, without also affecting `remember_me`?

transport_nantes/authentication/templates/registration/password_reset_form.html
Qlexio: Actually, I used a ready to use template and I forgot to remove CSS

transport_nantes/authentication/templates/registration/password_reset_form.html
Qlexio: It's kind of weird for this part. I use "python-dev-tools" which a powerfull linter and formatter but it doesn't seem to work with django. I've even looked in configuration file but I can't figure ot why it's not working well.

transport_nantes/transport_nantes/settings.py
Qlexio: I've done that because Amazon SES doesn't work on local environment and makes this part bug with normal use and tests.
Adding this line also permit me to see sent mail and login/ reset link and that variables are well passed into the mail.
It is pretty usefull to see the mail and reset password sends mail. And like we don't want to change django librairie, I add to find a way to get the mail.

transport_nantes/transport_nantes/urls.py
Qlexio: We actually can put "password_reset" path in authentication/urls.py but password_reset calls "password_reset_done" and doesn't seem find it unless if it is in global scope. 
I still have not found a way to by pass that.

transport_nantes/transport_nantes/settings.py
JeffAbrahamson: Distinguish between your ad hoc testing during development (important) and unit tests, regression tests, etc. that are run automatically.  What you "see" is useful for you during development.  It is not related to automatic tests.

Why can't you change the settings value just in your test?

transport_nantes/transport_nantes/urls.py
JeffAbrahamson: What does "seem to" mean here?

Qlexio: Probably a bad translation. "Doesn't seem to find" => "Ne semble pas trouver"
I am not sure that "seem to" is the good word to use.

transport_nantes/transport_nantes/settings.py
Qlexio: I probably can do that for tests but website still doesn't run properly on usual use if I don't do that

transport_nantes/transport_nantes/urls.py
JeffAbrahamson: Yeah, there's a difference, agreed.  I'm still not sure what you mean, though.  You don't provide explain the precise error.

transport_nantes/transport_nantes/settings.py
JeffAbrahamson: This assertion needs more detail.  "The website doesn't run" - what errors, what context, how, what do you expect, what do you observe?

transport_nantes/authentication/templates/registration/password_reset_form.html
JeffAbrahamson: Well, this file is html, I wouldn't expect _this_ one to be helped by pylint...

JeffAbrahamson: (So my confusing suggestion with specific problem, my bad.  But using a lint tool is still a good idea.)

transport_nantes/transport_nantes/settings.py
Qlexio: Here is the error from the website when it wants to send mail with Amazon SES:
>ClientError at /password_reset/
>An error occurred (InvalidClientTokenId) when calling the SendRawEmail operation: The security token included in the request is invalid.

transport_nantes/transport_nantes/urls.py
Qlexio: Here is the error when paths are in "authentication/urls.py instead of global scope:
>NoReverseMatch at /auth/password_reset/
>Reverse for 'password_reset_confirm' not found. 'password_reset_confirm' is not a valid view function or pattern name.

transport_nantes/transport_nantes/settings.py
JeffAbrahamson: Yes, you don't have AWS SES credentials in dev.

Your job as a developer is not just to find bugs but to understand them, at least when it comes to the code base on which you're working, and ideally to propose fixes.

So, what's the stack trace?  Why is it trying to send mail with AWS?  How can we make it not try to send mail?  Because no mails should be sent out of dev, it's far too dangerous.

transport_nantes/transport_nantes/urls.py
JeffAbrahamson: Did you look into the auth code to understand why it says that?  How does the auth code foresee dealing with this?  If all else fails, have you tried asking the authors of the auth code or asking the django community what the intent is?

I do not believe that django, a relatively well-thought-through framework, makes such demands on path names.

transport_nantes/transport_nantes/settings.py
Qlexio: Here is the error from the website when it wants to send mail with Amazon SES:
>ClientError at /password_reset/
>An error occurred (InvalidClientTokenId) when calling the SendRawEmail operation: The security token included in the request is invalid.

Qlexio: 1. It tries to send mail thru AWS because EMAIL_BACKEND constant in settings is set to use AWS.
2. password_reset uses PasswordResetForm which call send_mail function. So to not send mail, I would have to change this part of the code but I don't think that it's secure to modify django librairie. I may break the librairie or create a bug and mostly, we may have to do the job everytime django is upgraded. I'm pretty sure that nobody will think to write the changes again at each upgrade.
3. Send mail out of dev is too dangerous, so I must not use AWS. And I don't have credentials in dev so it is ok. But django allows me two others options:
  1. I can set django to write mail in file on my own computer in dev but I don't want to have plenty of useless files.
  2. I can set django print mail in terminal.
In both ways, mails stay on my machine, so in local. I prefer the second solution because the mail is emetted from local server to the terminal print every messages about the server itself. It stays on my computer, there is no save and I can easily use link inside the mail if needed.

transport_nantes/transport_nantes/urls.py
Qlexio: Actually, I think that it fails because of line 215 of auth code saying:
>success_url = reverse_lazy('password_reset_done')
There is the same problem on line 252:
>success_url = reverse_lazy('password_reset_complete')

So I can easily put 2 paths in authentication/urls but for the 2 others it looks like auth code must be modify.
I can do it and I am pretty sure that it will work but it is the same problem that above. I may be erased on django upgrade and nobody will remember to do the change the code every times.

JeffAbrahamson: Hmm, so this seems to have gotten stuck.

What does `reverse_lazy()` do?  What indication does that give you about the design intentions of the author?

transport_nantes/transport_nantes/settings.py
JeffAbrahamson: What does the code look like that tries to send mail?  What is the stack trace that gets there?  Can you see anything in that code or in the comments that indicates the design intentions of the authors?  Do you suppose the authors expected tests to send mails via production or other mail servers?

When tests pass, they should print almost nothing.  The state of passing or failing mustn't require a human to know anything about the test, only that it will say passed or failed.

transport_nantes/transport_nantes/urls.py
Qlexio: `reverse_lazy()` is a defered version of usual `reverse()`
I understand what it does but I am not sure to be able to well explain it. 
So here is a [link to the docs](https://docs.djangoproject.com/en/3.1/ref/urlresolvers/#reverse-lazy).
We could try to add this line: `path('accounts/', include('django.contrib.auth.urls')),` to authentication.urls.py.
This way is, it includes every needed path but with this option, we can't have custom templates, email, and so on;
because theses parameters are passed in urls.py.
So maybe if I learn more about `include` I could find a middle way between hard coding and use of of include

JeffAbrahamson: I know what it does, I'm trying to get you to think this through and not need to turn to me.  (I'm happy to help, but I'd rather help you learn than give you answers to specific problems.)

transport_nantes/transport_nantes/settings.py
Qlexio: Here is the code from auth.views
`class PasswordResetView(PasswordContextMixin, FormView):
    email_template_name = 'registration/password_reset_email.html'
    extra_email_context = None
    form_class = PasswordResetForm
    from_email = None
    html_email_template_name = None
    subject_template_name = 'registration/password_reset_subject.txt'
    success_url = reverse_lazy('password_reset_done')
    template_name = 'registration/password_reset_form.html'
    title = _('Password reset')
    token_generator = default_token_generator

    @method_decorator(csrf_protect)
    def dispatch(self, *args, **kwargs):
        return super().dispatch(*args, **kwargs)

    def form_valid(self, form):
        opts = {
            'use_https': self.request.is_secure(),
            'token_generator': self.token_generator,
            'from_email': self.from_email,
            'email_template_name': self.email_template_name,
            'subject_template_name': self.subject_template_name,
            'request': self.request,
            'html_email_template_name': self.html_email_template_name,
            'extra_email_context': self.extra_email_context,
        }
        form.save(**opts)
        return super().form_valid(form)`

And the form used by above class:
`class PasswordResetForm(forms.Form):
    email = forms.EmailField(
        label=_("Email"),
        max_length=254,
        widget=forms.EmailInput(attrs={'autocomplete': 'email'})
    )

    def send_mail(self, subject_template_name, email_template_name,
                  context, from_email, to_email, html_email_template_name=None):
        """
        Send a django.core.mail.EmailMultiAlternatives to `to_email`.
        """
        subject = loader.render_to_string(subject_template_name, context)
        # Email subject *must not* contain newlines
        subject = ''.join(subject.splitlines())
        body = loader.render_to_string(email_template_name, context)

        email_message = EmailMultiAlternatives(subject, body, from_email, [to_email])
        if html_email_template_name is not None:
            html_email = loader.render_to_string(html_email_template_name, context)
            email_message.attach_alternative(html_email, 'text/html')

        email_message.send()

    def get_users(self, email):
        """Given an email, return matching user(s) who should receive a reset.
        This allows subclasses to more easily customize the default policies
        that prevent inactive users and users with unusable passwords from
        resetting their password.
        """
        email_field_name = UserModel.get_email_field_name()
        active_users = UserModel._default_manager.filter(**{
            '%s__iexact' % email_field_name: email,
            'is_active': True,
        })
        return (
            u for u in active_users
            if u.has_usable_password() and
            _unicode_ci_compare(email, getattr(u, email_field_name))
        )

    def save(self, domain_override=None,
             subject_template_name='registration/password_reset_subject.txt',
             email_template_name='registration/password_reset_email.html',
             use_https=False, token_generator=default_token_generator,
             from_email=None, request=None, html_email_template_name=None,
             extra_email_context=None):
        """
        Generate a one-use only link for resetting password and send it to the
        user.
        """
        email = self.cleaned_data["email"]
        if not domain_override:
            current_site = get_current_site(request)
            site_name = current_site.name
            domain = current_site.domain
        else:
            site_name = domain = domain_override
        email_field_name = UserModel.get_email_field_name()
        for user in self.get_users(email):
            user_email = getattr(user, email_field_name)
            context = {
                'email': user_email,
                'domain': domain,
                'site_name': site_name,
                'uid': urlsafe_base64_encode(force_bytes(user.pk)),
                'user': user,
                'token': token_generator.make_token(user),
                'protocol': 'https' if use_https else 'http',
                **(extra_email_context or {}),
            }
            self.send_mail(
                subject_template_name, email_template_name, context, from_email,
                user_email, html_email_template_name=html_email_template_name,
            )`
We can see that views 's class calls `form.save()` which sends mail.
Actually, I think that the authors have choosen to send mail for password recovery because of security purposes.
It is more secure than changing password directly on website. It is a kind of two-factors authentication but by mail instead of using phones.

Anyway, my problem comes from `send_mail()`. No matters from where it is called. settings is configured to use AWS right now. And for dev purposes, django has this solution:
`EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'`
and [I join the doc](https://docs.djangoproject.com/fr/2.2/topics/email/#console-backend) for full explaination.
Just one thing about that is that we have to remove this line for production

transport_nantes/transport_nantes/urls.py
Qlexio: Actually, I just thougth that I could try to use heritage to pass the differents django password variables into a new function and override them with the values I want

Qlexio: This way new function behave like classical function in views and I can write paths in authentication.urls.
I'll probably try it this afternoon after I'm done with my actual work that I want to finish before.

transport_nantes/transport_nantes/settings.py
JeffAbrahamson: I have that source code, too.

If you believe that the library requires sending mail (which is inappropriate in a test), that would be a bug.  Why don't you file a bug against the library?  Or ask in some discussion forum where the library maintainers hang out what they intended.  I'm sure you'll learn something...

If you believe what you are saying, act on it.
If you don't believe it, then do the work to understand the code.
Developing software is work.  It happens in the cool glow of a monitor, mostly alone.  Sometimes we ask other people for help.  Those people will expect to offer small bits of help, just to help us get unstuck.  Not to solve our problems.

Qlexio: For testing, I can use the scratch code you gave me. It will work. And I don't really need mails in test so set `settings.ROLE = "test"` and change `if ROLE == "dev"` to `if ROLE == "test"` is perfect for the testing part to make send_mail() not to bug during tests. And for any values transmitted by mail, test code automatically generate them to simulate a sent mail. So for this part, it is ok

But where I am stuck is when I navigate on website in browser. It is not most of my work but I need to watch the design of the templates I write. 
Actually, I don't even need the mail, I need the token wrote in the mail to print pages.
It's about only 2 pages. So, for `send_activation()` you solved the problem. It still is password_reset template.
So, I got a solution from lots of different forums and even from django official documentation but you don't agree with it.
Now, I really don't know what to do except asking AWS credentials but it means sending mails on real mail account on web and we don't want it. It's dangerous for security of the project, it is spamming and most of all, we don't have unlimited mail address.

PR #74, opened 2020-12-11
Remember me

Qlexio:
# Add remember me checkbox

* Add test for session cookie expiration
* Add new variable in send_activation function to be passed to activation function thru mail
* Set session cookie expiration in activation function
* Change "activation/<token>" url path to "activation/<remember_me>/<token>" to take new variable get by mail in count
* Remove not used anymore SESSION_COOKIE_AGE constant in settings.py

issue #51


JeffAbrahamson:
This is way too big for a single PR.  Could you please, please, split this into individual PR's for individual things.

In addition, you're way behind origin/master in your branch points.

PR #79, opened 2020-12-15
Login by mail

Qlexio:
#Add two options authentication

* Add tests
* Add account creation with password
* Add logging in with password

### DIFF
--- a/transport_nantes/authentication/forms.py
+++ b/transport_nantes/authentication/forms.py
@@ -14,11 +14,13 @@ class SignUpForm(UserCreationForm):
     password1 = forms.CharField(
         widget=forms.PasswordInput, max_length=254,
         label="Mot de passe", required=False,
-        help_text="Facultatif")
+        help_text="Facultatif", min_length=8,
+        error_messages=dict(min_length="Le mot de passe doit faire plus de 8 caractères"))
     password2 = forms.CharField(
         widget=forms.PasswordInput, max_length=254,
         label="Mot de passe", required=False,
-        help_text="Encore la même chose")
+        help_text="Encore la même chose", min_length=8,
+        error_messages=dict(min_length="Le mot de passe doit faire plus de 8 caractères"))
     remember_me = forms.BooleanField(required=False, label="Se souvenir de moi",
         initial=False, widget=forms.CheckboxInput)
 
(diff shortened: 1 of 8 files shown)


JeffAbrahamson:
Commit 2f0c562 contains a comment whose title is that it is a cherry pick from a commit that is no longer visible.  This makes no sense.  In addition, reading commit logs, it is uniformative.

```
66a6871|2020-12-15|Alexandre BENARD| (HEAD, origin/login_by_mail) Add tests about cookie expiration
5c231a3|2020-12-15|Alexandre BENARD| Change for both authentication with/out password
130fed6|2020-12-14|Alexandre BENARD| Add condition for non existing user
2f0c562|2020-12-11|Alexandre BENARD| Cherry-pick from 6bed84  <-- what happened here??
2b7cd73|2020-11-25|Jeff Abrahamson| Add join menu item at top of page
3ee1409|2020-11-25|Jeff Abrahamson| Remove donation button in hero image
e41dc93|2020-11-25|Jeff Abrahamson| Customise html titles throughout
88cd01b|2020-11-25|Jeff Abrahamson| Print rather than send mail in dev mode
4c7a189|2020-11-25|Jeff Abrahamson| Remove backup file that should not have been committed
c414714|2020-11-25|Jeff Abrahamson| Add string conversion for ClickAndCollect
```

JeffAbrahamson:
2f0c562 contains too-long-lines, lines with trailing spaces.  Code style (google "code smell") is an important part of communication.   Code is a way we communicate with each other and with our future selves.

JeffAbrahamson:
tests fail

Qlexio:
I don't know the command to come back to a previews commit so I test on 66a6871 and all tests passes for me except the one about token that we still need to talk about.
On which commit tests fail? and what command have you used?
`python manage.py test` won't work, we have to use `python manage.py test --settings=transport_nantes.settings_tn` because `settings_tn.py` is the file getting all settings parameters.

JeffAbrahamson:
I've been testing with

    python manage.py test --settings=transport_nantes.settings_tn

At `origin/password_reset`, for example, I see multiple failures.
Same at `origin/login_by_mail` and `origin/remember_me`.

Qlexio:
At `origin/password_reset`, it is entirely my fault. I translated and corrected sentences in previous commit but I forgot to change the values in test.py.
At `origin/login_by_mail` and `origin/remember_me`, there should be one failure on `test_mail_with_used_token (authentication.tests.TokenMailTest)`. I know about it and ask you twice about the property of the token that causes the failure. 
A simple discussion about this property would have be good but we didn't have it.
If there are more failures at those two branches, I at least need to know what function fails or how many to solve it. 
I can do nothing with an only "tests fail". I need to know if you see the same thing than me or not to go further.

PR #84, opened 2020-12-17
New password reset

Qlexio:
Add custom password reset form and tests


JeffAbrahamson:
The branch hasn't changed since my comments a short bit ago.  So please fix those things first.  That will require a force push, which will corrupt this PR.  So I'm going to close this now.

PR #98, opened 2021-01-21
Change settings mail variables

Qlexio:
* In settings.py on line 105, set EMAIL_BACKEND to call
settings_local.EMAIL_BACKEND instead of directly AWS
* In settings_local.py.TEMPLATE which should be copy in settings_local:
  * Add condition according to ROLE value. If ROLE == "dev", mail
  variables are set to use Django core function and print mail in
  console.
  Otherwise, mail variables call AWS which should only be used in prod
  and beta versions.

issue #51

@JeffAbrahamson 
I want to add that I've tested what I've done changing ROLE value to something different of "dev"
I got the wanted bug telling me that AWS was properly set in EMAIL_BACKEND variable and well called.
For tests, I've tested without and without setting `ROLE` in `setUp` function as you suggested me. It makes no difference.
So, I think that it is alright but I still suggest you to test it on beta version if if work right.


JeffAbrahamson:
How did you test this?

I'm trying this, and it fails:

    ROOT_URLCONF=transport_nantes.urls_m DEFAULT_SITE_ID=1 \
        python manage.py test --settings=transport_nantes.settings_tn

The error is this:

```
    EMAIL_BACKEND = settings_local.EMAIL_BACKEND
AttributeError: module 'transport_nantes.settings_local' has no attribute 'EMAIL_BACKEND'

```

JeffAbrahamson:
Oh, scratch that, fat fingers on my part.

JeffAbrahamson:
Harumph.  That succeeds with this

```
ROOT_URLCONF=transport_nantes.urls_m DEFAULT_SITE_ID=1 \
      python manage.py test --settings=transport_nantes.settings_tn
```

but not with this

```
ROOT_URLCONF=transport_nantes.urls_tn DEFAULT_SITE_ID=1\
       python manage.py test --settings=transport_nantes.settings_tn
```

So that's certainly a fragility I inadvertenly introduced with the multiple URL scheme.  It wants to test things that don't have paths in the current scenario.

I wonder how best to handle that.  A shortcut is certainly to deprecate the old TN site, but that's a temporary solution, since we have other sites planned.

Qlexio:
@JeffAbrahamson 

You should take a look at the other PR. Tests run on both urls_m and urls_tn if I've done everything well

Qlexio:
I guess that it's not working because you probably have done a rebase; thing that i'm not sure to be not when I worked on it.

JeffAbrahamson:
"The other PR" ?  There are 3.  Which one?
I'm just going in numerical order...

Yeah, I rebased to origin/master.  I'm in detached head mode, so nothing really persistent (i.e., the rebase commit object will be cleaned up if I don't do something with it).

JeffAbrahamson:
Ah, but I see what you mean.  Without the tests you wrote, it doesn't make sense to test this.

Fair enough, I'll come back to this.

PR #99, opened 2021-01-29
Login by mail 3

Qlexio:
@JeffAbrahamson 

I made some changes to make tests run well.
I've done that step by step to let you being able to go back if you want.
Everything is explained in the commits. Let me know if you want me to explain you why I did that this way.

### DIFF
--- a/transport_nantes/authentication/forms.py
+++ b/transport_nantes/authentication/forms.py
@@ -14,11 +14,13 @@ class SignUpForm(UserCreationForm):
     password1 = forms.CharField(
         widget=forms.PasswordInput, max_length=254,
         label="Mot de passe", required=False,
-        help_text="Facultatif")
+        help_text="Facultatif", min_length=8,
+        error_messages=dict(min_length="Le mot de passe doit faire plus de 8 caractères"))
     password2 = forms.CharField(
         widget=forms.PasswordInput, max_length=254,
         label="Mot de passe", required=False,
-        help_text="Encore la même chose")
+        help_text="Encore la même chose", min_length=8,
+        error_messages=dict(min_length="Le mot de passe doit faire plus de 8 caractères"))
     remember_me = forms.BooleanField(required=False, label="Se souvenir de moi",
         initial=False, widget=forms.CheckboxInput)
 
--- a/transport_nantes/authentication/forms.py
+++ b/transport_nantes/authentication/forms.py
@@ -19,6 +19,8 @@ class SignUpForm(UserCreationForm):
         widget=forms.PasswordInput, max_length=254,
         label="Mot de passe", required=False,
         help_text="Encore la même chose")
+    remember_me = forms.BooleanField(required=False, label="Se souvenir de moi",
+        initial=False, widget=forms.CheckboxInput)
 
     class Meta:
         model = User
--- a/transport_nantes/authentication/templates/authentication/account_activation_email.html
+++ b/transport_nantes/authentication/templates/authentication/account_activation_email.html
@@ -9,5 +9,5 @@ Le lien suivant vous le permettra.
 Si toutefois ce mél ne correspond pas à votre demande, vous
 pouvez simplement supprimer ce mél.{% endif %}
 
-http://{{ domain }}{% url 'authentication:activate' token=token %}
+http://{{ domain }}{% url 'authentication:activate' token=token remember_me=remember_me %}
 {% endautoescape %}
--- a/transport_nantes/authentication/templates/authentication/account_activation_invalid.html
+++ b/transport_nantes/authentication/templates/authentication/account_activation_invalid.html
@@ -2,5 +2,5 @@
 {% load static %}
 
 {% block content %}
-  <p>The confirmation link was invalid, possibly because it has already been used.</p>
+  <p>Le lien de confirmation est invalide. Peut-être qu'il a déjà été utilisé ou qu'il a expiré.</p>
 {% endblock %}
--- a/transport_nantes/authentication/templates/authentication/account_activation_sent.html
+++ b/transport_nantes/authentication/templates/authentication/account_activation_sent.html
@@ -7,13 +7,11 @@
 	<div class="col-lg-12">
 	    {% if is_new %}
 	    <h1>Merci !</h1>
-	    <p>Un mél est un route pour que vous puissiez confirmer la
-		création de votre compte.
+	    <p>Un mail est un route pour que vous puissiez confirmer la création de votre compte.
 	    </p>
 	    {% else %}
 	    <h1>C'est parti</h1>
-	    <p>Un mél est un route avec un lien magique qui vous permettra
-		de connecter.
+	    <p>Un mail est un route avec un lien magique qui vous permettra de vous connecter.
 	    </p>
 	    {% endif %}
 	</div>
--- a/transport_nantes/authentication/templates/authentication/login.html
+++ b/transport_nantes/authentication/templates/authentication/login.html
@@ -39,6 +39,7 @@
 		    {% if form.password1.help_text %}
 		    <small style="color: grey">{{ form.password1.help_text }}</small>
 		    {% endif %}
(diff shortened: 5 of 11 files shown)


Qlexio:
@JeffAbrahamson 

I forgot to tell you that all theses changes might be used for the branch `new_password_reset`

JeffAbrahamson:
Ah, so here you've combined more than one task, I think.  I'm going to proceed commit by commit.

JeffAbrahamson:
In this line

```
path('activate/<remember_me>/<token>', views.activate, name='activate'),
```

 I wonder if it wouldn't be better to integrate `remember_me` into the token parser.  At issue is that someone could modify the URL in the email and so trigger an undesired state transition.

Qlexio:
@JeffAbrahamson 

I see what you mean.

We can change on line 122: 
`python
if` remember_me == "False":
            request.session.set_expiry(0)
else:
            request.session.set_expiry(60*60*24*30)`
to:
`python
if remember_me == "False":
            request.session.set_expiry(0)
elif remember_me == "True":
            request.session.set_expiry(60*60*24*30)
else:
            return render(request, 'authentication/account_activation_invalid.html')`

JeffAbrahamson:
You'll want to fix your markdown to make that readable. ;-)

Qlexio:
@JeffAbrahamson

I see what you mean.

We can change on line 122:
`if remember_me == "False":`
`    request.session.set_expiry(0)`
`else:`
`    request.session.set_expiry(60 * 60 * 24 * 30)`
to: 
`if remember_me == "False":`
`    request.session.set_expiry(0)`
`elif remember_me == "True":`
`    request.session.set_expiry(60602430)`
`else:`
`    return render(request, 'authentication/account_activation_invalid.html')`

I suggest that because I don't know to get back this boolean once in the token parser

JeffAbrahamson:
Well, the `60 * 60 * 24 * 30` is more readable.  Even better would be
```
SECONDS_IN_MONTH = 60 * 60 * 24 * 30
```
and then later
```
request.session.set_expiry(SECONDS_IN_MONTH)
```

JeffAbrahamson:
I'm seeing one test failure:

```
FAIL: test_mail_with_used_token (authentication.tests.TokenMailTest)
Traceback (most recent call last):
  File "/home/jeff/src/jma/transport-nantes/tn_web/transport_nantes/authentication/tests.py", line 503, in test_mail_with_used_token
    self.assertNotEqual(response2.url, "/")
AssertionError: '/' == '/'
```

Qlexio:
It is what I told you.
This token can be unlimitedly used until the end of its time.
It is what this failure shows

JeffAbrahamson:
Ah, I see.  so perhaps the test should check that it's not equal so that it passes?

JeffAbrahamson:
BTW, I've just pushed a commit on `login-by-mail-4`, which is `login-by-mail-3` rebased to `origin/master`.

Qlexio:
It we turn it to an equality, it would mean that the second time we use the token we go on index page and we don't want that.
We should only go to index page the first time we use the token.
Ok for login-by-mail-4

JeffAbrahamson:
Oh, I see what you mean, maybe.  So is this issue that running a test for `token_valid()` is hard because it's hard-coded to look at the current time, and so any time at all introduces a race condition?  So let's add an argument for testing.  Hold on a sec.

Qlexio:
It is tested to hold on for a couple of seconds and test passes.

JeffAbrahamson:
Doesn't that create a race condition (or a needlessly long test)?

JeffAbrahamson:
Yeah, you're counting on the test taking a second to complete, no?

Qlexio:
No, I've done like this:
`# Create token with user.pk`
`token = make_timed_token(user.pk, 1/60) `
`sleep(2)`

JeffAbrahamson:
Ah, missed the sleep.  Sorry.

JeffAbrahamson:
I've added a test for timed tokens (and support for said test).  Have a look at e00644b5.

JeffAbrahamson:
The timed token is only used in auth.  I'll add an extra argument for session persistence.

JeffAbrahamson:
You have a test `test_mail_with_used_token` but I don't see where you persist tokens to check on them later.  (I wasn't planning on storing that state, but if you did, I'd like to understand it.)

JeffAbrahamson:
I'm going to take a break for a bit, as I have a couple things to do this evening.

I'm on the head of `login-by-mail-4`, commit 0c9d2c7 if you want to have a look.

Qlexio:
@JeffAbrahamson 

I managed to make the token to be used only once since now.
You should take a look.
There is a fail in àsso_tn/tests` but I guess that it is because `token_valid` function is used multiple times in the test and it can only be used once now. So, this test should split in 3 parts.

JeffAbrahamson:
Thanks!

I've a number of concerns with this commit (cdf04122).

1.  I see this bit of code

```
+    user = User.objects.get(pk=user_id)
+
+    user.profile.once_token = True
+    user.save()
+
+    if user.profile.once_token:
```

The if clause seems spurious, since it's set to True just before, no?

More fundamentally, though,

2.  Suppose we have some reason to send the user more than one timed token (e.g., user signs up for a mailing list and a petition on different devices, not logged in, each sends a timed token asking for confirmation).  The first will work, the second will fail.  Putting the state on the user/profile object makes it a singleton.

3.  Why? I mean, what do we gain anyway by having these objects be usable once?  I can think of things we lose, like if the user sees the timed token on mobile, clicks, realises for some reason s/he wants to see it on desktop or in a different browser, but, now, sorry, can't do that without issuing a new timed token.

But what's gained in security?  If the token is valid for 20 minutes (useful for avoiding historical debts where a token from last month can be used for something), what is the thread model that we are addressing by making the token presentable only once?

Qlexio:
@JeffAbrahamson 

I see what you mean.
Actually, for me, a token is set to last some time or until it is clicked.
I stated that the token was used only used for authentication not in some other function.

But setting token to be used only once secures user's profile and data. I mean that if user requests a token, right now, he receives it in its mailbox and can click on it during 20mn to connect to the website. If the mailbox has been hacked, the hacked also can use this token during 20mn because of the time availability of the token. In this case, if the token is only used once, we can hope that the user click on the token link first.

It is not really a big deal but the eventuality I wrote above has to be taken in count.
You can remove this part of code if you want to or we can make modifications.

We can pass a value in the hmac, different for each token in `make_timed_token` function and keep it in db.
Then in `token_valid` function, we check for this value, I mean:
if the value still is in DB, token passes and we erase the value from db. So it can't be used anymore.
if the value is not in DB anymore, it means that token has been used and we return invalid page.

To be more perfect, we also should erase the value above from the DB if the token is timed out.

JeffAbrahamson:
> We can hope that

Hope isn't a strategy.  When it comes to security, where the threat model is intentional attackers rather than just people making mistakes, one has to assume that either a system can handle bad actors or else it will fail.  It at least has to make an attempt to handle bad actors. 

This is different from designing systems to help avoid mistakes.  This is why a low barrier works well to help people not fall over a cliff at a viewing platform (Aiguille du Midi) but a full enclosure might be needed in famous places that attract suicide attempts (Tour Eiffel).

I'm not convinced that there is a threat model against which we are defending with a one use policy.  In addition, this strategy adds the overhead of one or more database accesses per interaction that generates or that uses a token.

If the user has lost control of her/his email account, the game is already lost.  The only way I know that we could address that is with 2FA, and that is an overhead that few users of our site will likely find acceptable.  Even asking for a password fails in this case, since the attacker can request to reset the password.

Qlexio:
I actually totally agree with you.
But I had to make the suggestion to explore another way we should not have thought.

You can erase commit ([cdf0412](https://github.com/transport-nantes/tn_web/commit/cdf04122c128f66beecc99177b1e74d8d5225cfd)) which is not needed anymore.

Issue #121, opened 2021-04-26
Clean up old cruft


JeffAbrahamson:
in preparation for @Shriukan33 start

JeffAbrahamson:
Notes:
* I'm pretty sure `velopolitain` is obsolete and can be removed.
* I'm pretty sure `velopolitain_observatoire` is obsolete and can be removed.

JeffAbrahamson:
Tasks remaining:
* remove sites framework
* remove ROOTURLCONF (also from deploy/supervisor)

Thngs that remain:
* clickcollect is obsolete but we need to migrate the collected names, maybe send them mail -- once removed, we'll surely forget
* opengraph isn't currently used but has some stuff that might be useful for generating dynamic OG data
* velopolitain_observatoire, along with the button branch, has some useful stuff for the upcoming OGM

PR #122, opened 2021-04-26
Fix grand_nantes deletion #121

Shriukan33:
Part of #121 
Replaced urls linked to it with # 
Edited settings.py
Edited .gitgnore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,13 @@
 # Python
 venv
 *.pyc
+Pipfile*
 
 # Django
 db.sqlite3
 
 # Config
 aws_config
+
+#VSCODE
+.vscode/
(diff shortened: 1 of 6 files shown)


JeffAbrahamson:
Thoughts:
* The .gitignore change is unrelated and should be a separate commit.
* `base_transport_nantes.html` we want to delete, so it doesn't matter what you do with it.
* In `base_transport_nantes.html` you have a line that ends with white space, which is bad form.  It doesn't matter here since we want to remove the file, but it would be good to convince your tools to show you trailing white space.
* `asso_tn/panel/projects.html` is also slated for deletion...
* The `DEFAULT_SITE_ID` change is also something we intend to delete, but you should become adept at setting environment variables at launch time, this is a pretty standard paradigm.  It's often undesirable to set defaults like this, because in production a missing default is probably a serious error and _should_ die.
* `velopolitain/templates/velopolitain/gilets.html` is another file we intend to delete.  You can see this because it starts with `% extends 'asso_tn/hero_base_transport_nantes.html' %}`

More serious is the commit comment itself.  It should follow the guidelines I gave you of having a single declarative line ("Fix grand_nantes deletion" is fine) and then body comment in proper English (or French).  The "Part of #121." should be on a line by itself.

So could you amend the commit comment please?

Issue #78, opened 2020-12-12
Set up unit tests


JeffAbrahamson:
Make one test per application just to get going.

Some resources:
* [MDN](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Testing)
* [Writing and running tests](https://docs.djangoproject.com/en/3.1/topics/testing/overview/)
* [Writing your first Django app, part 5](https://docs.djangoproject.com/en/3.1/intro/tutorial05/)
* [Testing](https://www.django-rest-framework.org/api-guide/testing/)
* [well-known-urls](https://adamj.eu/tech/2020/06/28/how-to-add-a-well-known-url-to-your-django-site/) has a good (basic) bit about testing with link for further reading
* [sites](https://www.valentinog.com/blog/django-vhosts/) has suggestions on testing multiple site config

Aside:
* [MDN Django](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django)

PR #123, opened 2021-04-27
Add main page status code test

Shriukan33:
Part of #78
Tests http response on the main page

### DIFF
--- a/transport_nantes/topicblog/tests.py
+++ b/transport_nantes/topicblog/tests.py
@@ -1,3 +1,10 @@
 from django.test import TestCase
 
 # Create your tests here.
+class SimpleTest(TestCase):
+
+    def test_main_page_status_code(self):
+        response = self.client.get("/")
+        return self.assertEqual(response.status_code, 200)
+    
+    


JeffAbrahamson:
What if the TopicBlog table is empty?

Shriukan33:
Noticed that the settings on master branch load urls_tn.py, instead of urls_m.py
If I change that, then indeed the topicblog table being empty, it wont be able to pick one at random (because there is none). 
The test I wrote is only worth because the class MainTransportNantes doesn't look for topics, as opposed to TopicBlogView. (--> loaded on the main page if we use the urls_m.py ). 
If there are none, according to how it looks, I would have to use TopicBlogViewTrivial as a view to get a 200 response.

JeffAbrahamson:
I think the right response is to make it into two tests (and maybe modify the code).  One version of the test assumes a record in TopicBlog, the other does not and checks that we get a 500.

Note that the test framework doesn't look at the real db, you set harnesses with the data you want in the db and test against that.

PR #145, opened 2021-06-01
Geoplan howto

JeffAbrahamson:


transport_nantes/geoplan/views.py
```diff
@@ -18,7 +18,9 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         city = kwargs["city"]
         observatory_name = kwargs["observatory_name"]
-        map_definition = MapDefinition.objects.get(observatory_name=observatory_name)
+        map_definition = MapDefinition.objects.filter(
```
Shriukan33, line 21: Should use get to make sure we only have 1 result, or raise an error

PR #157, opened 2021-06-11
Observatoire velo

Shriukan33:
Adds new map tiles. 
Adds proper colors on top of appropriate layers (Base, Satisfait et Non Satisfait).


JeffAbrahamson:
Mostly looks good.  Note that you're not rebased to origin/master, though the rebase is trivial and can be done at merge time.

In bc4bc1c, please reword the commit message to be imperative and not past tense.  Also comment the code to explain what each additional layer represents and why it's there.

In 31fd27e, again, reword to "Add" rather than "Added".
Concerning "We need a number of layers <= number of styles" -- what happens if we have fewer layers than that?

In 3e4c57d7, I think this is like "commune" in the survey module -- I'm not convinced this will always be a city.  (Consider Clement's suggestion to do a bicycle parking observatory at regional level.)  Perhaps reword the error message to be more generic, like (this observatory doesn't exist) and mention the city_name (which might not be a city).  Or something.

In c0e148fd, its not immediately clear on reading that you've used an enum.  Reformat to be more like [the docs](https://docs.djangoproject.com/en/3.2/ref/models/fields/#choices), with one pair per line and all-caps for the constant enum values.  This makes it easier to read and understand by making the formatting suggest what to expect.

You do two things in b184517: you change the colour array, which is great, but might this be cleaned up with the original introduction?  The other thing you do is map layer names to layer depths, but you re-introduce the strings, which means you've written these string names twice, introducing a possible future error source.  Also, you detect an error but don't report it.  I think this should 500 if this ever happens, no?

Shriukan33:
> Mostly looks good. Note that you're not rebased to origin/master, though the rebase is trivial and can be done at merge time.
> 
> In [bc4bc1c](https://github.com/transport-nantes/tn_web/commit/bc4bc1c43b8f2b851978a4f5ebb6383ea3ccb63b), please reword the commit message to be imperative and not past tense. Also comment the code to explain what each additional layer represents and why it's there.

Done

> In [31fd27e](https://github.com/transport-nantes/tn_web/commit/31fd27e2e1ee116e3ef6869b746ba62dd4dbbf24), again, reword to "Add" rather than "Added".
> Concerning "We need a number of layers <= number of styles" -- what happens if we have fewer layers than that?

This is now outdated, with the introduction of layer_type. Deleted old comment.

> In [3e4c57d](https://github.com/transport-nantes/tn_web/commit/3e4c57d77fa3395d58ca6f920ecb211fcbad6359), I think this is like "commune" in the survey module -- I'm not convinced this will always be a city. (Consider Clement's suggestion to do a bicycle parking observatory at regional level.) Perhaps reword the error message to be more generic, like (this observatory doesn't exist) and mention the city_name (which might not be a city). Or something.

Opted for a simpler sentence.  {city} ou l'observatoire {observatory_name} n'existe(nt) pas.

> In [c0e148f](https://github.com/transport-nantes/tn_web/commit/c0e148fdf50ba95f6c9929e65cf0d1cb1f583c06), its not immediately clear on reading that you've used an enum. Reformat to be more like [the docs](https://docs.djangoproject.com/en/3.2/ref/models/fields/#choices), with one pair per line and all-caps for the constant enum values. This makes it easier to read and understand by making the formatting suggest what to expect.

Made it easier to read, added CAPS as well.

> You do two things in [b184517](https://github.com/transport-nantes/tn_web/commit/b184517757a4b62fa2385e5328a52b1cf38590e9): you change the colour array, which is great, but might this be cleaned up with the original introduction? The other thing you do is map layer names to layer depths, but you re-introduce the strings, which means you've written these string names twice, introducing a possible future error source. Also, you detect an error but don't report it. I think this should 500 if this ever happens, no?

Cleaned original commit. Not sure about how to not write the value in the if condition though.

JeffAbrahamson:
Thanks, well done!

I've merged this after a rebase, so you'll need to delete the branch locally and remotely yourself.

About the rebase (which was also because I cherry-picked the filename commit last week):

I reworded a couple commits, mostly s/Adds/add/ and removing final periods on first lines.  Also reformatted aa52f37a for better readability.

In f9d8e338e your comment about the exception is a bit strange, but I let it pass.  Does this mean (asks the reasonable reader) it still needs addressing and that tests fail?  Does it mean that that's why you added it (but that for some reason it is only needed for the test but never for real life)?  Why isn't it a code comment then?

JeffAbrahamson:
And deployed.  Please do a quick check on prod that all is as you wish.

Shriukan33:
> In [f9d8e33](https://github.com/transport-nantes/tn_web/commit/f9d8e338e327b8f0946962f2ca185938e70a2c34) your comment about the exception is a bit strange, but I let it pass. Does this mean (asks the reasonable reader) it still needs addressing and that tests fail? Does it mean that that's why you added it (but that for some reason it is only needed for the test but never for real life)? Why isn't it a code comment then?

Actually it's required because it would throw error 500 when you try to get to an invalid URL instead of a classic 404. 
So yeah it makes the test pass, but also provides a better information in the case you fall on this page. 

> And deployed. Please do a quick check on prod that all is as you wish.

Looks ok, but one layer is still wrong in its content, (still in "non satisfait" instead of "satisfait")

PR #161, opened 2021-06-22
Add RAC

JeffAbrahamson:
Create a directory for resources we might want to make available.  For
now, we're obliged to do this with static files.  Later, we might move
this sort of thing to MEDIA.


Shriukan33:
asso_tn/static/ressources, is it where you want to store all files the users will be able to download somewhere on the website ? 
For instance, how would user access this one ?

JeffAbrahamson:
This one: `{% static %}/ressources/20210517_RAC.pdf` if we generate the link,
`/static/ressources/20210517_RAC.pdf` if we don't.

So that was kinda my idea, to have a path that leads to resources we want to make available for download.  Maybe we should call the directory "download" instead?

Shriukan33:
Download would be clearer in my opinion, resources can take many significations. 
Why in the asso_tn directory though ? 
Would we need to go to mobilitains.fr/l/resources/<filename> or something different ?

PR #168, opened 2021-08-10
Integration stripe clean

### LINKED ISSUE #54: Integrate stripe

Shriukan33:
* Add a donation page with a form.
* Track users progress in the form
* Save donations and their attached information (personal data, amounts, timestamp) into the database.

Closes #54 
Closes #167


JeffAbrahamson:
Could you add to `TrackingProgression` better documentation of what the fields mean?  Also, explain whether this is only used by the donation form and so view-only for admins or whether there's an intent that admins might do something.  (For example, you use `verbose_name`, which suggests human data entry.)

Also, adding `{% comment %}` blocks on `radio/checkbox_custom.html` briefly explaining why they're there would surely help some future code reader.

Are the styles added in e54a5ad specific to the donation form?  Perhaps then they should be named as such?

58717b5 is awesome.  Do you see how with a two line commit you were able to turn a feature on (and so off again, if needed)?

PR #176, opened 2021-08-25
Topicblog v2 create view edit

Shriukan33:
Adds the ability to create, edit, and view content for Topicblog.


JeffAbrahamson:
In 212b349 you should not repeat yourself.  State the block of the `urlpatterns` that is common, then conditionally add the part that is not.

JeffAbrahamson:
The commit comment in f881c274 is incorrect -- these are not default settings, they are the suggested settings for development environments.  As such, they don't belong in the commit comment but simply in the dev template (which you've done).

JeffAbrahamson:
In f881c274, why does the `static()` call in `transport_nantes/transport_nantes/urls.py` belong in production code?

JeffAbrahamson:
In be2b76c, why is it necessary to repeat so much of `base_mobilitains.html`?

JeffAbrahamson:
In 2ec417e, the string compositions would be clearer with an f"" string, no?

JeffAbrahamson:
In e2bee69, specify in the commit comment that the commit contains no non-formatting changes.  E.g, add the word "only".

Shriukan33:
> In [be2b76c](https://github.com/transport-nantes/tn_web/commit/be2b76ce8206ef6b66cced5a0d9d7699f9553481), why is it necessary to repeat so much of `base_mobilitains.html`? Why not just override `app_content` or `content`?

Copy/pasted too much on the original template, the only extra blocks we need to modify are content, app_content and styles. 

All other comments are dealt with ! :)

JeffAbrahamson:
In 6f631d6, I'm really confused.  There's a set of paths that are ok for prod, and they should be in the primary array.  And then there are a couple that are not ok for prod, that need tests and authentication controls to be deployed, and they should only be added if role is not prod.

JeffAbrahamson:
The bulk of the commit comments in 6f631d6 should be code comments: the part where you explain how the code works.  The commit comment should explain what change you are making and why.  It should elucidate something that is  part of the flow of time, whereas code comments should elucidate the actual state of the code.

Your explanation of the URL path intentions describes something that (if it is true) will remain true until changed.  It has nothing to do with the specific commit, it will be true in future commits as well.  In this particular case, these are mostly the comments that belong on the views.

Shriukan33:
> In [6f631d6](https://github.com/transport-nantes/tn_web/commit/6f631d61bd107fd2795833354ddc34d62ac929bd), I'm really confused. There's a set of paths that are ok for prod, and they should be in the primary array. And then there are a couple that are not ok for prod, that need tests and authentication controls to be deployed, and they should only be added if role is not prod.

Prod url pattern conflicts with dev utrl pattern, for example both have `'admin/new/` . Now this being said, those URL where here from the beginning, but not valid anyway, I just tried the admin/new in prod, returns a 500. 
Could I delete every incorrect url from the production url pattern ?

JeffAbrahamson:
In 9ffa053 your commit comment still notes certain suggested settings for `settings_local.py.template`, a file that doesn't exist.  That's wrong on two counts.

First, the file `settings_local.py.DEV.template` _is_ a set of suggestions.  If you want you could write that at the top of the file (not now, it would be a distraction) but don't put such things in commit comments, they are a distraction and they are not related to your reasoning in making the change.

The second reason it's wrong is that it is only a suggestion for development environments.  It is 100% wrong for beta or production.

JeffAbrahamson:
The commit message on 251239f again includes mostly information that should be in the doc string for the view template.

JeffAbrahamson:
In 251239f8, what specifically are you doing in `content.html`?  Why does this need to change the base template so extensively?  And why does it belong in this commit about view functionality?

You are creating a view (`TopicBlogItemView`) that is designed to do something that `TopicBlogLegacyView` already does, just more dynamically.  But the view is still providing all of the same content.  Why does the template need to change at all?

JeffAbrahamson:
Prod doesn't have `admin/new`.  That's functionality that you are currently developing and it doesn't currently exist in prod.

Shriukan33:
> The bulk of the commit comments in [6f631d6](https://github.com/transport-nantes/tn_web/commit/6f631d61bd107fd2795833354ddc34d62ac929bd) should be code comments: the part where you explain how the code works. The commit comment should explain what change you are making and why. It should elucidate something that is part of the flow of time, whereas code comments should elucidate the actual state of the code.
> 
> Your explanation of the URL path intentions describes something that (if it is true) will remain true until changed. It has nothing to do with the specific commit, it will be true in future commits as well. In this particular case, these are mostly the comments that belong on the views.

Moved most of the content in docstrings, did the same for the other commit. 
It's true that I should have documented that in the code instead of the commit comment.

Shriukan33:
> In [9ffa053](https://github.com/transport-nantes/tn_web/commit/9ffa053594c9b4c3508e549eac6a241382ebfea1) your commit comment still notes certain suggested settings for `settings_local.py.template`, a file that doesn't exist. That's wrong on two counts.
> 
> First, the file `settings_local.py.DEV.template` _is_ a set of suggestions. If you want you could write that at the top of the file (not now, it would be a distraction) but don't put such things in commit comments, they are a distraction and they are not related to your reasoning in making the change.
> 
> The second reason it's wrong is that it is only a suggestion for development environments. It is 100% wrong for beta or production.

I admit I'm not familiar with MEDIA files deployment on live servers yet
Modified the commit to specify this is only in dev.

Shriukan33:
> In [251239f](https://github.com/transport-nantes/tn_web/commit/251239f84cc30eeeb3c5ee99dcaa05c527317788), what specifically are you doing in `content.html`? Why does this need to change the base template so extensively? And why does it belong in this commit about view functionality?
> 
> You are creating a view (`TopicBlogItemView`) that is designed to do something that `TopicBlogLegacyView` already does, just more dynamically. But the view is still providing all of the same content. Why does the template need to change at all?

content.html is a boilerplate for tbv2 basically. Not so much has been overridden, it's only the style that has been adapted to use the new names (hero_image became header_image for example). 
Overall it's mostly about using the new names, and making a template that would be the base for all other contents, so it would replace 2020.html eventually.

Shriukan33:
> Prod doesn't have `admin/new`. That's functionality that you are currently developing and it doesn't currently exist in prod.
> 
> I mean, I see that that path somehow made it into master, but what's it doing there, really? Wasn't that a bug do merge it without it working? (Check annotate, it was introduced with [ab3f8c8](https://github.com/transport-nantes/tn_web/commit/ab3f8c8dc5dac0a2370b238dd7274ef8edf59e00). It clearly just got missed at code review when being pushed to master.)

Prod has it in its URL pattern since ab3f8c8dc5dac0a2370b238dd7274ef8edf59e00 but it's never been actually used. 
You can go on https://www.mobilitains.fr/tb/admin/new/ and throw a 500 error...

It's just that I've been reluctant to delete chunks of your code, I probably should have in this specific case.

Shriukan33:
In this push I've just modified the URL pattern with incorrect paths we mentioned in the previous comments

JeffAbrahamson:
In c5fd2be, you're adding a media role for non-dev environments.  That line is only valid for dev.

JeffAbrahamson:
In 3cc6862, why do you need new field names in `content.html`?  This is leading you to duplicate quite a lot of base.  The output of `s/` and `t/` should be (almost) identical, why does this require a content template that differs much from `2020.html`?

JeffAbrahamson:
In 3cc6862, why don't you use `get_object_or_404()`?

Shriukan33:
> In [c5fd2be](https://github.com/transport-nantes/tn_web/commit/c5fd2be7df9fd72531a5cb84a69194d780b0b863), you're adding a media role for non-dev environments. That line is only valid for dev.

Could you expand on this, I don't understand what line you're referring to

JeffAbrahamson:
> > In [c5fd2be](https://github.com/transport-nantes/tn_web/commit/c5fd2be7df9fd72531a5cb84a69194d780b0b863), you're adding a media role for non-dev environments. That line is only valid for dev.
> 
> Could you expand on this please ? I don't understand what line you're referring to.

`MEDIA_{ROOT,URL}` and `STATIC_{ROOT,URL}` are defined per environment.  They are not even necessarily served from the same host on beta or production.  You shouldn't add paths for them except for dev.  (Understand why.)

Shriukan33:
> In [3cc6862](https://github.com/transport-nantes/tn_web/commit/3cc6862b8a8cd223701e3d5a0d2ffb82dfe284b2), why do you need new field names in `content.html`? This is leading you to duplicate quite a lot of base. The output of `s/` and `t/` should be (almost) identical, why does this require a content template that differs much from `2020.html`?
> 
> I'm not saying 100% that this is wrong, just that it's not obvious. If you have to do this, it merits comment in the commit log. Just "to fit new field names" sounds like you're refactoring, and then the reader doesn't really understand.
> 
> I think the issue is that you wanted to use media tags instead of static tags. (Careful, just like you need to use `{% static ...`, you need to use `{% media ...` or something like that.)

Because if you check on base_mobilitain, the conditional style uses "page.hero_image" 
<details> <summary>See lines</summary>

https://github.com/transport-nantes/tn_web/blob/44a03438fddad21cad4a4494ca7e4189f0c76398/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html#L70
https://github.com/transport-nantes/tn_web/blob/44a03438fddad21cad4a4494ca7e4189f0c76398/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html#L78
</details>

The reason it  works for TopicBlogLegagacyView is because it's based on TopicBlogPage model which use the hero_image field
<details><summary>See line</summary>

https://github.com/transport-nantes/tn_web/blob/44a03438fddad21cad4a4494ca7e4189f0c76398/transport_nantes/topicblog/models.py#L71</details>

While in TB v2, we now use a header_image field for the same purpose
<details> <summary>See line</summary> 

https://github.com/transport-nantes/tn_web/blob/c3edf34247a1090424dc60d2355657ca44860ee0/transport_nantes/topicblog/models.py#L225
</details>

So importing base mobilitain wouldn't work because the field used for the condition wasn't appropriate for TopicBlogItem. 
The same story follows for few other fields like hero_title that became header_title, 
That's why I did append to the  blocks these modifications : 
<details> <summary> See the modifications </summary>

<h3> Style block </h3>

https://github.com/transport-nantes/tn_web/blob/c3edf34247a1090424dc60d2355657ca44860ee0/transport_nantes/topicblog/templates/topicblog/content.html#L28

https://github.com/transport-nantes/tn_web/blob/c3edf34247a1090424dc60d2355657ca44860ee0/transport_nantes/topicblog/templates/topicblog/content.html#L36
<h3> Content block </h3>

https://github.com/transport-nantes/tn_web/blob/c3edf34247a1090424dc60d2355657ca44860ee0/transport_nantes/topicblog/templates/topicblog/content.html#L55

https://github.com/transport-nantes/tn_web/blob/c3edf34247a1090424dc60d2355657ca44860ee0/transport_nantes/topicblog/templates/topicblog/content.html#L56

<h3>App content block</h3>

This :
https://github.com/transport-nantes/tn_web/blob/c3edf34247a1090424dc60d2355657ca44860ee0/transport_nantes/topicblog/templates/topicblog/2020.html#L20

Became :
https://github.com/transport-nantes/tn_web/blob/c3edf34247a1090424dc60d2355657ca44860ee0/transport_nantes/topicblog/templates/topicblog/content.html#L64
</details>

The replacement for `{% static 'my _image' %}` is `{{ page.my_image.url }}` in my commit. 

I hope I managed to get it clearer, not sure how to explain that amount of details in the commit.

PR #182, opened 2021-08-31
Topic blog v2 list

Shriukan33:
Add a List view to TopicBlog

At allows the user to see a list of all TopicBlogItems OR filter the results to be limited to those containing a specific slug. 

Follows PR #176


JeffAbrahamson:
LGTM

PR #183, opened 2021-08-31
Tb v2 tests

Shriukan33:
Add tests to each view in the URL pattern : 
- TopicBlogItemEdit
- TopicBlogItemView
- TopicBlogItemList

it tests status code, the item_sort_key of the loaded item and the length of queries when appropriate. 

Follows PR #182


JeffAbrahamson:
LGTM

Issue #192, opened 2021-09-19
Add test info to stripe pages in dev/beta


JeffAbrahamson:
Add a reminder on all stripe pages (that we generate):
* that we're in test mode
* what numbers to provide to trigger success
* what numbers to provide to trigger failure
* link to stripe developer docs on how to interact with test mode (i.e., the above info and a bunch more that I remember is there)

PR #204, opened 2021-09-21
One euro donation btn 198

### LINKED ISSUE #198: one euro donation button
For this weekend we want a button we can deploy on the home page that asks the user for **one euro** and on the next line gushes about how swell it would be to offer an additional contribution.  Plus all the lovely monthly stuff, too.

Shriukan33:
Ready to merge

This PR also includes the branch integration-design

Allows users to land on a simplified donation form where the amount is already set. 
You can access it by going to  /donation/<int:amount>

better texts are required to deploy. 

Closes #198


JeffAbrahamson:
305645ee - this is wrong.  We shouldn't modify templates for every single type of button we have, these should be template tags so that they can be inserted via markdown.  I can give you a hand with how to modify the markdown parser, but you should make a templatetag like the `{% don_bouton %}` etc.

JeffAbrahamson:
40fc914 - "one time payments" and "one-time payments" are not the same thing.  You need a hyphen. ;-)

"Quick" might be the wrong term, since one could easily imagine other ways we will make donations quicker.  But not important for now.

JeffAbrahamson:
2cd215bbf - please don't commit print statements.  If it deserves to outlast your development, make it a log statement and then make it clear enough to be understood in the logs.  Otherwise, remove before commit (even if you just add it again after commit for yourself and continued development).

Also, getting a zero isn't a non-numeric element, though it could perhaps be the result of a string to int conversion.  Make sure your error messages are correct, you can waste hours chasing what your error message told you only to find out painfully later that your error was something else.

In `QuickDonationView` you are acting _very_ confident that those keys will be in `kwargs`.  Are you sure that it's absolutely impossible for the world to be otherwise?  You will 500 if you get this wrong.

JeffAbrahamson:
d3f909237f - the tax reduction in france is 66%, which is not precisely 1/3.

JeffAbrahamson:
d38ff65d - for #192, could you please display this information on all donation pages if role is not production?  The person who needs it is not necessarily looking at github at that moment, nor remembering where you put this.  They _are_ looking at a donation page.  Putting information at point of need is optimal.

Shriukan33:
> [305645e](https://github.com/transport-nantes/tn_web/commit/305645ee4e02738757e31172d364bcad6d3cff3a) - this is wrong. We shouldn't modify templates for every single type of button we have, these should be template tags so that they can be inserted via markdown. I can give you a hand with how to modify the markdown parser, but you should make a templatetag like the `{% don_bouton %}` etc.

Done, now uses the {% set_amount_donation 1 %} template tag.

Shriukan33:
> [d38ff65](https://github.com/transport-nantes/tn_web/commit/d38ff65d9eb09b16067270716456be1f29749932) - for #192, could you please display this information on all donation pages if role is not production? The person who needs it is not necessarily looking at github at that moment, nor remembering where you put this. They _are_ looking at a donation page. Putting information at point of need is optimal.

Added a rectangle in the bottom part of the donation forms. 
It includes the links in the readme :slightly_smiling_face:

JeffAbrahamson:
4c20da6f1 - the name `set_amount_donation` is too wrapped up in your current mental state.  Think of the person coming to this without any context.  Does this function set the amount of a donation?  Perhaps call this `stripe_fixed_plus_free` ?

What is a "link banner"?  I've done searches on "link banner", "html link banner", and "web link banner", and I'm now convinced this term doesn't exist.

I believe, from the code, that you are creating button.

JeffAbrahamson:
0ff7bb0 -- learn about """ (triple quotes) in python please.

If text will be rendered in html, call it `motivation_html`.  If it will be passed through markdown, then `motivation_md`.  When I see `_text`, I understand that it is of type "text/plain".

JeffAbrahamson:
988e414 - on those "return int(float(...))" I'd be inclined to have a well-named intermediate variable that helps the reader understand what you're doing.  E.g.,

```
payment_amount_euros = float(items["payment_amount"])
payment_amount_centimes = int(100 * payment_amount_euros)
return payment_amount_centimes
```

JeffAbrahamson:
93c61cf - README is good, but when you render it on the web page, drop the md and just do user-readable html.  Think user, what does the testing user need?  I think it's just this, and in order of importance:

* reassurance that we're in test mode
* card number to succeed
* card number to fail
* pointer to stripe_app/README (link to github page) for more info

JeffAbrahamson:
e1e43ad - I don't feel qualified to review this.  I'll just comment that you've got some really long lines that make readability harder than necessary.

Also, I'm not sure it's right to put user visible text in the javascript.  So on the `real_cost` computation, why not use a named `<span>` to isolate the number so that the js is only manipulating that number and not the text around it?  That way django remains in charge of content.

I think you could improve your comments.  E.g., rather than "Event to trigger ..." (we already know it's an event trigger), write "Calculate and update the tax deduction amount when donation amounts change", or something like that.

JeffAbrahamson:
Read more about how to comment code and commits. The web is full of advice, not all of it right.  Which is why you need to read a lot and get a lot of experience...

For example your commit message on 988e414 says "this view is meant to ...".  Things that jump out:
* "is meant to" -- what subtlety is intended, that you have a doubt?
* if you want to explain the view, that's a code comment
* the commit comment should talk about what's changed, not what the code does

JeffAbrahamson:
> > [2cd215b](https://github.com/transport-nantes/tn_web/commit/2cd215bbfb22a8805857341c3eb331815d4b7667) - [...]
> 
> > In `QuickDonationView` you are acting _very_ confident that those keys will be in `kwargs`. Are you sure that it's absolutely impossible for the world to be otherwise? You will 500 if you get this wrong.
> 
> The only expected kwarg is "amount".
> If no amount is provided in the kwargs, then it's the normal donation page that is fired, it's a different URL path.

"Expected" and "guaranteed" are not the same thing. ;-)
Are you prepared to 500 for your expectations?

Shriukan33:
> [4c20da6](https://github.com/transport-nantes/tn_web/commit/4c20da6f1c29f63dbd30063766aef7703565dcde) - the name `set_amount_donation` is too wrapped up in your current mental state. Think of the person coming to this without any context. Does this function set the amount of a donation? Perhaps call this `stripe_fixed_plus_free` ?
> 
> What is a "link banner"? I've done searches on "link banner", "html link banner", and "web link banner", and I'm now convinced this term doesn't exist.
> 
> I believe, from the code, that you are creating button.

Since that comment, I broke it into 2 tags, 
https://github.com/transport-nantes/tn_web/blob/71f9023d5f511615d26498d8e4a28d888c8c4234/transport_nantes/asso_tn/templatetags/don.py#L79
and
https://github.com/transport-nantes/tn_web/blob/71f9023d5f511615d26498d8e4a28d888c8c4234/transport_nantes/asso_tn/templatetags/don.py#L94

The first creates a combination of a ""banner"" (don't know how you call a message on a blue stripe of the width of the screen) which contains a message, and a button leading to a simplified donation page. 

The 2nd only creates a <a> tag which looks like donation button, the link redirects to a simplified donation form as well. It's the same tag called in the previous one.

Shriukan33:
> [0ff7bb0](https://github.com/transport-nantes/tn_web/commit/0ff7bb06714340a2e63af7db82065df1ee0454ae) -- learn about """ (triple quotes) in python please.
> 
> If text will be rendered in html, call it `motivation_html`. If it will be passed through markdown, then `motivation_md`. When I see `_text`, I understand that it is of type "text/plain".

I know about """, I probably should have used it from the start, corrected anyway. 

renamed to motivation_html.

Shriukan33:
> [988e414](https://github.com/transport-nantes/tn_web/commit/988e4146209c861f188f09eea656ac792f453109) - on those "return int(float(...))" I'd be inclined to have a well-named intermediate variable that helps the reader understand what you're doing. E.g.,
> 
> ```
> payment_amount_euros = float(items["payment_amount"])
> payment_amount_centimes = int(100 * payment_amount_euros)
> return payment_amount_centimes
> ```

Fixed : Used clearer names
image

Shriukan33:
> [93c61cf](https://github.com/transport-nantes/tn_web/commit/93c61cf9a3407893cda99ff4706d7470b6a7d822) - README is good, but when you render it on the web page, drop the md and just do user-readable html. Think user, what does the testing user need? I think it's just this, and in order of importance:
> 
>     * reassurance that we're in test mode
> 
>     * card number to succeed
> 
>     * card number to fail
> 
>     * pointer to stripe_app/README (link to github page) for more info

Updated it in the meantime.
It now specifies the test mode, and the credit cards in first. 
Also added some information regarding the fields to fill.

Shriukan33:
> [e1e43ad](https://github.com/transport-nantes/tn_web/commit/e1e43ad990989a45f07ccfedc084a64d88c2c0d4) - I don't feel qualified to review this. I'll just comment that you've got some really long lines that make readability harder than necessary.
> 
> Also, I'm not sure it's right to put user visible text in the javascript. So on the `real_cost` computation, why not use a named `<span>` to isolate the number so that the js is only manipulating that number and not the text around it? That way django remains in charge of content.
> 
> I think you could improve your comments. E.g., rather than "Event to trigger ..." (we already know it's an event trigger), write "Calculate and update the tax deduction amount when donation amounts change", or something like that.

I'm definitely less comfortable with JS than I am with Python. 

Regarding the content part, you're right that text is displayed by Javascript instead of Django, however the calculation wouldn't make sense if we removed it from django, while I can delete this part from JS without having any sort of impact on Django, I don't know it that makes sense.

Shriukan33:
> Read more about how to comment code and commits. The web is full of advice, not all of it right. Which is why you need to read a lot and get a lot of experience...
> 
> For example your commit message on [988e414](https://github.com/transport-nantes/tn_web/commit/988e4146209c861f188f09eea656ac792f453109) says "this view is meant to ...". Things that jump out:
> 
>     * "is meant to" -- what subtlety is intended, that you have a doubt?
> 
>     * if you want to explain the view, that's a code comment
> 
>     * the commit comment should talk about what's changed, not what the code does

Reworded

Shriukan33:
> > > [2cd215b](https://github.com/transport-nantes/tn_web/commit/2cd215bbfb22a8805857341c3eb331815d4b7667) - [...]
> > 
> > 
> > > In `QuickDonationView` you are acting _very_ confident that those keys will be in `kwargs`. Are you sure that it's absolutely impossible for the world to be otherwise? You will 500 if you get this wrong.
> > 
> > 
> > The only expected kwarg is "amount".
> > If no amount is provided in the kwargs, then it's the normal donation page that is fired, it's a different URL path.
> 
> "Expected" and "guaranteed" are not the same thing. ;-)
> Are you prepared to 500 for your expectations?

My URL pattern is the following, regarding the donation form : 
Either it has no args, then it's a normal donation form. 
Or it has an int as an arg and then it's a QuickDonationForm. 
If it's anything else, it 404 rightfully. 

This view is only called when an int is given, if there is not, this is a normal donation form or a 404 (if given anything else than an int.)
Did you spot a problem I didn't see in this pattern ? :open_mouth:

JeffAbrahamson:
> > [e1e43ad](https://github.com/transport-nantes/tn_web/commit/e1e43ad990989a45f07ccfedc084a64d88c2c0d4) - I don't feel qualified to review this. I'll just comment that you've got some really long lines that make readability harder than necessary.
> > Also, I'm not sure it's right to put user visible text in the javascript. So on the `real_cost` computation, why not use a named `<span>` to isolate the number so that the js is only manipulating that number and not the text around it? That way django remains in charge of content.
> > I think you could improve your comments. E.g., rather than "Event to trigger ..." (we already know it's an event trigger), write "Calculate and update the tax deduction amount when donation amounts change", or something like that.
> 
> I'm definitely less comfortable with JS than I am with Python.
> 
> Regarding the content part, you're right that text is displayed by Javascript instead of Django, however the calculation wouldn't make sense if we removed it from django, while I can delete this part from JS without having any sort of impact on Django, I don't know it that makes sense.

At a high level, I think what you're doing is looking up an element in the DOM and changing its value.  That element here is a paragraph or sentence or something.  I'm suggesting that that element could be precisely the number you want to change.  That's what `<span>` is for.  It's the horizontal version of `<div>`.

JeffAbrahamson:
> > > > [2cd215b](https://github.com/transport-nantes/tn_web/commit/2cd215bbfb22a8805857341c3eb331815d4b7667) - [...]
> > > 
> > > 
> > > > In `QuickDonationView` you are acting _very_ confident that those keys will be in `kwargs`. Are you sure that it's absolutely impossible for the world to be otherwise? You will 500 if you get this wrong.
> > > 
> > > 
> > > The only expected kwarg is "amount".
> > > If no amount is provided in the kwargs, then it's the normal donation page that is fired, it's a different URL path.
> > 
> > 
> > "Expected" and "guaranteed" are not the same thing. ;-)
> > Are you prepared to 500 for your expectations?
> 
> My URL pattern is the following, regarding the donation form :
> Either it has no args, then it's a normal donation form.
> Or it has an int as an arg and then it's a QuickDonationForm.
> If it's anything else, it 404 rightfully.
> 
> This view is only called when an int is given, if there is not, this is a normal donation form or a 404 (if given anything else than an int.)
> Did you spot a problem I didn't see in this pattern ? open_mouth

I may have just been confused.  Perhaps a code comment that makes clear why your assumption is safe (i.e., that you have two URL patterns and that you therefore can't not have this value).  And a test that will fail if a URL pattern some day changes and the assumption becomes false. ;-)

The thing is that assumptions now are not necessarily true tomorrow, and so it's important to document why we believe what we believe so that that the day it's false, it's (relatively) easy to figure out what assumption was violated.

PR #212, opened 2021-09-28
Code refactoring.

### LINKED ISSUE #209: Use temporary to avoid repeated same element lookups
Python dict lookups are modestly expensive.  If you're going to do the same lookup repeatedly, use a temporary variable.

It also makes your code more readable.

This is in reference to `transport_nantes/stripe_app/views.py`, lines 338 ff.

Shriukan33:
Closes #209

Use a variable to store metadata and improve code readability.


Shriukan33:
Implemented in #213 PR

PR #213, opened 2021-09-28
Tracking bis 205

### LINKED ISSUE #205: Donations should record the view and eventually parameters that caused them
This is two extra columns in the donation table: `originating_view` and `originating_parameters`.

Both should start nullable.  We'll back-fill the old values and then make `originating_view` non-nullable later.

Shriukan33:
Add originating view and parameters in the Donation entry. 

The Originating view variable is set in the view and passed through context in the <meta> tags. 
JS then checks the value of this tag and acts accordingly. 

Closes #205 
Includes and closes #212


JeffAbrahamson:
I'm merging this, because we have to move on.  But `form_flow.js` is really poorly commented.  I'd wager you'll have major trouble reading your own code here a few months from now.

The term "standard donation" is really wrapped up in your current thoughts.  "Standard" relative to what?  It just happens to be the first donation view you wrote, that gives it no special rights, no special status.  The only name it's been given is "StripeView", which is arbitrary but as good as any other, and has the good grace to match the View class name.  That's good enough.

JeffAbrahamson:
I'm also more than a bit concerned that you're presenting an alert (to whom?  to unsuspecting visitors?) suggesting that we've configured something wrong.  We have, but it's not their problem.  We should contact ourselves.  This is code, it can do that.  Send yourself mail.  Maybe we even have an address set up to do that.  Or write something to a table of things to get to some day.  Or log it.  It depends how urgent it is that it get fixed.

A user can't possibly know the urgency.  Should they call you?  Is it worth waking you?  Should they escalate?  Maybe tweet it to the world with @mobilitains.fr tagged because they don't know how to contact the administrator, whose contact details are not provided with the error message?

Shriukan33:
I went for a very generic message, because this alarm shouldn't be set outside of dev environnement, at worse beta. This is only a trigger to say "Hey, you did a typo in the View name". 
Maybe the contact the admin was too much, given that the trigger to that error shouldn't happen in prod. 

You asked a else statement in case neither of the views were the originating view. Opted for this, should be enough to warn us something is wrong before going to prod. 
However I'll take any suggestion to improve this, as it's extremely basic. :slightly_smiling_face:

JeffAbrahamson:
And if it will only execute in dev || beta, then it's enough to give a tech message, like "missing case in stripe donation form if/then/else".

Shriukan33:
Nothing stops it to fire in prod, except if we didn't try to look at the donation page in dev and beta. Now, I could make a check to see if we're not in prod, and only then fire the alarm ? 
I'd place the role in the meta tags as a value to check.

JeffAbrahamson:
Perhaps make a function that alerts in dev/beta but mails in prod?

PR #216, opened 2021-09-28
Delete extra head and Doctype tags from geoplan.

### LINKED ISSUE #193: Fix invalid HTML on OV page
The [OV page](https://beta.mobilitains.fr/observatoire/nantes/velopolitain) still has invalid html.

image

Make a new django project as a MRE for SO and/or the package site, push the repo with a good test name, and ask on SO or as an issue on the package's site, whichever seems more appropriate.  (Look for evidence of questions with answers to make your decision.)

In asking the question, provide context and code in your question -- don't require the reader to go to the repo.  It's there only as backup.  In addition, it's not part of the question site, so it's good practice not to make the question depend on something so exterior.

Shriukan33:
Although they didn't seem to be problematic for now,
we don't want any sort of issue because of these extra tags
Better safe than sorry.

Closes #193


JeffAbrahamson:
This isn't the right way to do this.  We should understand why it's happening and fix that.

PR #217, opened 2021-10-04
Add a sentence to remind user what amount it's giving.

### LINKED ISSUE #180: Page 2 of the donation form should echo current status
On the second page of the payment pipeline, we don't remind the user what they're in the process of confirming.  (Stripe does at step 3.)  But this contributes to a feeling of confidence, which encourages completion, so we should say something up top that summarises it: "a one-time donation of X €" or "A monthly donation of X €".  With better sauce, as we'd say in French...

In doing this, we should make sure we keep in mind that "monthly" might change, so the two forms need to communicate with each other about what, exactly, the product is that the user is choosing.

Shriukan33:
Goal is to support completion of the form, and provide
feedback to user.
Closes #180

### DIFF
--- a/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
+++ b/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
@@ -420,3 +420,10 @@ span.centered {
 #card_velorution_2:hover {
     box-shadow: 7px 1px 16px -1px rgba(212, 212, 212, 0.30);
 }
+#recall_amount {
+    font-weight: bold;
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
I really think you're asking for future maintenance pain by creating a blank `<p>` and using javascript to manage its contents.  It would be more maintainable to make the paragraph be what you want it to be and just use a named `<span>` for the amount, initially set to zero or whatever the donation is initially set to.

Maybe you need to go through that pain to discover how bad it will be... ;-)

PR #244, opened 2021-10-19
TBv2 Feedbacks

### LINKED ISSUE #239: TBv2 item submit feedback
On creating a new item or modifying an existing one (or creating a variant), the submit should provide feedback.  Probably the thing to do is to provide a page with the following:
* Confirmation that item is saved
* LInk to item
* Link to edit the item again
* Link to create a new item

Providing the new item form at the same time is likely confusing.

Shriukan33:
Add confirmation banner to confirm that the item has been successfully created
Add links to edition and view on the confirmation message.

Closes #239

image


JeffAbrahamson:
Can you redo the commit message so that it stays under 50 characters or so?  It's hard to view a short log.

Shriukan33:
image
Does it feel better to you ?

Shriukan33:
About the share thing, what do you mean ? Like copy to clipboard ?

JeffAbrahamson:
The link to re-edit is the `admin/edit/` link with slug and pk id.  That should 403 for someone not logged in or without permission.  So providing a slug-only link would be a good move.

Not important just now, we can do that later.

PR #248, opened 2021-10-21
Add recurring payments handling for subscriptions.

### LINKED ISSUE #233: Make sure that stripe webhook processes recurring donations
Need to make sure that the webhook handles recurring donations so proper history is kept.

Shriukan33:
Recurring payments were ignored, they now are saved.

closes #233


JeffAbrahamson:
Here you should positively copy the fields you want, even if there are a lot of them, rather than negatively delete the ones you don't.  The negative approach is fragile against new fields that perhaps would cause damage by propagating through.  (Consider as an innocent example a timestamp from the old event.)  If you forget to propagate a field, it may suck but you're unlikely to copy something sensitive that you shouldn't.

```
+    last_donation_kwargs = customer.__dict__
+    try:
+        # Delete two unnecessary fields to create a new donation entry.
+        del last_donation_kwargs["_state"]
+        del last_donation_kwargs["id"]
+    except KeyError as e:
+        logger.debug("KeyError while deleting _state and id")
+        logger.debug(last_donation_kwargs)
+        logger.debug(e)
+        pass
```

PR #255, opened 2021-10-25
Fix borked dropwdown on edition form.

### LINKED ISSUE #250: Editing a TBItem borks the content type and template
The content type is set to "article" (please verify this is set from the db and not by javascript).  The template name is not set at all.  (It must be set from the db and not by javascript.)

Shriukan33:
Splitting the form into two classes avoid the former default behavior of
the "template" field to be empty when the form is loaded.

Closes #250

### DIFF
--- a/transport_nantes/topicblog/forms.py
+++ b/transport_nantes/topicblog/forms.py
@@ -1,9 +1,12 @@
 from django.forms import ModelForm
-from django.forms.models import ModelChoiceField, ModelMultipleChoiceField
+from django.forms.models import ModelChoiceField
 from .models import TopicBlogItem, TopicBlogTemplate
 
 
 class TopicBlogItemForm(ModelForm):
+    """
+    Generates a form to create and edit TopicsBlogItem objects.
+    """
 
     class Meta:
         model = TopicBlogItem
@@ -12,8 +15,9 @@ class TopicBlogItemForm(ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.fields['template'] = ModelChoiceField(
-            queryset=TopicBlogTemplate.objects.none())
+        if self.fields["content_type"] is None:
+            self.fields['template'] = ModelChoiceField(
+                queryset=TopicBlogTemplate.objects.none())
 
         # self.data is the POST data
         if 'template' in self.data:
--- a/transport_nantes/topicblog/static/topicblog/update_template_list.js
+++ b/transport_nantes/topicblog/static/topicblog/update_template_list.js
@@ -28,7 +28,7 @@ async function update_template_list() {
 }
 
 function select_first_item() {
-  $('#id_template option[value="1"]').attr('selected', 'selected');
+  $("#id_template")[0].selectedIndex = 1;
 }
 
 $("#id_content_type").change(async () => {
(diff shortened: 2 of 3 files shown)


JeffAbrahamson:
Although, disabling modifying the slug only happens once we publish.  Editing something unpublished leaves the slug modifiable.

JeffAbrahamson:
There's something fundamentally awry here, and I'm not sure exactly what it is.  If you put a print statement at the beginning of each view and form function and at the key branch points for the new and edit views, you see this:
* `admin/new` - `TopicBlogItemNewForm` -> `TopicBlogItemEdit.get_context_data()` -> else branch (`pk_id <= 0`) -> `TopicBlogItemNewForm`
* `admin/edit` - `TopicBlogItemNewForm` ->  `TopicBlogItemEdit.get_context_data()` -> primary branch (`pk_id > 0`) -> `TopicBlogItemEditForm`

Note that `FormView`, as you might imagine from its name and the fact that you already told it what form you want, creates the form for you.

So at the very least we're being very inefficient by creating and recreating forms, and not always the forms we intend eventually to use.  But this could get worse without much warning.  Imagine some day we put some interesting functionality in one or both of these forms, and now they're both being created!

In addition, this means you're doing too much work, because django is really trying to save you effort by providing convenient hooks for stuff.  Assuredly you're trying to do something at the wrong place and so having to compensate for that.

I'd suggest you read the source and really understand how `FormView` works.  You surely already have the habit of checking source, but note that the relevant file is at `lib/python3.9/site-packages/django/views/generic/edit.py`.  If that's not obvious, let's make it obvious -- having source code available is too valuable not to have checking it be second nature.

As you can see, having two forms in a `FormView` is against the laws of nature...

Shriukan33:
Updated with a simpler solution :) 

I made sure to call get_context_data in any case but without initializing a form for naught, by reading carefully how FormView works ;) 

Also minor but fixed a bug in the JS, it indeed selected the first item ... as long as its PK was 1 ! Now it selects the 1st option in the dropdown.

JeffAbrahamson:
Why do you need not to call `super().get_context_data(**kwargs)` at the top but only after setting `form`?

Shriukan33:
> Why do you need not to call `super().get_context_data(**kwargs)` at the top but only after setting `form`?

Looking into how FormView works, I noticed that the view will set the context["form"] variable to be the one defined in form_class variable (aka TopicblogForm). However it will only initialize this action if no 'form' argument is passed to the get_context_function.

So to avoid the FormView to intialize a form just for it to be immediately overridden by the next statement, I initialize myself the TBIForm with the instance I want, and pass it as an argument to the get_context_function. Doing so, the FormView sees there is a form argument, and doesn't create a new TBIForm instance. However the context is still passed through the fonction and through the get_context_data (gdc) of ContextMixin. This way my gcd function  still goes through django machinery, and spare django from creating an useless instance.

In the case we don't have an instance (a New object), we just call the gdc function without the form argument, and django does its thing : it creates a new instance of the TBIForm normally, as the FormView does when no form is given as argument.

In both cases, our context goes through django gdc functions. In the case where we edit an existing object kwargs will have 'form' : TBIForm(instance=pk), otherwise kwargs will be empty.


Does it make sense to you? Sorry for format I'm typing this on mobile without the code to back this up

PR #272, opened 2021-11-03
Allow CTA labels to render HTML.

Shriukan33:
Goal is to render arrows and various symboles in the CTA labels.

image

image

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/content.html
+++ b/transport_nantes/topicblog/templates/topicblog/content.html
@@ -88,7 +88,7 @@
           {# First CTA, if any #}
           {% if page.cta_1_slug|length > 0 %}
                <a href="{% url 'topic_blog:view_item_by_slug' page.cta_1_slug %}" 
-               class="btn donation-button my-3 centered-inline-button">{{ page.cta_1_label }}</a>
+               class="btn donation-button my-3 centered-inline-button">{{ page.cta_1_label|safe }}</a>
           {% endif %}
 
           {# Second paragraph #}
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
I think the problem you are trying to solve is "I want arrows on my CTA buttons".

Your solution is to mark all CTA's safe.  I presume you intend this for all templates, though you didn't modify any other templates, so our CP's, for example, would emit raw (quoted) html in this scenario.

Why is this the right solution?  It would certainly appear to open us to html insertion attacks on CTA fields.  What controls have you thought to put in place to avoid that?

In addition, this requires TB users to know how to add that html, which seems like a stretch for non-technical users.  It also seems kind of error-prone even for the competent among us, for example open to typo's.

Are there CTA's that should _not_ have an arrow?  I think the answer is no for the template in question.  Why is the correct solution then not to add the font-awesome directly to the template?  If some template wants to offer buttons without, that's the template's choice.

Shriukan33:
> I think the problem you are trying to solve is "I want arrows on my CTA buttons".
> 
> Your solution is to mark all CTA's safe. I presume you intend this for all templates, though you didn't modify any other templates, so our CP's, for example, would emit raw (quoted) html in this scenario.
> 
> Why is this the right solution? It would certainly appear to open us to html insertion attacks on CTA fields. What controls have you thought to put in place to avoid that?
> 
> In addition, this requires TB users to know how to add that html, which seems like a stretch for non-technical users. It also seems kind of error-prone even for the competent among us, for example open to typo's.
> 
> Are there CTA's that should _not_ have an arrow? I think the answer is no for the template in question. Why is the correct solution then not to add the font-awesome directly to the template? If some template wants to offer buttons without, that's the template's choice.

You're right it's a stretch, there must be a better solution. Maybe something along the lines of what's been done for Markdown custom tags ? Like [[arrow-right]] ?

JeffAbrahamson:
I think the right solution is PR #277 .  
I'm going to close this PR.

PR #270, opened 2021-11-02
Add @StaffRequired decorator to secure Ajax endpoints

Shriukan33:
This decorator is used to secure Ajax endpoints.

With this decorator, only staff users will be able to access these
endpoints.

Closes #268

PR #273, opened 2021-11-03
Add a message under the Slug Field in TBv2

### LINKED ISSUE #246: Warn on existing slug creation in TBI editor
If the user is creating or modifying a TBItem and modifies a slug (including from empty), we should use ajax to signal that it's a duplicate if it is (and isn't blank).  This isn't outlawed, but it would mean the pages should be interchangeable.  If the slug exists, we should offer a button that opens (in a new tab/window) a selector for all pages with that slug.

Ideally, query per keystroke, like search engines do (did) since Google introduced Instant Search back in 2010 (and then deep sixed because it makes no sense on mobile).  No one noticed, because they now just do promoted search...

Shriukan33:
Now OK to merge

This message informs the user if the slug its entering is already
in use.

The goal is to not use slugs that have already been used unknowingly.

Closes #246

image

image

Ajax endpoints have been secured with @StaffRequired decorator (see #270)

```cmd
benja@Benja-desktop:~/Documents/GitHub/tn_web/transport_nantes$ curl -I localhost:8000/tb/ajax/get-slug-dict/
HTTP/1.1 403 Forbidden
Date: Wed, 03 Nov 2021 13:20:58 GMT
Server: WSGIServer/0.2 CPython/3.8.10
Content-Type: text/html; charset=utf-8
X-Frame-Options: DENY
Content-Length: 68
Vary: Cookie
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Set-Cookie:  sessionid=un4xiufkkbnb4kuy5za09wzb4f7sugfr; expires=Mon, 02 May 2022 13:20:58 GMT; HttpOnly; Max-Age=15552000; Path=/; SameSite=Lax

benja@Benja-desktop:~/Documents/GitHub/tn_web/transport_nantes$ curl -I localhost:8000/tb/ajax/get-url-list/
HTTP/1.1 403 Forbidden
Date: Wed, 03 Nov 2021 13:21:14 GMT
Server: WSGIServer/0.2 CPython/3.8.10
Content-Type: text/html; charset=utf-8
X-Frame-Options: DENY
Content-Length: 68
Vary: Cookie
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Set-Cookie:  sessionid=nu2bic1q9yxz60mu8uc3i2rgno01byt0; expires=Mon, 02 May 2022 13:21:14 GMT; HttpOnly; Max-Age=15552000; Path=/; SameSite=Lax
```

### DIFF
--- a/transport_nantes/topicblog/static/topicblog/same_slug_warning.js
+++ b/transport_nantes/topicblog/static/topicblog/same_slug_warning.js
@@ -1,0 +1,53 @@
+// Calls server to get a list of all slugs and
+// how many of each slug there are.
+async function get_slug_dict() {
+    data = await $.ajax({
+        url: '/tb/ajax/get-slug-dict/',
+    })
+    return data;
+}
+
+// Get an url from the server that leads to a list
+// of all items with the given slug
+async function get_url_list(slug) {
+    data = await $.ajax({
+        url: '/tb/ajax/get-url-list/',
+        data: {
+            slug: slug,
+        },
+    })
+    return data;
+}
+
+$(document).ready(async function () {
+
+    var slug_dict = await get_slug_dict();
+
+    // Create a div in wich we will put the warning
+    $("input[id='id_slug']").after('<div class="small" id="slug_warning"></div>');
+    // On each change, we check the value of the slug.
+    // If it matches with an item of the list of existing slugs, 
+    // we display a message with a link to the list of items with the same slug.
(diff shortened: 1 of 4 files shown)


JeffAbrahamson:
The object is a `TopicBlobItem` or `TBItem` for short.  A slug is just a human-visible name for a TBItem.

PR #274, opened 2021-11-03
Add function to update unamed Users on donation

### LINKED ISSUE #178: Record name on donation
When creating accepting a donation, whether the user is new or old, if the first and last name are blank, we should fill them in with the name provided at donation.

If _either_ has been provided by the user, we should respect that and not modify them.  (And we should comment in the code that this is intentional.)  This avoids things like saying that my name is Joe but I make a donation as Joseph Biden and suddenly I'm greeted as "M Biden" or "Joseph" when I clearly said I'm "Joe".

___

Benja Edit to remember how we suggested this should happen : 

```
def name_if_unnamed(self, firstname, lastname):
    """Name user if unnamed.

    If the user has no first or last name, provide them.
    If either exists, however, we should respect them.

    This is for instances where the user provides a name without
    meaning to edit her/his user profile.  For example, if the user
    intends to be called "Joe", we shouldn't take a legal name
    attached to a bank card and start saying "Hello, Joseph".
    """
    if self.user.firstname or self.user.lastname:
        return
    self.user.firstname = firstname
    self.user.lastname = lastname

```

+ Tests

Shriukan33:
Some users only ever gave their mail address, so if they happen
to give us their name, we can use it to update their name in database.

Of course, if they did give a name that is different, we leave it alone.

Closes #178

### DIFF
--- a/transport_nantes/stripe_app/views.py
+++ b/transport_nantes/stripe_app/views.py
@@ -302,6 +302,7 @@ def stripe_webhook(request):
         except Exception as error_message:
             logger.debug("="*80, "\n", "Error while creating \
             a new Donation. Details : ", error_message)
+        update_user_name(event)
 
     # Event for subscription payments (initial or recurring)
     # cf https://stripe.com/docs/billing/subscriptions/webhooks#tracking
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
LGTM

Some questions:
1.  Do you need to check that metadata contains what you think it does?
2. Can you be more specific in catching the exception?  You're taking it on faith that the only possible exception is that the given email address is unrepresented in our database, no?
3. Perhaps instead of calling the variable `event` we could be more specific.  For example, `donation_event`?

PR #276, opened 2021-11-04
Fix wording and calculus of tax reduction.

Shriukan33:


JeffAbrahamson:
Thanks.  Although isn't it 66% and not 67%?
Check on the Fisc site.

Shriukan33:
> Thanks. Although isn't it 66% and not 67%? Check on the Fisc site.

it is, just checked ! The worst is, I did the change in the JS file :laughing: going to add a commit to that

PR #277, opened 2021-11-06
Add right arrow on CTA labels in TB article template

JeffAbrahamson:
We meant to have CTAs include a right arrow in articles.  This
corrects that omission.

Closes #272.


Shriukan33:
LGTM

PR #278, opened 2021-11-08
Add Variant creation

### LINKED ISSUE #240: TBv2 variant creation
On editing a TBItem, we should _always_ (unless the item is new) offer to create a variant.  After a configurable period of time (let's put it in `settings_local` for now, we'll set to 12 hours in prod but devs might have other needs) the only option is a variant.

Variant means same slug but otherwise same content except what the user has changed.

Note that this is from publication time.  Before publication, changes are fine.  After publication, the slug should no longer be modifiable.

We should also offer an option to make a copy.  The only difference between copy and variant is that a copy permits a new slug whereas a variant does not.  But that means that copies are variants if the user doesn't change the slug.

Shriukan33:
By creating variants, we want to have some history of a slug,
and allow ourselves to test different variations of a slug, for example,
changing social media images and titles.

:rotating_light: /!\ Implementation requires changing the settings_local.py file. /!\ :rotating_light:

Closes #240


JeffAbrahamson:
Looks mostly good.

* Are there tests?  ;-)
* Perhaps `time_since_publication` would be clearer as `seconds_since_publication_date`?
* Should we not always maintain the option to create a variant? It is the option to modify that closes, no?  Which would argue for a variable name change on `variant_available`?
 
I don't think you're disabling the slug field so much as marking it read-only.  And that fact is pretty obvious from the very readable code that comes after, so perhaps better to explain why: once a TBItem has been published for longer than the edit window, we no longer permit slug changes.  Perhaps we should have a  name for that (in the code), like calling the item frozen so that we can just say "beyond the publication edit window, the slug is frozen and so we don't permit modifying the slug".  And then that argues that there should be a model method called `is_frozen()` where this comment could live.

Indeed, as we explore how to explain this, it starts to seem clear that we need vocabulary.  If a TBItem can freeze, then it makes sense to call the period between publication and freezing the edit window (as you have).  And then in a comment in the model, maybe on the above suggested `is_frozen()` method and/or on the `publication_date` variable, perhaps we should just explain that the publication date sets the beginning of an edit window that allows for correcting mistakes and at the end of which we freeze the TBItem, and that freezing means the following...

Wdyt?

JeffAbrahamson:
BTW, when sending a PR for something that will require a change in deployment (here, adding a variable to `settings_local.py` on beta and prod), it's good form to remind ourselves of that in the PR text.  One less way to forget and hose beta (or prod).

Shriukan33:
> Looks mostly good.
> 
>     * Are there tests?  ;-)

No test in this, would require to test the interface to do so. :/ How would I Pythonize the action of clicking the Create a variant so I can test it ? 

> 
>     * Perhaps `time_since_publication` would be clearer as `seconds_since_publication_date`?

`time_since_publication` doesn't specify the time unit, because the value of this is a `timedelta`, so you can convert it into hours, seconds, days or whatever. Specifying seconds, I'd expect to find an Int in it... 
I added the type hint `timedelta` but wouldn't know how to make it clearer. wdyt ? 

>     * Should we not always maintain the option to create a variant? It is the option to modify that closes, no?  Which would argue for a variable name change on `variant_available`?
> 

In this current version, the only case variant_available == False is when you don't have any tbitem loaded in the form. basically, on /new, it would be weird and confusing to have the option to create a variant of an item that doesn't exist. 


> I don't think you're disabling the slug field so much as marking it read-only. And that fact is pretty obvious from the very readable code that comes after, so perhaps better to explain why: once a TBItem has been published for longer than the edit window, we no longer permit slug changes. Perhaps we should have a name for that (in the code), like calling the item frozen so that we can just say "beyond the publication edit window, the slug is frozen and so we don't permit modifying the slug". And then that argues that there should be a model method called `is_frozen()` where this comment could live.
> 
> Indeed, as we explore how to explain this, it starts to seem clear that we need vocabulary. If a TBItem can freeze, then it makes sense to call the period between publication and freezing the edit window (as you have). And then in a comment in the model, maybe on the above suggested `is_frozen()` method and/or on the `publication_date` variable, perhaps we should just explain that the publication date sets the beginning of an edit window that allows for correcting mistakes and at the end of which we freeze the TBItem, and that freezing means the following...
> 
> Wdyt?

Ok with the frozen term.
So when an item is published, its publication date starts the edit window. 
Once passed this window, it's "frozen" and can't be edited.

JeffAbrahamson:
Tests: this is one of the (many) reasons to refactor these functions into methods on the model -- you can test that.

`time_since_publication` - yes, that's a reasonable argument.  I can't help wondering if it's not more readable immediately to convert to seconds, though, since usage is local.  You don't  need the flexibility of a `timedelta` as part of an interface.

I agree that variants don't make sense if the item is new.  To me, this argues again for model methods like `is_new()` (just checks existence of pk id), `is_variant_ok()` and `is_editable()`.  (I'd like parallel construction there but I can't think of a single word that captures both.)  Those model methods are available in the template, then, like any model attribute.  If I'm not horribly misremembering how that works.

Shriukan33:
> Tests: this is one of the (many) reasons to refactor these functions into methods on the model -- you can test that.

In this case, clicking this button isn't even a function, it just adds the name of the button to the request.post. No JS in this. 
Making it a model function would probably require to add an Ajax layer on this, listening to clicks on this button, call server on some ajax url, create the variant the same way we do but adding something to the post so Django knows we want to create a variant somehow... 
I can do that, unless you see a better solution ? The lack of interaction between front and back end in Django is sometimes painful, although I understand it makes it independent on whatever front end framework you might use

> `time_since_publication` - yes, that's a reasonable argument. I can't help wondering if it's not more readable immediately to convert to seconds, though, since usage is local. You don't need the flexibility of a `timedelta` as part of an interface.

I think timedelta are readable, with the args being seconds, hours, days and so on, it's difficult to get it wrong. 
Now, converting it to seconds is probably really cheap if you think this will improve code readability, let me know what's your take !

> I agree that variants don't make sense if the item is new. To me, this argues again for model methods like `is_new()` (just checks existence of pk id), `is_variant_ok()` and `is_editable()`. (I'd like parallel construction there but I can't think of a single word that captures both.) Those model methods are available in the template, then, like any model attribute. If I'm not horribly misremembering how that works.

I'll add those functions, I think they are relevant. Although `can_create_variant()` sounds more precise to me, if it sits with you ?

JeffAbrahamson:
tests - ok, let's leave it for later, something still sits wrong with me about it, but clearly we're not converging on this, so let's not let it hold us up

timedelta - I agree with you for an interface, I disagree for encapsulated usage within a single function.  Modulo indentation, something like this:

```
            seconds_since_publication = \
                (datetime.now(timezone.utc) - \
                 self.instance.publication_date).seconds
            if seconds_since_publication > TOPIC_BLOG_EDIT_WINDOW_SECONDS:
                self.is_editable = False
```

`can_create_variant()` is better, yes!

Shriukan33:
> tests - ok, let's leave it for later, something still sits wrong with me about it, but clearly we're not converging on this, so let's not let it hold us up
> 
> timedelta - I agree with you for an interface, I disagree for encapsulated usage within a single function. Modulo indentation, something like this:
> 
> ```
>             seconds_since_publication = \
>                 (datetime.now(timezone.utc) - \
>                  self.instance.publication_date).seconds
>             if seconds_since_publication > TOPIC_BLOG_EDIT_WINDOW_SECONDS:
>                 self.is_editable = False
> ```
> 
> `can_create_variant()` is better, yes!

Timedelta converted to seconds, and functions created ! I dropped the is_new() for now, didn't find an use case, because what I really wanted to do was knowing if can_create_variants() so I ended up not using is_new().

JeffAbrahamson:
This is looking good.

One quibble: why aren't these model methods instead of form methods?  Surely they would be applicable even if you use a different form to interact about the model, no?

JeffAbrahamson:
"It worked" is a common justification for things.  Not always a good one.  LOL.

JeffAbrahamson:
LGTM

I'm merging this, but I don't have time this afternoon to update saltstack.  Send a PR on infra if you want to.  Note that the _value_ should go in pillar (and should easily be readable, like `12 * 3600`).

Shriukan33:
> "It worked" is a common justification for things. Not always a good one. LOL.

Haha indeed, there was no good reason to just keep it in the form model, at least now it's accessible anywhere in the code. 

> LGTM
> 
> I'm merging this, but I don't have time this afternoon to update saltstack. Send a PR on infra if you want to. Note that the _value_ should go in pillar (and should easily be readable, like `12 * 3600`).

PR made 

> "It worked" is a common justification for things. Not always a good one. LOL.

JeffAbrahamson:
Merged and deployed.

PR #279, opened 2021-11-09
Capture expired checkout session into database.

### LINKED ISSUE #247: Capture failed and abandoned payment events
and record them in the donations table as such

Shriukan33:
People that have got to checkout session but did not complete it
within 24 hours will be saved in database for later contact.

Closes #247

### DIFF
--- a/transport_nantes/stripe_app/models.py
+++ b/transport_nantes/stripe_app/models.py
@@ -78,6 +78,7 @@ class Donation(models.Model):
     # Any other value means repeat at N month intervals.
     # We should limit this to 12 max.
     periodicity_months = models.IntegerField(verbose_name="Période")
+    # A 0 cts amount is considered to be a cancellation.
     amount_centimes_euros = models.IntegerField(
         verbose_name="Montant (Centimes €)", default=0)
     timestamp = models.DateTimeField(auto_now_add=True, verbose_name="Date")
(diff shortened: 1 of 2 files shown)


transport_nantes/stripe_app/models.py
```diff
@@ -87,13 +88,21 @@ class Donation(models.Model):
...
     def __str__(self):
-        if self.periodicity_months == 0:
-            return self.first_name + " " + self.last_name + \
-                " ({}) ".format(self.email) +\
-                str(round(float(self.amount_centimes_euros/100), 2)) \
-                + "€ ONE-TIME" + str(self.timestamp)[:19]
+
+        display_name = ""
+        amount_given = str(
+            round(float(self.amount_centimes_euros/100), 2)) + "€"
```
JeffAbrahamson, line 94: Please force yourself to think separation of view (function) and display.  The euro sign is not part of the amount.  The rounding should be in the template (django floatformat filter).

transport_nantes/stripe_app/views.py
```diff
@@ -298,7 +298,8 @@ def stripe_webhook(request):
         logger.info("Stripe payment webhook succeeded.")
         logger.debug("Details attached to event : \n\n", "="*30, "\n", event)
         try:
-            make_donation_from_webhook(event)
+            donation = make_donation_from_webhook(event)
+            donation.save()
         except Exception as error_message:
             logger.debug("="*80, "\n", "Error while creating \
```
JeffAbrahamson, line 304: Why are you sure this exception doesn't deserve to be visible?

transport_nantes/stripe_app/views.py
```diff
@@ -317,6 +318,18 @@ def stripe_webhook(request):
             amount = int(event["data"]["object"]['amount_due'])
             save_recurring_payment_details(customer_id, amount)
 
+    if event["type"] == 'checkout.session.expired':
+        logger.info("Stripe checkout session expired")
+        logger.debug("Details attached to event : \n\n", "="*30, "\n", event)
+        try:
+            donation = make_donation_from_webhook(event)
+            donation.amount_centimes_euros = 0
```
JeffAbrahamson, line 326: Does this zero deserve a code comment?

JeffAbrahamson (approved):
I'm approving, but I'd appreciate if you could clean up the points I mention.  Thanks!

transport_nantes/stripe_app/models.py
Shriukan33: In what template ? This is purely for admin panel.

transport_nantes/stripe_app/views.py
Shriukan33: What do you mean by Visible ? Moving it up to Info ?

transport_nantes/stripe_app/views.py
Shriukan33: Yes it does, will add that

transport_nantes/stripe_app/models.py
JeffAbrahamson: ah, ok, I have less intuition about what the right way to interact with the admin panel is.

i missed that this was `__string__()`.

transport_nantes/stripe_app/views.py
JeffAbrahamson: Right, if that exception occurs, do you want anyone to be able to know?  Or is this only for debugging?

I'd say that if you get an exception processing a web hook, you'd want that to show in the log so that when the actual symptom arises (customer complaint, etc.) you can see something about what has happened.

You should read logs regularly to understand what your software is doing.  If yoiu can't tell from the log, it means you're not logging well.

Shriukan33: You're completely right, I've been just stuck in the mindset that debug = most complete log, so I definitely wanted to log the error if it occurred. 
That was without considering that we actually only check the info log, by default.

PR #280, opened 2021-11-09
Move text from form_flow.js to Django

### LINKED ISSUE #210: Avoid putting logic in javascript unless there's no other way around it
At lines 250 ff of `transport_nantes/stripe_app/static/stripe_app/form_flow.js`, you set the originating view and parameters of the form.  But those values are available to the view in django and so can and should be set there.

This makes the code more readable and easier to debug.  Pushing logic client-side makes debugging much harder when something goes wrong.

Shriukan33:
By doing so we aim to have an easier time locating and changing the
text that appears in the form.

Closes  #210

### DIFF
--- a/transport_nantes/stripe_app/forms.py
+++ b/transport_nantes/stripe_app/forms.py
@@ -126,7 +126,16 @@ class AmountForm(forms.Form):
                   template="stripe_app/inline_radio_custom.html",
                   id="payment_amount_rb", style="display: none"),
             Field("free_amount", id="free_amount", style="display: none"),
-            HTML("<p id='real_cost'></p>"),
+            HTML("""
+            <div id="real_cost_message" style="display: none">
+                <p>
+                Votre don ne vous revient qu'à
+                <span id='real_cost'></span>€
+                après réduction d'impôts !
+                </p>
+            </div>
+            """
+                 ),
             ButtonHolder(
                 Submit('submit', 'Continuer',
                        css_class='btn navigation-button',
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
Why is this in the .py and not the .html?

Shriukan33:
> Why is this in the .py and not the .html?

The form is generated by Crispy Forms, the HTML looks like this : 
```py
<div class="container-fluid d-flex-stretch col-xl-5 col-lg-7 col-md-10 col-sm-12 pt-1">
	<form id="amount_form" action="" method="post">
		{% crispy amount_form %}
	</form>
</div>
```
See [here](https://github.com/transport-nantes/tn_web/blob/master/transport_nantes/stripe_app/templates/stripe_app/donation_form.html)

JeffAbrahamson:
Have a look at `javascript-210-jeff-partial`.

JeffAbrahamson:
Little things can have big consequences.  You've been focused on making the text appear before the button, which forces you into a hard interaction with crispy.  (Aside: please tell me you have read [the docs](https://django-crispy-forms.readthedocs.io/en/latest/index.html) often enough to have internalised them, that you know all the hooks that crispy provides.)

If you accept that the message could appear after button, the problem becomes substantially easier.  (In the example that I pushed, there's a minor spacing problem on the text, but you'll see and fix that easily.)

Shriukan33:
> Little things can have big consequences. You've been focused on making the text appear before the button, which forces you into a hard interaction with crispy. (Aside: please tell me you have read [the docs](https://django-crispy-forms.readthedocs.io/en/latest/index.html) often enough to have internalised them, that you know all the hooks that crispy provides.)
> 
> If you accept that the message could appear after button, the problem becomes substantially easier. (In the example that I pushed, there's a minor spacing problem on the text, but you'll see and fix that easily.)

Things aren't as easy at they seem, commented on your solution [here](https://github.com/transport-nantes/tn_web/commit/0406ad719963cddd1545cdcec7b208d251aec567) 
I don't feel it makes it particularly easy ? What's the difference in the end ? It being in crispy isn't a problem, I only have to edit the HTML in forms.py instead of the template

PR #281, opened 2021-11-10
Move Text from Javascript to HTML template.

### LINKED ISSUE #210: Avoid putting logic in javascript unless there's no other way around it

Shriukan33:
By doing so we aim to have an easier time locating and changing the
text that appears in the form.

Closes #210


JeffAbrahamson:
Thanks!

PR #283, opened 2021-11-14
Publish button 241 jeff v2

JeffAbrahamson:
I had a go at sorting this one, since it looked like this was going to spin around for a while. Have a look and tell me what you think.

It needs tests -- I ran it through its paces, but automated unit tests would be great. They're tedious here, since we should have a test per field plus tests for the one-of and dependent cases.


Shriukan33:
Did a rebase of this branch, so SuperUser is here from the start, and applies on the view next. 

Also fixed a typo where you had "daetime" and deleted unused import (import re) 

Flake8 scan could help you with those :slightly_smiling_face: 

The rebased branch : https://github.com/transport-nantes/tn_web/tree/publish-button-241-from-Jeff-rebased

JeffAbrahamson:
I hate this about python: if my test didn't hit that precise code path, I don't detect the typo on datetime.  :-(

JeffAbrahamson:
oops, I just merged the PR, but I think you meant to close it and open a new one on a different branch...

I'm going to muck with master to set things right, hold tight

Shriukan33:
> I hate this about python: if my test didn't hit that precise code path, I don't detect the typo on datetime. :-(

Yup, some things can be broken but for typos, a good linter does the job, flake8 does a great job for this purpose. Typo, unused variables / imports, they usually show.

PR #291, opened 2021-11-15
Extend slugifier to cta and suggestions 265

### LINKED ISSUE #265: Extend JS slugifier to CTA slugs
CTA slugs don't enforce slug on the fly, to the contrary of the item's slug.

Shriukan33:
This branch adds : 

* Autocomplete for the item's slug field, the field will have suggestions based on the input of the user, as long as it is more than 5 characters long.

Example : 

Peek 2021-11-17 16-31

Note : There is a [known bug on Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1474137), which make the suggestion list not show up before user presses backspace. So you do the same process, and then hit backspace to see suggestions show up. 
I tried the autocomplete on/off that some claimed to work few years ago, but it doesn't seem to be the case anymore. 

Extend the slugifier already used on item's slug field to other slug fields. 
The slugifier also now deletes trailing hyphens if any. 

Closes #265


JeffAbrahamson:
In 639d05 aren't you introducing a newline into a log message?  That makes the message ungreppable.

JeffAbrahamson:
In 958eb66 you have a white-space only line in tb_item_edit.html.

It looks like you are not using a crispy filter on the slug fields.  Why is this not fragile?  What if crispy changes, how do you track the change?

You pass to the html document a list of all slugs that exist.
* How will this scale when we have 100 slugs?  Or 1000 slugs?  Or 10000 slugs?
* Will we always want all slugs to be discoverable?  What if some slugs correspond to restricted content, for example?  This code would enforce a requirement that slug names not leak content on restricted pages.

JeffAbrahamson:
In 026496814 you're matching the string "slug" in field names.  This is fragile.  Consider a field called "optional_slug_description" that's used to remind us internally why we named the slug what we did, what marketing decisions, etc.

JeffAbrahamson:
Aside: do you look at the HTML you generate?  You're transmitting a lot of white space over the wire.  This isn't a problem per se, but it's inefficient and something to avoid as you go.

JeffAbrahamson:
Aside: do you look at the HTML you generate?  You're transmitting a lot of white space over the wire.  This isn't a problem per se, but it's inefficient and something to avoid as you go.

Shriukan33:
> In 639d05 aren't you introducing a newline into a log message? That makes the message ungreppable.
> 
> Perhaps you meant to add two strings?
> 
> In a similar vein (of greppability) perhaps it would be better to repeat the import statement than to split over two lines?
> 
> ```
> from asso_tn.utils import StaffRequiredMixin, StaffRequired
> from asso_tn.utils import SuperUserRequiredMixin
> ```

Fixed. I don't really think in terms of greppability ! 



> In [958eb66](https://github.com/transport-nantes/tn_web/commit/958eb661c7966bc0d60ee289a6e42d147d5ed123) you have a white-space only line in tb_item_edit.html.
> 
> It looks like you are not using a crispy filter on the slug fields. Why is this not fragile? What if crispy changes, how do you track the change?

The `|as_crispy` filter doesn't produce the sort of input field I want.
To have the suggestions, I want to use datalist html tag, and crispy doesn't attach one to input fields.
However, I did reuse all classes and divs crispy uses. If we notice changes (difference between this field and the others) then we'd have to update the template as well to make it fit. However I don't expect this to happen all too often, so I don't think this is a big issue. 

> You pass to the html document a list of all slugs that exist.
> 
>     * How will this scale when we have 100 slugs?  Or 1000 slugs?  Or 10000 slugs?
> 
>     * Will we always want all slugs to be discoverable?  What if some slugs correspond to restricted content, for example?  This code would enforce a requirement that slug names not leak content on restricted pages.

So reworked this part with your observations and what we discussed in the issue #265, and now it only triggers a search if the input field is more than 5 char long, and you get only up to 20 results suggested, ordered by pk (which roughly should correspond to creation dates).


> In [0264968](https://github.com/transport-nantes/tn_web/commit/0264968144a59d0bf7d754689f1cd296a7e5aec7) you're matching the string "slug" in field names. This is fragile. Consider a field called "optional_slug_description" that's used to remind us internally why we named the slug what we did, what marketing decisions, etc.

Also reworked this, now I make the model give its list of `models.Slugfield` so I can be sure that we expect slug values in those. 
I then attach an event listener to each of those fields and use the same process that we used for the first slug field. 


It's been quite some work to get it right. 

Going to edit PR with visuals.

JeffAbrahamson:
LGTM.

PR #292, opened 2021-11-16
Test nonservable or not published 245

### LINKED ISSUE #245: Test that we don't serve unpublished or non-servable TBItems
Recall:
* We make an item publishable so that it may be served.  For now, only admins should be able to do that (in the admin interface, until we create a button).  This can not be undone.
* We mark them non-servable at EOL.  This _can_ be undone.

Shriukan33:
Add a check that item is servable before serving.

This acts as a fail-safe to ensure that the item is not served
when it shouldn't be.

Deleted published field, as discussed in the #245 issue. 

Closes #245


JeffAbrahamson:
`servable` defaults to true, when set false it stays false.

On the log line, the intent is that the log line emitted is one physical line.  The code can split over two.  You had introduced a newline into the emitted string.

In views, you do your query so that you only select things that are servable.  It's great that you check against the `is_servable` function.  If that fails, you should 500, we've made a mistake on our side.

Shriukan33:
> `servable` defaults to true, when set false it stays false.
> 

image
???

> On the log line, the intent is that the log line emitted is one physical line. The code can split over two. You had introduced a newline into the emitted string.
> 

image
This is correct right ? The + concatenates, it doesn't introduce breaklines, does it ? 

> In views, you do your query so that you only select things that are servable. It's great that you check against the `is_servable` function. If that fails, you should 500, we've made a mistake on our side.

Replaced the 404 by a 500 ! :slightly_smiling_face:

JeffAbrahamson:
servable -- *sigh* -- this is why it's important to document models well.  Our memories move around...

JeffAbrahamson:
```
In [1]: print("dog cat "\
   ...: "cow horse")
dog cat cow horse

In [2]: print("""dog cat
   ...: cow horse""")
dog cat
cow horse

In [3]: : print("dog cat "
   ...: "cow horse")
dog cat cow horse

In [4]:
```
Part of the point of triple quotes is that they permit you to print newlines without escaping them.

PR #295, opened 2021-11-22
Fix TBv2 form referring to .published attribute

Shriukan33:
The validation process was still referring to the .published attribute.
This leads to a 500 error on submit.


JeffAbrahamson:
Make the prints be logs?
Otherwise, LGTM.

PR #308, opened 2021-12-02
Revert the suggestion feature in TBv2

### LINKED ISSUE #306: Fix slug not saving on item creation and edit.
Slug field is absent from request.POST and saving, instead of modifying in place the current object, creates another with an empty slug field.

Shriukan33:
It introduced a breaking bug that made the slug field not save.

the suggestion feature is a nice to have one, but I'd say it's not exactly a priority either, as we have a warning if we're using the same slug as an existing article. 

Wdyt ?  

Closes  #306


JeffAbrahamson:
LGTM

PR #323, opened 2021-12-08
New index page

Shriukan33:
Based on gaelle's work : 
[20211208-index.pdf](https://github.com/transport-nantes/tn_web/files/7676652/20211208-index.pdf)

:warning:  Branch based on deployed branch nuke-banner ! 

Not ready to merge, need to have pages Agir and Petite Mobilité to complete the navbar. 

Agir page has its own PR : https://github.com/transport-nantes/tn_web/pull/324

PR #324, opened 2021-12-08
Add "Agir" page.

Shriukan33:
This page is meant to describe what Mobilitains needs in terms of
volunteers.


JeffAbrahamson:
This should have been TBv2...

LGTM.

JeffAbrahamson:
The jobs page also needs to die and be replaced with a set of TBItems.  Let's not add extra things that need to be mitigated in the near future.

As noted in [PR 323](https://github.com/transport-nantes/tn_web/pull/323), asso_tn is meant for items that require serious tracability.  Our goal is to make a site that we can rapidly evolve, and these heavily designed pages are not getting us there.

Issue #314, opened 2021-12-07
TBv2 list should display nicer


JeffAbrahamson:
It should offer data useful to users, not to developers.  Although a developer mode is fine `{% if is_developer %}`.

Some thoughts:
* snippets of the text
* the CTAs
* content type and template
* the title
* publication date
* servable

It should also offer buttons to publish, view, modify (if within modification time -- do we still permit that?), and create a variant.

If you need inspiration, look at github's issue list.

PR #329, opened 2021-12-12
Lots of usability improvements

### LINKED ISSUE #256: Progressively remove TBv1 code
# Phase One

In this issue we want to start by removing this line from `topicblog/urls.py`:
```
    path('s/<slug:item_slug>/', TopicBlogItemView.as_view(),
         name='view_item_by_slug'),
```

The ramification of that is that we will invalidate some tests that point to `view_item_by_slug`.  (If there had been any TBv1 tests, my commit b08e69d8 would have invalidated them, but we dodged that bullet...)

As always for such things, we want to remain as local as possible.  On the other hand, `view_item_by_slug` is clearly the right name for this path.  So I think the path is the following:

1. Remove the above line.  Change the name from `view_topic` to `view_item_by_slug`.
2. Check that tests still pass.
3. Audit the remaining code base for instances of `view_topic` and change them to `view_item_by_slug`.  The base template comes to mind, but we have to check _everywhere_ to be sure we don't 500 somewhere.

# Phase Two

After a couple days, when we're sure that we're happy with the TBv2 deployment, we should remove the code that we no longer need, such as `TopicBlogLegacyView`.

### LINKED ISSUE #326: Improve TBv2 usability/ergonomics
This is a bit of an umbrella issue for items that don't otherwise have their own issues.

JeffAbrahamson:
This PR is pretty wide-ranging.  It represents a part of my weekend and a bunch of things that have been itching me for a while now.

**Improve usability and ergonomics** of TBv2 by making TBItem state easily visible to editors.  On merge, this closes #326.  And #315.  And #314.  

**Add a general admin entry point**, which I'm calling a dashboard in the code, and which we'd call a *tableau de bord* for our users, no doubt, and which I kinda don't want to call the *Mobilibord*, even though some variety of pun is screaming to be made.  Discretion is the better part of valeur, surely.

**Remove most of the remainder of TBv1**, which can close #256.

**Stage permissions for TBv2 admins** so that we don't have to go giving people admin privs all the time.  We need to stage the permissions into the production database before we can assign anyone to them.  Once the necessary people are assigned, we can add the appropriate code to require they be set (instead of admin).

**Point users to general login** and not admin login.  I may have missed a few spots, we'll change others as we find them, if any there are.

**Misc improvements.**  Remove some dead stuff.  Move the dev/beta warning to top of page so it doesn't mess with responsiveness testing.  Reactivate the dev warning, which had accidentally been deactivated in git (a good indication that it was more annoying than useful).


Shriukan33:
LGTM

PR #333, opened 2021-12-13
Jeff weekend utm

JeffAbrahamson:
Basic traffic accumulator so we can understand where our clicks come from.

Probably I should add some data from ua_parser in a second commit.


Shriukan33:
Needs a rebase, otherwise LGTM

What's your plan with those data ?

Shriukan33:
LGTM

Issue #322, opened 2021-12-07
TBI save + edit loses images


JeffAbrahamson:
Make a new TBI, click save, then edit the form that's displayed, then click save.

If you actually come back to it in a new edit (by following a link), you're ok.

JeffAbrahamson:
If I hadn't seen the slack conversation, I don't think I'd have any idea what you were talking about here. ;-)

The issue title should summarise what you intend to do.

The issue body (in the case of a bug) should provide the repro such that someone other than you can reproduce it and so that it's clear whether or not the set of commits on the issue satisfy the problem / fix the bug.

Issue #331, opened 2021-12-13
timing bug with publication model


JeffAbrahamson:
We had the clever idea to create variants vs simply resaving TBItems.

And we had the quite reasonable goal of maintaining controls on publication, so that someone authorised to modify an article might need a second LGTM to publish it.

An unintended side-effect of this is that if X modifies a TBItem that is the only servable instance of that slug, then the servable instance is immediately marked unservable whereas the new items is not yet servable (because not yet published).  So that's a bug.

Maintaining a concept of "articles to retire when this one is published" introduces quite a lot of complexity to get right for very little benefit.  So I'm quite hesitant to go that route.

I think the simplest thing is simply to let the new item be unpublished.  You see the article immediately now (on PR 329) so it's easy enough to act on.  And after publication, you can retire the old one or not as you see fit.  The link to do so is shown in the control header.

_Aside: There are still issues with retirement.  The primary blocking issue is how to handle retiring the sole servable instance of a slug.  This isn't urgent today._

Shriukan33:
Trying to make it clear : 
So for now you say that when a servable item (that has been published) is being issued a variant (copy), it sets the item's `servable=False` and make the page unpublished. 

This happens here : https://github.com/transport-nantes/tn_web/blob/4b9be357475cceacd906425c667c61b475abe98c/transport_nantes/topicblog/views.py#L104

When you click "Sauvegarder", the form_valid() function will make unservable the item you sent through your post request. 

Idea would be to remove that line, correct ? Doing so, you would save in place the changes, before creating a variant with another pk. Note that variant would also be servable unless we decide otherwise, I just want to make clear your intentions. :slightly_smiling_face:

Shriukan33:
@JeffAbrahamson  As we progressed with TBv2 with the list view, what's your input on this issue ? I struggle to see the exact description on the problem here beside it's still not yet possible to retire a content? I hardly see what you mean with timing bug now

JeffAbrahamson:
Well, I think we've slowly evolved into something where "variant" doesn't even make a lot of sense any more.  We'd talked about save / save as draft / preview / cancel.  But it's becoming clear that the first three are all the same: save something and redirect to viewing it.  None of these things unpublishes anything.  Indeed, I think we should move to a model where no one is allowed to self-publish: you always need an LGTM provided by someone else clicking the publish button.

We might store the user id of the creator and of the publisher at that point.

[...]

PR #335, opened 2021-12-13
Fix images not saving between copies of a TBItem

Shriukan33:
Images are lost if we save a copy for modification in our current build.
This PR intends to fix this behavior by copying the originally edited item's images and transferring them to the next copy. 

Part of #322

### DIFF
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -528,6 +528,18 @@ class TopicBlogItem(models.Model):
 
         return slug_fields
 
+    def get_image_fields(self) -> list:
+        """
+        Return the names of fields that are Django ImageFields
+        """
+        all_fields = self._meta.get_fields()
+        image_fields = list()
+        for field in all_fields:
+            if isinstance(field, models.ImageField):
+                image_fields.append(field.name)
+
+        return image_fields
+
     def get_servable_status(self):
         """Return True if page is user visible, False otherwise."""
         if not self.servable:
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
I'm afraid I'm not too fond of this PR...

* It needs a test that fails without it and that succeeds with it.
* It's a bit of a hack.  I'd have preferred you read the django code to understand _why_ this copy didn't work and how we mis-used django.

This seems slightly wrapped up in #331, I do wonder if sorting that won't let us solve this more efficiently.

I'll propose something in return.

JeffAbrahamson:
I'm increasingly thinking we have to do something akin to what you propose.  But I think it's important that we understand why and document that, because otherwise it remains mysterious and could incite others to try to change/improve it and maybe pass through the same bit of discovery that we have.

The definition of Field in `django/db/models/fields/__init__.py` has a function `clone` with this comment:
```
Uses deconstruct() to clone a new copy of this Field.
Will not preserve any class attachments/attribute names.
```
The definition of `Field.deconstruct` is somewhat involved, but basically wraps up the field for cloning.

Image and File fields have `deconstruct` methods but not clone.  So the clone that gets called is in the base class and so it doesn't see the extra data for the file/image fields.  I think.  I get a bit confused sometimes with python resolution order.

So that might explain why we aren't pulling along the `ImageField` data.  This would therefore present the same problem for `FileField`.

The part I still don't fully understand is why the form doesn't include the image data, because it's being rendered as a "this is what the image was" field but not as a value on the image.

JeffAbrahamson:
And that "Currently" string _only_ appears in `contrib/admin/widgets.py` (and the .po files).

JeffAbrahamson:
I wonder if this comes down to our having used `FormView` instead of `ModelView` (or `UpdateView`).  Clearly `FormView` isn't expecting to preserve a model instance in the background.

JeffAbrahamson:
Merged with just a minor change in the comment.

Well done.

This issue annoys me.  There must be a better way, but I don't see it.

Issue #337, opened 2021-12-14
Improve TBv2 list view


JeffAbrahamson:
* briefer global list
* id's and better sort in local list

JeffAbrahamson:
@Shriukan33 I wanted to add a distinct on the query so that the list view would show only distinct slugs and then drop the buttons, only linking to the single slug list view.  The display is getting quite messy.

And that's when I was reminded of that "distinct on" bug in django, where it only supports this on postgresql.

Thoughts on a work-around?

The ORM is really convenient, but it's quite annoying how much is easy in SQL that is hard in the ORM.

Shriukan33:
> @Shriukan33 I wanted to add a distinct on the query so that the list view would show only distinct slugs and then drop the buttons, only linking to the single slug list view. The display is getting quite messy.
> 
> And that's when I was reminded of that "distinct on" bug in django, where it only supports this on postgresql.
> 
> Thoughts on a work-around?
> 
> The ORM is really convenient, but it's quite annoying how much is easy in SQL that is hard in the ORM.

If sqlite doesn't support DISTINCT ON statement, I only see 2 ways : 
  - use PostgreSQL as dev db
  - Not use this statement it in dev, but use it in prod / beta with a ROLE check. Sounds risky though. 

Writing raw queries, if not supported, wouldn't work anyway. That's for using purely the ORM.

Else we could probably come up with a Python script that takes the whole query and do the legwork that the ORM can't do. Querysets are python objects after all.

PR #338, opened 2021-12-14
Add id field to slug-specific list view

JeffAbrahamson:
Also include modification date in sort order.  Date modified is also
date created for us.

Part of #337.


Shriukan33:
LGTM

PR #344, opened 2021-12-16
Update CF page

Shriukan33:
After 1st goal has been reached, Gaelle updated the content of the page to generate more hype and simplify the content. 

It's been approved by Gaelle. 

image


JeffAbrahamson:
Be aware that deleting a file that's been deployed to static will not delete it from being servable.

Shriukan33:
> Be aware that deleting a file that's been deployed to static will not delete it from being servable.

The JS file was about the small screen menu, ultimately is was an useless choice to give to users so we decided to delete it. 
However it's still functional with current layout if it's still served. 

Should I delete it on static.mobilitains.fr ?

JeffAbrahamson:
Not necessarily, I just wanted to make sure you are acutely aware that you are deleting from git, not from production.

For example, if there were a reason the file should no longer be served, you'd want to make the file empty, commit and deploy that, make sure that's being served and in caches, and only then remove from git.

Shriukan33:
Once the server will refresh the cache, if it's empty or absent the file will be deleted in Prod, right ? 

Deploying doesn't clear the cache, iirc. So what's the point of deploying a blank file if the time of death is when the server refreshes the cache ?

JeffAbrahamson:
How does a cache work?

Shriukan33:
> How does a cache work?

Server (NGINX) has a special storage place dedicated to static files, to me this storage is the cache. 
it has a copy of files that are normally inside the Django so it doesn't have to pass as much data per request (and static files being mostly images, it's comparatively a lot more data to transfer). 
Periodically, NGINX checks for changes in the static files so it can update the cached files : No changes ? Great, can continue to serve this image. Not here anymore ? Can delete it from cache and serve whatever is replacing it. 

So to me, NGINX would have to check the list of static files it's caching against what django has, and update accordingly. 
In our case, on the periodic check the Webserver would notice that the .js file is not among django's static files anymore, and don't need to keep it as a consequence. 
Deploying would have as an effect to update the Django's list of static files. The js would still be served as long as NGINX doesn't update its cache. 

Am I far off ?

JeffAbrahamson:
Well, you have some notion of what a cache is. ;-)
It's complicated by the word "caching" meaning different things at different moments.

Maybe [this](https://www.cloudflare.com/en-gb/learning/cdn/what-is-caching/) can help some.

Nginx is a reverse proxy.  It doesn't cache at all.  It's nonetheless faster for nginx to serve a request for a file, which it serves directly, than to serve a request that results in a `proxy_pass` instruction (cf. nginx docs, our nginx configs).  The proxy_pass here hands off to wsgi which hands off to gunicorn which hands off to django's router and whatever else django does.  And then hands back in the other direction.

The context switches (control changing from one process to another) only take a few micrcoseconds, maybe a few tens of µs.  But there's also processor cache flushes (because there's some caching at nearly every layer, and the processor has the TLB and the L1 and L2 caches), so changing processes will cause new cache misses on the processor.  Add more tens of microseconds, even some milliseconds.

And then, of course, the messages themselves have to be passed.  It's not enough to wake up wsgi, you have to tell it why you woke it.  And so forth down the line and then back again.

All of which is a bit of a distraction on caching.

Where do copies of the .js file end up?  When someone on the internet requests the file, where does the request go and who serves it and why?

Shriukan33:
> Well, you have some notion of what a cache is. ;-) It's complicated by the word "caching" meaning different things at different moments.
> 
> Maybe [this](https://www.cloudflare.com/en-gb/learning/cdn/what-is-caching/) can help some.
> 
> Nginx is a reverse proxy. It doesn't cache at all. It's nonetheless faster for nginx to serve a request for a file, which it serves directly, than to serve a request that results in a `proxy_pass` instruction (cf. nginx docs, our nginx configs). The proxy_pass here hands off to wsgi which hands off to gunicorn which hands off to django's router and whatever else django does. And then hands back in the other direction.
> 
> The context switches (control changing from one process to another) only take a few micrcoseconds, maybe a few tens of µs. But there's also processor cache flushes (because there's some caching at nearly every layer, and the processor has the TLB and the L1 and L2 caches), so changing processes will cause new cache misses on the processor. Add more tens of microseconds, even some milliseconds.
> 
> And then, of course, the messages themselves have to be passed. It's not enough to wake up wsgi, you have to tell it why you woke it. And so forth down the line and then back again.
> 
> All of which is a bit of a distraction on caching.
> 
> Where do copies of the .js file end up? When someone on the internet requests the file, where does the request go and who serves it and why? And how long do you expect that to take at the different levels that it can happen?

Thanks for explanations. 

For us, the server storing our Data in cache would be cloudflare, it's our CDN.
Every request would have to go through Cloudflare first. So I suppose that's where the static file is stored.

Shriukan33:
On the topic of the branch, is this ok to merge ? Or there is still something not sitting right ?

JeffAbrahamson:
LGTM.
I just merged for you.  I'll let you deploy to make sure you're ready.  (I've not been paying attention to slack this afternoon, in the interests of getting Jellybooks work done.)

JeffAbrahamson:
On caching, you're right that cloudflare is our front end cache, but you're forgetting that there are many layers.  So a request for a static file (static.mobilitains.fr/something) hits cloudflare.  If cf has it in cache and it's not stale, it serves it.  Otherwise, it makes a new https request to nantes-6p.  Then the nginx config on nantes-6p for static.mobilitains.fr says "hey, my instructions are to serve files", so it looks up that path, and if it exists, streams the contents back on the connection, otherwise it serves a 404.  No django involved.

Issue #336, opened 2021-12-14
bullets are styled to none everywhere


JeffAbrahamson:
We style bullets to none everywhere.  I'm sure we need this somewhere, such as for the navbar, but it surely isn't right to make it the base case.

Example: the bullet list [here](https://www.mobilitains.fr/tb/t/benevolat-membre-ca/) generated by markdown on a TBv2 page.

JeffAbrahamson:
This is getting more urgent if Gaelle is working around it instead of accepting that bullets will arrive.

PR #345, opened 2021-12-16
Revert change on list style

Shriukan33:
Revert back to enabled the list styles.
This change was too rigid and difficult to use.

Part of #336

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/communique_presse_1.html
+++ b/transport_nantes/topicblog/templates/topicblog/communique_presse_1.html
@@ -49,16 +49,6 @@
           </style>
 
      {% endif %}
-     <style>
-          ol, li {
-          list-style:none;
-          margin-left: 0;
-          padding-left: 0;
-          }
-          li {
-          padding-bottom: 0.2rem;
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
I'm not sure that commit comment is understandable without today's context.

The commit content looks fine, of course.

Shriukan33:
Reworded the commit, don't know how to put it any better, suggestions welcome

JeffAbrahamson:
Yeah, fine.

Better, if you want, would be to think,
* why did we do this?
* what do we lose by undoing it?
* what do we gain by undoing it?

That's the sort of context one would want when looking at it six months later.  It's the context we don't really have looking at the commit message from six months ago.

JeffAbrahamson:
Yeah, fine.

Better, if you want, would be to think,
* why did we do this?
* what do we lose by undoing it?
* what do we gain by undoing it?

That's the sort of context one would want when looking at it six months later.  It's the context we don't really have looking at the commit message from six months ago.

JeffAbrahamson:
LGTM

PR #346, opened 2021-12-17
Og twitter

mcmikashi:
Hello I do 2 commit for solve this issue:

1. I add unit test for testing the context of topicblog
2. I implement the solution on the model topicblog


JeffAbrahamson:
Feedback:
* It's a bit weird that the test commit occurs before the code commit.  Indeed, at c57fdc1 tests fail (naturally enough).  Could you please reorder the commits (git rebase -i)?
* I think these files should be in asso_tn/ rather than in topicblog/, as they aren't (shouldn't be) specific to topicblog but to any page served by tn_web.  Could you please move the files (git mv)?
* You implement this by changing model data.  That doesn't seem right.  Why would a user expect his data to change?  In addition, it doesn't make clear what is happening.  I'd expect this to be mostly a change in the base template.

PR #348, opened 2021-12-18
Og twitter template

mcmikashi:
Hello I do 2 commit for solve this issue:
1. I implement the solution on the base template in asso_tn
2.  I add unit test for testing the meta in the template

### DIFF
--- a/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
+++ b/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
@@ -90,11 +90,20 @@
 				{% endif %}
 
 			{% endblock styles %}
-
-			<meta property="og:title" content="{{ social.twitter_title }}" />
+			{# OG title #}
+			{% if social.og_title %}
+				<meta property="og:title" content="{{ social.og_title }}" />
+			{% else %}
+				<meta property="og:title" content="{{ social.twitter_title }}" />
+			{% endif %}
 			<meta property="og:type" content="website" />
 			<meta property="og:url" content="{{ request.build_absolute_uri }}" />
-			<meta property="og:description" content="{{ social.twitter_description }}" />
+			{# OG description #}
+			{% if social.og_description %}
+				<meta property="og:description" content="{{ social.og_description }}" />
+			{% else %}
+				<meta property="og:description" content="{{ social.twitter_description }}" />
+			{% endif %}
 			{% comment "About is_static" %}
 			On views that use static images and not ImageFields, the .url property doesn't
 			exist (e.g. AssoView uses a string to designate the image to use).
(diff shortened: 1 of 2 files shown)


Shriukan33:
Good job on the test, these look coherent, however : 

The logic in the template has a flaw, for example : 

* [Here](https://github.com/transport-nantes/tn_web/pull/348/commits/51c23388d35dbcc05b22915dd8f7ae10880e9f01#diff-6fda6a8ea94ad1c733260a4909c974b0597fd3c22d05ff70b0210e7417d6cd18R159) You check if page.twitter_image != "asso_tn/mobilite-pour-tous.jpg"

Let's assume the case where page.twitter_image == "". 
It's indeed different from "asso_tn/mobilite-pour-tous.jpg"
But do we want to replace "asso_tn/mobilite-pour-tous.jpg" by "" ? I don't think so. 

Does your code handle the blank case like this one ? The same is worth for all "!=" comparisons. 

Besides, you say there are 2 commits when there are 3, one being a merge to master. You should rebase your branch on master and then `git push --force` your branch. 
Are you familiar with rebases ?

mcmikashi:
Thanks.

Yes, when I try every blank is replaced with the default, but I make the change.

I was waiting to see if everything was good and see I was behind I make the pull directly on github but now I understand that was not the best practice.

I am not very familiar with rebases I know that we can squash all commit into one and reorder commit but if you ask  me to do somethings i will find the way to do it.

JeffAbrahamson:
Hi, @mcmikashi .  Sorry for the long silence.  Holidays and stuff.  (Happy new year!)

Is this ready for us to review again?  I see there was a push but no comment on whether you were ready or had more to do.  Don't want to review if you're just interrupted for holidays.

mcmikashi:
Hi @JeffAbrahamson. I know It will take time because of the holidays. Thanks Happy new year!
Sorry for the late response I am actually sick.
Yes, this is ready for review again.

transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
```diff
@@ -103,11 +112,27 @@
 			if they are designated with a string.
 			{% endcomment %}
 			{% if is_static %}
-				<meta name="og:image" content="{% static social.og_image %}" />
+				{# OG image #}
+				{% if social.og_image and social.og_image != "velopolitain/v1.png" %}
+					<meta property="og:image" content="{{social.og_image.url}}" />
+				{% elif social.twitter_image and social.twitter_image != "asso_tn/mobilite-pour-tous.jpg"%}
```
JeffAbrahamson, line 118: Here and later, why are you comparing to this hard-coded fixed string?  The base template has no knowledge of what or why any particular image file would be supplied.

transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
```diff
@@ -103,11 +112,27 @@
...
+					<meta property="og:image" content="{{ social.twitter_image.url }}" />
+				{% elif page.header_image  %}
```
JeffAbrahamson, line 120: This is a nice touch, falling back to the header image.

transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
```diff
@@ -103,11 +112,27 @@
...
+					<meta property="og:image" content="{{page.header_image.url}}" />
+				{% else %}
```
JeffAbrahamson, line 122: You seem confident that social.og_image_url will be defined
That confidence merits a comment explaining itself.

JeffAbrahamson (changes requested):

transport_nantes/asso_tn/tests_og_twitter.py
```diff
@@ -0,0 +1,208 @@
...
+from datetime import datetime, timezone
+
+
+class TestOg_Twitter(TestCase):
+    def setUp(self):
+        # Creates a user for FKs
+        self.user = User.objects.create(username='user-og-twitter',
+                                        password='mdp-og-twitter')
+        # Creates a base content type for FKs
+        self.content_type = TopicBlogContentType.objects.create(
```
JeffAbrahamson, line 14: There are two types of pages.  Pages, like TopicBlog, source their images from MEDIA.  Pages like legal source their images from STATIC.  The base template has to respond to this distinction, checking for media or static.

You are only testing one case here.

JeffAbrahamson (changes requested):

transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
mcmikashi: When I try everything empty i still got some " default reponse " this is why I check if this is not the " default " the default are in This file  tn\tn_web\transport_nantes\asso_tn\middleware\default_context.py

transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
mcmikashi: I think the default will be always there because when i try empty the default context replace everythings (og image,twitter_image,twitter_title and twitter_description) that is  empty.

transport_nantes/asso_tn/tests_og_twitter.py
mcmikashi: Ok I look the legal source and the jobs page seems to be STATIC too but they have no og or twitter data on the template of job i can see  {% block og_title %} but there no  {% block og_title %} on the base template. I will try too look more deep for that and implement the test for this kind of page.

Shriukan33:
@mcmikashi Do you want to close this one ?

mcmikashi:
Yes I close this one.

Issue #339, opened 2021-12-14
integrate rollbar on free tier


JeffAbrahamson:
* [django](https://docs.rollbar.com/docs/django)
* [pricing](https://rollbar.com/pricing/#)

PR #349, opened 2021-12-20
Fix Logging in stripe app

Shriukan33:
Logger wasn't properly configured, leading to an absence of logs.
Also now handles the case where the event is an unknown type and returns
a 400 error.

Part of #233


JeffAbrahamson:
Cf. #339

JeffAbrahamson:
Generally, one increases logging level when one needs something extra for a while.  If one wants it all the time, it's level info.  If one wants it flagged, it's warning (not good, please look at me) or error (not good, definitely look at me).

The specific section stuff is useful in really big projects, but it's excessive complexity for us.  We're way too small for that to be worthwhile.

JeffAbrahamson:
Just before we 500, perhaps those should be errors?

Shriukan33:
Modified info to errors before 500 and also added the "django" logger in the TBv2 views as you pointed it out;

JeffAbrahamson:
LGTM.

PR #354, opened 2021-12-22
Fix logging messages to handle variables.

Shriukan33:
The logger didn't accept extra arguments as additional text like print do, fixed it so now we have full details about the events going through our webhook. 

Part of #233


JeffAbrahamson:
LGTM.

PR #355, opened 2021-12-22
Remove unpublish

### LINKED ISSUE #331: timing bug with publication model

Shriukan33:
Closes #331 

Reverts the unpublishing upon hitting save button. 
Also some code cleaning to abide by PEP8

### DIFF
--- a/transport_nantes/topicblog/views.py
+++ b/transport_nantes/topicblog/views.py
@@ -14,7 +14,6 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.urls import reverse, reverse_lazy
 
 from asso_tn.utils import StaffRequiredMixin, StaffRequired
-from asso_tn.utils import SuperUserRequiredMixin
 from .models import TopicBlogItem, TopicBlogTemplate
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
LGTM.

PR #356, opened 2021-12-22
Add script to initialize the database easily.

### LINKED ISSUE #310: Ease onboarding experience for new contributors
Currently, our readme doesn't detail steps to take to be able to contribute to this project. 

We should have a step by step guide to launch the website on a local environment, and contribution guidelines. 

Finally, create a tag for new contributors to find good first issues.

Shriukan33:
Objective is to make the onboarding as easy as possible for
possible new contributors.

Closes #310

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
-This repository contains the Transport Nantes website code.
-
-This program is distributed under the GNU GPL v3 licence.  See the file LICENCE for its text.
-
 # Project description
 
-(to come)
+This repository contains the Transport Nantes website code.
+We aim to promote a better, safer, more reliable mobility in France.
+
+This program is distributed under the GNU GPL v3 licence.  See the file LICENCE for its text.
 
 # Table of contents
 - [1. How to install](#how-to-install)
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
* Why is a db init command located under topicblob?
* By putting it here, it will be available on all hosts, including prod and beta.  What would happen if someone ran it in those instances?

Shriukan33:
* Thought I would place it as it mainly creates a topicblog item. 
* That would be remote code execution, which would be worrying. However better be safe than sorry, could add a check for role == "dev" ? 
* Any newcomer need to setup its db at some point, what do you suggest ?

JeffAbrahamson:
My thought is that we could ask the question "suppose the database is initialised as documented in the readme", so, basically, just a superuser and nothing else.  What can go wrong?

I think the main thing that can go wrong is that you try to visit the home page. ;-)  That by itself doesn't seem to warrant a new manage action.

Let's deprioritise this and just make a design issue where we list the things that can go wrong on an empty db, then think through what we want.  It's only opinion, but I'm inclined on this one to trust my hunch that we've started in a direction that could create nasty goblins later.

PR #370, opened 2022-01-02
Mailing list cleanup 342

JeffAbrahamson:
Clean up the mailing list / petition code a bit.

Note that this is based on a branch of yours, which looks ok to me but for which you didn't send a PR.  So perhaps there is something missing there.


Shriukan33 (changes requested):

JeffAbrahamson:
Good catch, fixed.
Also removed the help text, which crispy provides.

JeffAbrahamson:
Should be good now.

Shriukan33 (approved):

Shriukan33:
LGTM

PR #375, opened 2022-01-03
Update Readme for mac users

DennisOsei:
1. Updated Readme for M1 Mac users.
2. Removed force_text from  /transport_nantes/authentication/views.py because there was no function named force_text in django/utils/encoding.py.


JeffAbrahamson:
Hey, Dennis.  Thanks for the PR!

It looks good.  Can I request a really tiny change: remove the view mod.  You're absolutely right, but (1) it belongs in a separate commit and (2) there's currently a much more extensive PR pending from this weekend that refactors that file, so let me push that change to the other PR.

Some day we should probably replace the pathnames (which have strong platform-specific smells) with something more generic.  But that's for another day, just struck me in reading the doc.

Thanks again!

JeffAbrahamson:
Ideally, this would be squashed into a single commit...

JeffAbrahamson:
I've squashed and merged.

Thanks!

PR #378, opened 2022-01-03
Update Readme for M1 Mac users.

DennisOsei:
Don't know if I'm doing this right but I tried to revert the changes to remove the views.py from the pull request.


JeffAbrahamson:
Your first PR also updated, and I've merged that one.  So I'll close this one, because it's already merged.  Many thanks for the doc improvements!

The way to do this is to do the commit you did, but instead of making it a new commit object, you'd do a `commit --amend`.  After the fact, and so what I did, was to do a `git rebase -i master`, which lets you edit commit objects post facto.  Here I squashed the second into the first, which has the effect of creating a new history in which the view modification never happened.

Reading code change history is hard enough without leaving all the little explorations and errors that are normal to human thought and development. ;-)

Anyway, thanks again!

PR #385, opened 2022-01-09
Markdown 382

JeffAbrahamson:
This doesn't complete many issues, but it does seem to work and advances the state of affairs.


transport_nantes/topicblog/tn_links.py
```diff
@@ -60,12 +60,14 @@
...
-def render_inclusion_tag_to_html(tag_source, tag_name):
+def render_inclusion_tag_to_html(tag_source, tag_name, kwargs):
     """Render an inclusion tag to a string.
     """
 
     template_string = f"{{% load {tag_source} %}}{{% {tag_name} %}}"
-    template_context = {'csrf_token': settings.csrf_token}
+    template_context = {'csrf_token':  settings.csrf_token,
+                        'mailinglist': kwargs['mailinglist']}
+    print('tc', template_context)
```
Shriukan33, line 70: Dev print here

transport_nantes/topicblog/tn_links.py
```diff
@@ -38,7 +38,7 @@
     [[don:adhésion]]((text))   equivalent to {% bouton_join "text" %}
     [[don:fixed|1]]((text))    equivalent to {% fixed_amount_donation_button 1 "text" %}
 
-    [[news:name]]((text))      equivalent to {% show_mailing_list %}
+    [[news:name]]((text))      equivalent to {% show_mailing_list name %}
```
Shriukan33, line 41: The news MD tag is pretty confusing in the name, I'd expect news to be somewhat related to articles (The news ..) not to show a subscribe to mailing list block.

This isn't new for the name, it's just that it made me realize it wasn't intuitive for me.

transport_nantes/topicblog/test_tn_links.py
```diff
@@ -51,10 +51,14 @@ def test_buttons(self):
         self.assertEqual(self.parser.transform('[[don:fixed|5]]((give!))'), \
                          don.fixed_amount_donation_button(5, 'give!'))
 
-    def test_news(self):
-        settings.csrf_token = 'some text that must exist'
-        self.assertEqual(self.parser.transform('[[news:kangaroo]]((aardvark))'), \
-                         render_inclusion_tag_to_html('newsletter', 'show_mailing_list'))
+    ## Disabled due to captcha matching.
```
Shriukan33, line 54: Faced the captcha validation problem with the login form, turns out there is a way to get away with it in tests. 
 Can see here how it's done : 
 https://github.com/transport-nantes/tn_web/commit/e077ef0173a990cbfc76d25367972e682043913b#diff-e4c08575b99cb05e4808251bfb547bc1a429237cd6af2ac65d313cf1e59c02b0

TLDR; Captcha has a [test mode](https://django-simple-captcha.readthedocs.io/en/latest/advanced.html#captcha-test-mode)

transport_nantes/topicblog/tn_links.py
JeffAbrahamson: Yes, I agree.  I want to change the name, but that seems best as a separate change, properly staged (introduce new name, deploy, make sure old is out of service, then retire old).

transport_nantes/topicblog/test_tn_links.py
JeffAbrahamson: Yeah, I looked at that.  But it makes the captcha accept whatever it gets so that the submit validates, it doesn't make it produce the same string.  And here I was testing that the two functions generate the same html.

transport_nantes/topicblog/tn_links.py
Shriukan33: Agreed

transport_nantes/topicblog/test_tn_links.py
Shriukan33: Ha, got it. Indeed, seems like we shall have to do without it.

JeffAbrahamson:
Back to you.

Shriukan33:
LGTM

PR #386, opened 2022-01-11
Fix invalid token for non existing users.

Shriukan33:
Django would look to existing users to log in, which it couldn't find.
If the link is clicked and the token is valid, we create a user
without a password and a random username.
This way they can actually log in.

Part of #119


Shriukan33:
Left a dot in the git commit message let me reword it :)

Shriukan33:
> LGTM.
> 
> Suggestion/questions:
> 
>     * We should avoid the use of the word "account" for users, even internally.  It's not our external metaphor, and using it internal presents heightened risk of leakage.
> 
>     * At the end of the diff, why do you need to catch `User.DoesNotExist` ?
> 
>     * In that same try/except, where can the type, value, and overflow errors come from?

Higher risk of leakage in the sense someone looking for the keyword ? 

For the last try/except : 
Those errors were there before me, so i left them. 
You should ask yourself a Year ago in 3919e05ae6e2dc2e06895edd372bcc78ac863cd7 haha

JeffAbrahamson:
> > LGTM.
> > Suggestion/questions:
> > ```
> > * We should avoid the use of the word "account" for users, even internally.  It's not our external metaphor, and using it internal presents heightened risk of leakage.
> > 
> > * At the end of the diff, why do you need to catch `User.DoesNotExist` ?
> > 
> > * In that same try/except, where can the type, value, and overflow errors come from?
> > ```
> 
> Higher risk of leakage in the sense someone looking for the keyword ?

We (humans) tend to use in public the words we use in private.

> For the last try/except : Those errors were there before me, so i left them. You should ask yourself a Year ago in [3919e05](https://github.com/transport-nantes/tn_web/commit/3919e05ae6e2dc2e06895edd372bcc78ac863cd7) haha

Let's clean that up, then.  Can be a second commit, though, to maintain historical sanity.  Maybe just make an issue (and get to it later today, perhaps, or perhaps not).

I can't think of any reason that we'd throw in those last five lines.

Shriukan33:
Added a commit on the spot, wdyt ?

JeffAbrahamson:
LGTM

PR #389, opened 2022-01-11
Delete deprecated templates

### LINKED ISSUE #366: Clean up templates
We have templates lying around from transport_nantes days that we no longer user.  For each template file, search the code for instances of its use.  If there are none, remove the file.

Note that you should be able to do this with a couple invocations of `find`.

Shriukan33:
Dated from before the Transport Nantes -> Mobilitains branding
transition they are no longer used.

panel folder deleted -> Replaced by mpanel

hero_base_transport_nantes -> No other file refers to it, was used before the branding transition. 

Closes #366


JeffAbrahamson:
LGTM

PR #390, opened 2022-01-11
Captcha error 384

### LINKED ISSUE #384: Fix error reporting on bad captcha text
The `show_mailing_list` templatetag, on presenting a bad captcha, runs the invalid code path but doesn't say why.

Shriukan33:
When the user got something wrong (either missing email or invalid
captcha), no error was displayed, which could lead to confusion.

Minor improvement : The mail entered in the invalid form is pre-filled
in the new form.

Closes #384

Aside, some light PEP8 formatting, in a separate commit.

### DIFF
--- a/transport_nantes/mailing_list/forms.py
+++ b/transport_nantes/mailing_list/forms.py
@@ -1,9 +1,9 @@
 from django import forms
 from django.contrib.auth.models import User
-from django.forms import Form, ModelForm
+from django.forms import ModelForm
 from captcha.fields import CaptchaField
-from authentication.models import Profile
 from .models import MailingList
+
 
 class MailingListMMCF(forms.ModelMultipleChoiceField):
     """A custom form that gives us pretty mailing list labels in
@@ -18,6 +18,7 @@ class MailingListMMCF(forms.ModelMultipleChoiceField):
     def label_from_instance(self, member):
         return '{name}'.format(name=member.mailing_list_name)
 
+
 class MailingListSignupForm(ModelForm):
     captcha = CaptchaField(
         label="Je suis humain",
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
How do we know that the email address is incorrect?

What would trigger the error?  I think (from memory of testing) that a syntactically invalid email address will prevent form submission.  So what's an example other than bad captcha that leads to this message?

All this leads up to, if it's a bad captcha, let's say that.  If the problem is with the email address, we should be more specific, because we don't want to send the user searching for an alternate email, thinking they got that wrong.

Shriukan33:
Actually, due to some quirks in how the form is made (ModelForm and Meta class), the email field is not even mandatory. Couldn't find how to make it mandatory without rewriting the whole form. 

The only form of error is that form_invalid will be called server side.

JeffAbrahamson:
image

JeffAbrahamson:
I'm unable to submit the form with empty or syntactically invalid email address.

What happens for you?

Shriukan33:
empty-mail

I can enter an empty field. If I start writing it will ask for an email though.

JeffAbrahamson:
Hmm, curious that the two browsers behave so differently.  The joys of web programming!

So maybe we should just change the error message so that we check which it is?  Either bad captcha or no email address (or, eventually, malformatted address, say without "@" or without domain)?

JeffAbrahamson:
What is the status on this?  Last comment is 12 Jan concerning only refining an error message.  Is something blocking @Shriukan33 ?

Shriukan33:
Just didn't get back on this yet, so no, it's pending.

Shriukan33:
Now a separate errors appear for email and captcha.

JeffAbrahamson:
Very nice.

Is it possible to make the error appear closer to where it matter?  I'm imagining this being easy for some users to miss / not understand.

image

Usually web forms (the good ones) indicate errors just above or below the pertinent field.

Shriukan33:
The email error will show right below it, but I think the captchaField is kind of special (maybe because this is a lib), I couldn't get it to be formatted below in the same fashion as email field. I really tried but couldn't figure how to make it appear, so i decided to display all errors as a list so there is something. 
I don't think it's worth spending more time to get the red border around the captcha field.

JeffAbrahamson:
Fair enough.

LGTM.

PR #396, opened 2022-01-12
Make mailinglist template tag context sensitive

### LINKED ISSUE #391: Make inclusion-tag context-sensitive
The Mailing List markdown tag renders HTML code that isn't context-sensitive. 

The render_inclusion_tag_to_html() function, called whenever we use the `[[news:arg]]((text))` markdown templatetag, creates a new context space in the html it creates. 
https://github.com/transport-nantes/tn_web/blob/f838bb0c8da753431cbdec9cf915c4632aec3ad8/transport_nantes/topicblog/tn_links.py#L63

It means that the HTML that is produced by this tag can't access the context variables contained in the context dictionary returned by the view. it prevents this part of the code to access variables such as `user`, and its attributes `user.is_authenticated`

Example of this in a TBItem where in body_1_markdown I entered the  `[[news:arg]]((text))` markdown templatetag with index as a slug. 
This page is particular because : 
*  it has the `{% show_mailing_list mailinglist="arg" %}` tag in its html (see this line : https://github.com/transport-nantes/tn_web/blob/b7829054e5f7c48ca178a5b8be27e2e270b599d2/transport_nantes/topicblog/templates/topicblog/content.html#L123) 
* It also has the markdown templatetag. 

The result is this : 
image
image

To display the `this_context` var, I added it to the render_inclusion_tag_to_html() function as this : 
```py
def render_inclusion_tag_to_html(tag_source, tag_name, kwargs):
    """Render an inclusion tag to a string.
    """

    template_string = f"{{% load {tag_source} %}}{{% {tag_name} %}}"
    template_context = {'csrf_token':  settings.csrf_token,
                        'mailinglist': kwargs['mailinglist'],
                        'this_context': "render_inclusion_tag_to_html"}  # <-- HERE
    html = Template(template_string).render(Context(template_context))
    return html
```

And in the `transport_nantes/mailing_list/templates/mailing_list/panel/mailing_list.html`, template responsible for displaying both the markdown templatetag and the `{% show_mailing_list mailinglist="arg" %}` tag, I inserted this : 
```html
	{% if request.user.is_authenticated %}
		<h1>User is auth : {{ user.email }} </h1>
		<h1> This context var : {{ this_context}} </h1>
	{% else %}
		<h1> user not auth </h1>
		<h1> This context var : {{ this_context}} </h1>
	{% endif %}
```

This shows two things : 
* At rendering time, `{% show_mailing_list mailinglist="arg" %}` tag doesn't have access to `this_context` but `[[news:arg]]((text))` does. 
* Still at rendering time, `[[news:arg]]((text))` does not have access to `request.user` context variable, but `{% show_mailing_list mailinglist="arg" %}` tag does.

To solve this issue we'd need to make `[[news:arg]]((text))`  MD tag aware of the context of the page, at least the request. 

This issue is a pre-requisite of #383

Shriukan33:
The HTML code produced by this tag is now aware of the general context
provided by the context processor.
This means we can now check the user's login status for example.

Closes #391

### DIFF
--- a/transport_nantes/topicblog/templatetags/markdown.py
+++ b/transport_nantes/topicblog/templatetags/markdown.py
@@ -3,9 +3,9 @@ from django.utils.html import escape, mark_safe
 from markdown2 import markdown
 
 from topicblog.tn_links import TNLinkParser
-from django import template
 
 register = template.Library()
+
 
 @register.simple_tag(takes_context=True, name="tn_markdown")
 def tn_markdown(context, value):
@@ -18,5 +18,5 @@ def tn_markdown(context, value):
 
       {% tn_markdown md_text_variable %}
     """
-    parser = TNLinkParser(verbose=False)
+    parser = TNLinkParser(context, verbose=False)
     return mark_safe(markdown(parser.transform(escape(value))))
--- a/transport_nantes/topicblog/test_tn_links.py
+++ b/transport_nantes/topicblog/test_tn_links.py
@@ -9,7 +9,7 @@ from .tn_links import TNLinkParser, render_inclusion_tag_to_html
 
 class TnLinkParserTest(TestCase):
     def setUp(self):
-        self.parser = TNLinkParser(verbose=False)
+        self.parser = TNLinkParser({}, verbose=False)
 
     def test_broken_links(self):
         self.assertEqual(self.parser.transform('[['), '[[')
@@ -71,7 +71,7 @@ class TnLinkParserTest(TestCase):
         html = 'dog ' + don.external_url(url, label) + ' cat'
         self.assertEqual(self.parser.transform(markdown), html)
 
-        self.parser = TNLinkParser(verbose=False)
+        self.parser = TNLinkParser({}, verbose=False)
         pdll = 'Pays de la Loire'
         pdll_url = 'https://dog/cat/horse'
         markdown = '[[externe:{label}]](({url}))'.format(label=pdll, url=pdll_url)
@@ -89,7 +89,7 @@ class TnLinkParserTest(TestCase):
         html = 'dog ' + don.external_url_button(url, label) + ' cat'
         self.assertEqual(self.parser.transform(markdown), html)
 
-        self.parser = TNLinkParser(verbose=False)
+        self.parser = TNLinkParser({}, verbose=False)
         pdll = 'Pays de la Loire'
         pdll_url = 'https://dog/cat/horse'
         markdown = '[[EXTERNE:{label}]](({url}))'.format(label=pdll, url=pdll_url)
--- a/transport_nantes/topicblog/tn_links.py
+++ b/transport_nantes/topicblog/tn_links.py
@@ -1,3 +1,5 @@
+from enum import Enum, unique, auto
+
 from django.conf import settings
 from django.template import Template, Context
 from django.urls import reverse, NoReverseMatch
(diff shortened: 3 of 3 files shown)


JeffAbrahamson:
No test change?

Shriukan33:
Forgot to make test, updated with the new arg for TNParser !

JeffAbrahamson:
What is the status on this PR?

* You've added a test, but haven't signalled that it's ready for review.
* On review anyway ;-), I see the tests don't actually verify anything about the context, just provide an empty one.  That it least sorts the syntax error side, but doesn't test the new code.  The tests also fail at c906ecf, so shouldn't these commits have been squashed anyway?
* The code suggestion / simplification in the review lacks a response.  I just tested this by hand, it seems to work fine, which would argue for simplifying if no other issues impede it.

Shriukan33:
> 

I can't find commit c906ecf ? Anyway if you meant update_tn_links tests I can squash it with "Make mailinglist template tag context sensitive" commit. Then make another commit with new tests.

JeffAbrahamson:
> > 
> 
> I can't find commit c906ecf ? Anyway if you meant update_tn_links tests I can squash it with "Make mailinglist template tag context sensitive" commit. Then make another commit with new tests.

Did you try checking it out?  It's yours, it should still be available.

```
╭╴ (master $%=) [virt]╶╮
╰ jeff@birdsong:transport_nantes $ git checkout c906ecf
Note: switching to 'c906ecf'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at c906ecf Make mailinglist template tag context sensitive
╭╴ ((c906ecf...) $%) [virt]╶╮
╰ jeff@birdsong:transport_nantes $ git log -2
commit c906ecf29fc6907b7adeba862d16d369b00a88f8 (HEAD)
Author: Benjamin Mourgues <benjamin@mobilitains.fr>
Date:   Wed Jan 12 10:24:16 2022 +0100

    Make mailinglist template tag context sensitive
    
    The HTML code produced by this tag is now aware of the general context
    provided by the context processor.
    This means we can now check the user's login status for example.
    
    Closes #391

commit b21b0d1683521ea9a23a42e9c341a1823ac355be
Author: Jeff Abrahamson <jeff@p27.eu>
Date:   Fri Jan 14 11:48:16 2022 +0100

    Increase padding around questionnaire questions
╭╴ ((c906ecf...) $%) [virt]╶╮
╰ jeff@birdsong:transport_nantes $ 

```

JeffAbrahamson:
But, yes, the point was that you updated code in one commit, causing tests to fail, then updated tests in a second.  I'd argue you should update the tests in the same commit (i.e., squash them together).

Shriukan33:
> git checkout c906ecf

This is curious, because in facts, I did try to checkout on it : 
```cmd
(venv) benja@Benja-desktop:~/Documents/GitHub/tn_web$  git checkout c906ecf
error: pathspec 'c906ecf' did not match any file(s) known to git
```

JeffAbrahamson:
So I think the only thing left here is the `Context[...] = ...` suggestion ?

JeffAbrahamson:
Yes, that would be good.

I think you should also replace the 4 lines or so of context manipulation with something brief and direct, like
```
context['mailinglist'] = kwargs['mailinglist']
html = Template(template_string).render(context)
```

Shriukan33:
> Yes, that would be good.
> 
> I think you should also replace the 4 lines or so of context manipulation with something brief and direct, like
> 
> ```
> context['mailinglist'] = kwargs['mailinglist']
> html = Template(template_string).render(context)
> ```

Can't find a way to test that the context is persistent for now, can't find the logic actually. 
I think I'm fried maybe sleep will help.

JeffAbrahamson:
I think we can merge it without testing the context.  I remember thinking something was obvious on Saturday morning, but it's not this evening. :-(

JeffAbrahamson:
Um, tests fail at 331f3315c8e.  Might you have squashed the two commits so that the tests were updated at the same time as the code that invalidated the old tests?

JeffAbrahamson:
I've sent you a PR for the context simplification I suggested above.

PR #417, opened 2022-01-15
Og twitter template middleware

mcmikashi:
Hello I do 3 commit for solve this issue:
  1. I rename variable in the middleware
  2. I implement the solution on the base template in asso_tn
  3. I add unit test for testing the meta in the template

### DIFF
--- a/transport_nantes/asso_tn/middleware/default_context.py
+++ b/transport_nantes/asso_tn/middleware/default_context.py
@@ -8,15 +8,19 @@ from django.template import RequestContext
 # (now no longer used) Sites module.  We may or may not still need
 # this, at least in its current form.
 
+
 class DefaultContextMiddleware:
     social_media_context = {
         # This assumes images are served from STATIC.
         # We'll need to change something (in base?) when using MEDIA.
-        'og_image': 'velopolitain/v1.png',
+        'og_image_default': 'velopolitain/v1.png',
         'og_image_alt': 'mobilité',
-        'twitter_image': "asso_tn/mobilite-pour-tous.jpg",
-        'twitter_title': "Mobilitains - Pour une mobilité multimodale",
-        'twitter_description': "Nous agissons pour une mobilité plus fluide, plus sécurisée et plus vertueuse",
+        'twitter_image_default': "asso_tn/mobilite-pour-tous.jpg",
+        'og_twitter_title_default': ("Mobilitains - Pour une "
+                                     "mobilité multimodale"),
+        'og_twitter_description_default': ("Nous agissons pour une mobilité "
+                                           "plus fluide, plus sécurisée "
+                                           "et plus vertueuse"),
         'twitter_site': "@Mobilitains",
         'twitter_creator': "@Mobilitains",
         'twitter_card': "summary_large_image",
@@ -28,7 +32,8 @@ class DefaultContextMiddleware:
 
     def __call__(self, request):
         context = RequestContext(request)
-        settings.csrf_token = request.COOKIES.get(settings.CSRF_COOKIE_NAME, "")
+        settings.csrf_token = request.COOKIES.get(
+            settings.CSRF_COOKIE_NAME, "")
         response = self.get_response(request)
         return response
 
--- a/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
+++ b/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
@@ -90,11 +90,24 @@
 				{% endif %}
 
 			{% endblock styles %}
-
-			<meta property="og:title" content="{{ social.twitter_title }}" />
+			{# OG title #}
+			{% if social.og_title %}
+				<meta property="og:title" content="{{ social.og_title }}" />
+			{% elif social.twitter_title %}
+				<meta property="og:title" content="{{ social.twitter_title }}" />
+			{% else %}	
+				<meta property="og:title" content="{{ social.og_twitter_title_default }}" />
+			{% endif %}
 			<meta property="og:type" content="website" />
 			<meta property="og:url" content="{{ request.build_absolute_uri }}" />
-			<meta property="og:description" content="{{ social.twitter_description }}" />
+			{# OG description #}
+			{% if social.og_description %}
+				<meta property="og:description" content="{{ social.og_description }}" />
+			{% elif social.twitter_description %}
+				<meta property="og:description" content="{{ social.twitter_description }}" />
+			{% else %}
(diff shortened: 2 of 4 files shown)


Shriukan33:
Hey, using the Middleware to have a base context is a good idea :) 
You did a good job ! 

On the git history though, the commit 1a92407 seems like it be either squashed to f26f022 or put all base_template modification in the  1a92407, but not in two separate commits :)

JeffAbrahamson:
Indeed, cleaning up code prior to commit is an important service to future coders (including the commit author, who is likely human and may forget things).

I think it would make more sense to future readers of the code to squash all three commits into one and then clean up the commit comment so that it makes sense, describing succinctly in under 50 characters (cf. the readme) what the commit does and then explaining in simple declarative prose what problem the commit solves so that future readers have context.

mcmikashi:
> Hey, using the Middleware to have a base context is a good idea :) You did a good job !
> Hello thanks you.

> On the git history though, the commit [1a92407](https://github.com/transport-nantes/tn_web/commit/1a92407aab16d51e7a1f26e1f51a36be407adee1) seems like it be either squashed to [f26f022](https://github.com/transport-nantes/tn_web/commit/f26f022cdbe6e3826c63c9e18fdaee4c95362fa0) or put all base_template modification in the [1a92407](https://github.com/transport-nantes/tn_web/commit/1a92407aab16d51e7a1f26e1f51a36be407adee1), but not in two separate commits :)

> I don't know how this happens, but yes, it should be like you said.

JeffAbrahamson:
Hey, your markdown-fu could use improving.  Do you see that all your comments are in quotes?

JeffAbrahamson:
This is looking good, thanks.  @Shriukan33 ?

I'm thinking after we merge this, that a subsequent clean-up commit would be to write and use a template tag that takes three arguments and returns the first that is non-null.  Then we could write things like

```
<meta property="og:title" content="{% first_of_three social.og_title social.og_twitter_title social.ogt_twitter_title_default }}" />
```

which is a bit easier to parse for humans than the eight-line blocks we currently need.  Wdyt?

mcmikashi:
Hi, I thought that this will be easier to read, but that wasn't the case.

Thanks you.

JeffAbrahamson:
> Hi, I thought that this will be easier to read, but that wasn't the case.
> 
> Thanks you.

That would be for a second commit, however.

JeffAbrahamson:
Pulling your code this morning, I see tests don't pass.  They do pass against master.  Otherwise, I think this is mergeable.

JeffAbrahamson:
It would be helpful if you specifically indicate when you're looking for a new code review.  Otherwise, it's too easy to read your comments and think "good, sounds reasonable, he's making (slow) progress".

mcmikashi:
> Pulling your code this morning, I see tests don't pass. They do pass against master. Otherwise, I think this is mergeable.

Yes I got the same issue I fix it now.

mcmikashi:
Hello I make the change I create 2 template tags :

1. first_of_three for og and twitter title and description
2. first_of_for_image  og and twitter image and it's working for static and not static.

I asking for a code review.

JeffAbrahamson:
I see tests failing at both 697bca5 and at 885476c.

The first line of the commit comments don't give much idea to readers what is happening.  (Imagine you are not thinking about this issue, maybe have never heard of it.  What is a good 50 character maximum title for the commit?  Follow on with a textual description of what the goal of the commit is.  Possibly reread some of the good blog articles on what makes a good git commit message.)

Among other things, "OG" for open graph is common in context, but not obvious out of context.  And what does it mean for "twitter to be empty", for example.

```
885476c|2022-01-23|mickael anicette| (HEAD -> og-twitter-template-middleware) Update og twitter
697bca5|2022-01-18|mickael anicette| Fix OG and twitter
```

mcmikashi:
Hello I made the update for the test and change the commit message.

I asking for a code review.

Shriukan33:
Hello Mickael :) 

In c59c662
* I appreciate the docstring but it really is stating the obvious, it's not helpful toward understanding why it's there, why 3 strings, in what context you want to use it. Besides, it gives a very #This is a comment vibe
* The logic of the function is to return the first non-None argument, so might as well give write it this way :
```python
def first_not_none(*args):
    for arg in args:
        if arg is not None:
            return arg
    return None
```
* I think the naming of the function image_type could be better :
```python
def image_type(image: SA) -> str:
    """
    Return the static url if the image is a string
    else return the url of the ImageFieldFile class.
```
Out of context, what would we expect this function to return ? I'd bet on a type (ImageField ??), but would be unclear. Renaming it get_image_url would make it much clearer.

* Not sure why SA would be any useful for type hinting, where does SA even come from ? (I mean, what does S and A mean)
first_of_for_images has a typo in the name (I guess you implied four), but then again, the *args would have worked just as well. You did unnecessary repetition there.
* This function too has an opaque name.
* Description doesn't tell us what's supposed to be in the args. image_1 is the first image. Great. What image ?
* You modifiy `base_mobilitains.html` in two commits in a row, which should trigger something in your git-fu.(just in case, git-fu is a ref to kung-fu, meaning, your mastery of git)
* The commit message mentions "Add open graph and twitter tags", however no tags are named like this, and we fall again in the poor naming problem.


5577d91
The commit message's title is quite "???" what you do mean you fix twitter ?? 
```
"Improve defaultContext Middleware for social medias
Add a fallback to pages shared on social medias so if OG or 
Twitter metadata aren't set, they can fall back either to their counterpart's
 settings or our default settings."
```
-- Something along those lines would have give much more sense, in my opinion.


On the + side :
You did think about prototyping (type hints)
Tried to produce a Docstring (but not very helpful)
Tried to improve readability with the template tags, which wasn't required but nice to have

I really have the feeling that you're willing to do well, but the execution isn't quite there yet, we will work on that :slightly_smiling_face:

mcmikashi:
Hello Benjamin,

I take note on what I have to improve. Thanks for the sharing.

mcmikashi:
Hello, I made changes on all commit message and create two custom template tags and write the test for them.

I asking for a code review.

mcmikashi:
Hello,

The first commit is the same I just update the test to fit with the new topicblog modele. 
The second commit I added 2 custom template tags and add the test for them.

Shriukan33:
@JeffAbrahamson  tests pass ok once rebased on master, this _could_ have been one commit but the 2 commits formats eases the "where this comes from" reflexion, so I'm fine with it.. WDYT ?

JeffAbrahamson:
Yeah, there's quibblable stuff, but LGTM overall.

PR #422, opened 2022-01-17
Drop the end of year banner

JeffAbrahamson:
We added a button at the end of 2021 to encourage donations.  (We
received none.)  It's not something we intended to keep long term.
Let's drop it now that the new year has started.

### DIFF
--- a/transport_nantes/topicblog/views.py
+++ b/transport_nantes/topicblog/views.py
@@ -101,11 +101,6 @@ class TopicBlogBaseView(TemplateView):
         context['page'] = tb_object
         tb_object: self.model  # Type hint for linter
         context = tb_object.set_social_context(context)
-        context["banner_is_present"] = True
-        context["banner_text"] = ("C’est grâce à votre soutien que nous "
-                                  "pouvons agir en toute indépendance.")
-        context["banner_button_text"] = "Je participe"
-        context["banner_button_link"] = reverse('stripe_app:stripe')
 
         return context
 


Shriukan33:
Maybe you want to turn off the banner then setting banner_is_present to false is enough. 
If you want to get rid of the whole banner feature, then you also need to delete it in the base_mobilitains.html template  :slightly_smiling_face:

Shriukan33:
> Agreed, but we would be unlikely to re-enable the same banner in the future, so it would just be dead code executing on each request for no reason.

Ok, then I'd rather have the base template updated as well, so it doesn't try to render something that doesn't exist anyway

JeffAbrahamson:
It's in a `{% if ... %}` block.  I thought the point of the mechanism was to enable us to activate it and deactivate it from time to time.  Did I misunderstand?

Shriukan33:
Nah you're right. But as you delete the code that enables it, I wonder why keeping it then ? Else we could make banner_is_present = False and put the rest under a if condition (if banner_is_present: ... ) 

This way only the bool is evaluated, which is not that expensive. 

The only thing that bugs me is why keeping the HTML in the template if the code responsible to make it appear is deleted. Enabling it again would mean writing again the very same code. 
So in my head it's either we delete both, or we put that under a condition so we can enable it later if we feel like it ? 

Is there something I'm not seeing ?

JeffAbrahamson:
Yeah, next time we'd write code like
```
context["banner_is_present"] = True
context["banner_text"] = "New stuff"
# etc.
```
But what we wouldn't have to do is think about how the mechanism works or how to format it.

Shriukan33:
> Yeah, next time we'd write code like
> 
> ```
> context["banner_is_present"] = True
> context["banner_text"] = "New stuff"
> # etc.
> ```
> 
> But what we wouldn't have to do is think about how the mechanism works or how to format it.

Ok, so you'd rather delete the code than making it conditional ? 
Either way, it's easy enough to get back to this, even in 6 months. Adding a var to context shouldn't be too hard haha

LGTM

Issue #413, opened 2022-01-14
slug fields limitations


JeffAbrahamson:
* slug fields are limited to 50 characters

Is this necessary?  If it is, we should document it in the model.  If it's not, we should remove the constraint.

* pasting non-ascii doesn't do the right thing

For example, pasting "mobilité" gives "mobilit".

JeffAbrahamson:
For the length, the [default is 50](https://docs.djangoproject.com/en/4.0/ref/models/fields/#slugfield).  Once this morning's PRs are merged, I'll add a `max_length=90`.  Probably 80 is enough, but 90 gives extra space.  (Too long makes for weird copy/paste.)

Shriukan33:
This is solved, we sanitized slug inputs in 5792be7dd03f72e9c8936cf3f9bc28a65c8c9cb3 , closing

PR #423, opened 2022-01-17
Increase slug length to 90 characters

JeffAbrahamson:
The default is 50, which is slightly short.

From 70-80 is probably adequate, but better to allow slightly longer
than needed.

Part of #413.


Shriukan33:
LGTM

PR #430, opened 2022-01-18
Delete all references to TBv2's Content_type field

### LINKED ISSUE #429: Remove Content-type for TBv2
Content-type removal :
- the edit view sets the content-type to "article" if it's new (i.e., if it's not already set)
- the edit view sets the dropdown
- the form loses the value
- the model loses the value
- delete the javascript that supports all that
- test that you haven't broken anything, that the field deletion from the model doesn't cascade any deletes, etc.

Shriukan33:
Removes references to content-type field.

Closes #429


JeffAbrahamson:
Have you explored the consequences of removing a field with `on_delete=models.CASCADE`?

Shriukan33:
Removing the item does delete in cascade, deleting the field however doesn't  delete previous records. I tried it on my dev db. 

If we delete TBT item, it deletes the TBI that have a FK pointing toward that specific item (If I understood the doc correctly). However straight up deleting the Table and altering the content_type field is replaced with the previous Pk 
For instance I tried (on another branch) to delete ContentType table and changed content_type to forms.Charfield. 
Instead of being = to a ContentType, it was equal to "1". 

So it doesn't seem to delete previous records. We could also change the on_delete param and set it to SET_NULL for example.

In this specific PR, deleting the fields didn't erase any of my TBItems.

Shriukan33:
Now this being said, I undestood, after the opening of this PR, that you wanted to keep a Content type but don't want it to be in a table. 
Is that correct ?

JeffAbrahamson:
Each of the TBItem, TBEmail, TBLauncher etc. will have pointers to (or names of) templates.  Does each have its own template table that, perhaps, only contains 1-2 records for a long time?

Maybe we should get rid of the template table, too, and just have a map (dict) that configures templates, since template creation and modification is otherwise in get.  I think that configuration structure is only used by the object editor.

```
{template_name: {
  'active': True,
  'fields': (
      'cta_1_slug',
      'cta_1_label',
      ...
    ),
  }
}
```

JeffAbrahamson:
I would like to make sure that an inadvertent clean-up of an unused record in an unused table doesn't delete all of our content in an accidental cascade.

JeffAbrahamson:
I think, like `TopicBlogPage`, that we should leave an empty model behind.  Deleting the model doesn't delete the table, and so it's useful to have the trace of what the table came from, that it wasn't an accident, etc.

JeffAbrahamson:
Where do you set the template name list?  Is this still in javascript?

JeffAbrahamson:
With apologies, I think we should update this slightly more:

* make the content-type model empty (but not absent)
* make the templates model empty (but not absent)
* replace the existing template records (article is the only one I think) with a map structure along the lines of above
* replace the `template` foreign key with a simple string called `template_filename` (non-nullable field, with migration value equal to `template.template_name`)
* set the template list manually from the keys of that table

Wdyt?

JeffAbrahamson:
Otherwise, well done!

Tricky designs like this have a habit of morphing somewhat as we go.  We see further as we climb higher...

JeffAbrahamson:
Good thinking, updating the README.

Suggestion: leave the "congratulations" at "localhost", no need to encourage binding to 0.0.0.0.

Do be sure to test that this doesn't trigger a cascading delete. ;-)
It looks like it won't since it only deletes columns and not rows, but ...

Minor point:  I got a bit confused why the thing is called `template` (as a foreign key) and not `template_name` (is a string, a name).  In addition, changing name when you refactor serves the double purpose of calling out errors automatically.

Shriukan33:
I think you commented on the wrong place but I'll assume this is PR for step 3 : 

I precise the 0.0.0.0:8000 stuff because this is in the previous instructions, besides that's the address to go to when you use vagrant. going to 127.0.0.1:8000 doesn't work when using vagrant. 

I'm in the process of renaming it template_name for clarity purpose.

Shriukan33:
This PR wont be merged, as we widened the scope of this to TBTemplate, it's decouple-content-type-455 that is going to handle this

PR #432, opened 2022-01-19
Fix tbv2

Shriukan33:
Changes :
* items are displayed as withdrawn when they are not [Fixed]
* Item's publisher email was showing the mail of current user instead
of actual publisher [Fixed]
* Fix edge case where wrong item would be served. Would happen if we modified a previously withdrawn item through django
admin. date_modified attribute would be the most recent and be served while the publication date didn't update.
* Item with the latest publication_date and date_modified is tagged
as current page. [new]
* Items with publication_date that follow the first item are tagged
as withdrawn.[new]


Shriukan33:
Actually I noticed that the publisher's name is the author of the article. We don't have yet a field to save the name of the person that clicks "publish"

JeffAbrahamson:
Commit comment: "fix admin list panel for TBv2 and improvement"

1. Should start with capital letter.
2. I don't know what this means, and would be even less clear reading a short-format git log with tens of commits.

JeffAbrahamson:
No tests test what has changed.

JeffAbrahamson:
This seems wrong.  The page was created (modified) after it was published.

image

JeffAbrahamson:
I think the list view text is a bit confusing, and the logic to produce it is also a bit confusing.  This could be simplified thus:

```
[1234] (page actuelle)
Cré mardi 18-01-2022 à 18:02:03 par joe@example.com
Publié mardi 18-01-2022 à 18:03:05  (première publication : mardi 18-01-2022 à 16:08:09)

[1232]
Cré mardi 18-01-2022 à 16:07:08 par joe@example.com
Publié mardi 18-01-2022 à 16:08:09

[1228]
Cré lundi 17-01-2022 à 04:06:08 par jon@example.com
Brouillon

```

So,
* you only need to show first publication time for the currently published item,
* you always show creation time
* you always show published time if published, else "draft"
* the creation use is shown as such, because it's never been the publication user

JeffAbrahamson:
You sometimes say "Publié" and sometimes "A été publié".  We should stick with one.  I vote for the terse, "Publié".

JeffAbrahamson:
I think it would be modestly clearer if the list view and the view-one selector used as close as possible the same query.  It makes clear that we're selecting the same thing.

The way to do this is to use negatives in view-one and select `first()`.

Shriukan33:
Note that no tests have been written for now, as I thought about this PR more like continuous review / change process. 
tests are to be done, just wanted to get it ok for you before and then testing it. 



> You sometimes say "Publié" and sometimes "A été publié". We should stick with one. I vote for the terse, "Publié".

This was a way to say "Has been published (but isn't anymore)", but yeah it introduce some complexity that doesn't help a lot anyway so I sticked with "Publié"

PR #439, opened 2022-01-20
Add notes section to TB objects

### LINKED ISSUE #437: Add notes section to TBObject
Let's add a text field to TBObject and a field to the edit forms so that author's can add their notes, such as research links and URL's, directly to the articles they write.

These are things we won't display, but it lets us track and hold on to them more easily.

Shriukan33:
These fields wont be displayed, they are only there
for references for the editors.

Can consult them in the edit form or through django admin for now. 

Closes #437


JeffAbrahamson:
Why does the commit have _two_ migrations?  Can you clean that up?

JeffAbrahamson:
"notes section" describes something visual, not a field.  A field name like "author_notes" would make much more sense.

On the form, perhaps we could provide feedback that it is internal and won't display to users?

JeffAbrahamson:
The social media section of the edit form is already really long.  Perhaps we should put this on the admin page instead, or even better, create a new section for it?

Shriukan33:
Added a new tab, holding the new field
Made it one migration
Renamed it author notes
Added a precision in the form informing the user that these notes aren't displayed on the website.

JeffAbrahamson:
LGTM

PR #441, opened 2022-01-20
Fix and improve TBObject list view

### LINKED ISSUE #438: Fix TBObjectList
Something is amiss.

More importantly, we aren't giving good visibility on the current publishing data model.

JeffAbrahamson:
Correctly show published item if there is one.
Sort correctly (by pub date, then mod date).
Show creating user as such, not as publisher.

Closes #438.


Shriukan33 (commented):
The wording "modifié" is debatable now, as we create new items each time we make a modification. We can't have separate date_modified and date_created in the end.

Shriukan33:
It solves our TBIList problem for now, so LGTM

Issue #442, opened 2022-01-20
tb edit cancel button is broken (for Jeff, on ff)


JeffAbrahamson:
It seems dubious that that button needs javascript.

Every time you use javascript, there are two questions you have to ask yourself:
* What do I gain, because this will surely be a _bit_ heavier, and
* What will be the experience of people for whom this doesn't work

That last happens more often than you'd think: visual impairment, screen readers, text-only browsers or people browsing in secure environments, really old browsers.

If you get something that's at the heart of your business, it's easily justifiable.  But a cancel button?

PR #444, opened 2022-01-20
Rename tab in TBI and clarify buttons

Shriukan33:
The cancel button could be misunderstood, changing the wording
clears that confusion.
Added a button to return to the list of articles, which is the
function one could expect from the cancel button.

Part of #442


JeffAbrahamson:
LGTM.

That said, might be better to go back to the list _for that slug_ and not the general list.

JeffAbrahamson:
No prob.

PR #450, opened 2022-01-25
Fix reference before assignement on form_valid 

Shriukan33:
The form_post_process only makes sense for already existing objects,
moving it to a condition where we know we're editing an existing object
prevents us to referencing to one when there is none.

This commit would be cherry picked once we go back to master as the deployed branch.


Shriukan33:
precision - merge is on telenantes branch, not master, for it to be deployed soon. Beta first if you're ok with the code

JeffAbrahamson:
I don't think the indent is right.  A post-processing action might be appropriate for a new object as well.  Indeed, the one action that exists tests for the object existing or not.

I think the right thing would just be to initialise `is_existing` to `self.Model()` in the `else` clause of the above `if`.

It's a bit hard to be certain, of course, because there's only one...

Shriukan33:
Like so ?

Issue #455, opened 2022-01-26
Stage database to transition to state without ContentType and TopicBlogTemplate models


Shriukan33:
In preparation for the next TB projects (Email, launchers...) we're merging the models ContentType and TBTemplate to attributes of TBItem. That means that instead of ForeignKeys pointing toward these models we will have attributes that provide the same functionalities. 

To achieve this transition without losing information in the database, we need to fetch the template name for the current ForeignKey, otherwise its value will be lost once the relationship disappear.

PR #456, opened 2022-01-26
Step 1- Save template.template_name in Charfield

Shriukan33:
In preparation for the next TB projects (Email, launchers...) we're merging the models ContentType and TBTemplate to attributes of TBItem. That means that instead of ForeignKeys pointing toward these models we will have attributes that provide the same functionalities.

To achieve this transition without losing information in the database, we need to fetch the template name for the current ForeignKey, otherwise its value will be lost once the relationship disappear.

This is a first step.

Couldn't get to default to template.template_name's value (because it's a ForeignKey object at this point, not a TBTemplate). 
So I wrote a small function that does the fill, but requires to use the shell to perform. 

Input commands : 
```python
(venv) benja@Benja-desktop:~/Documents/GitHub/tn_web/transport_nantes$ python manage.py shell
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.12.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from topicblog.models import TopicBlogItem                                                                                                                                          

In [2]: qs = TopicBlogItem.objects.all()                                                                                                                                                    

In [3]: for i in qs: 
   ...:     i.fill_template_name() 
   ...:                                                                                                                                                                                     

In [4]: exit
```

Result : 
image

Tests pass OK


JeffAbrahamson:
**Reviewing stage 1**

The git comment doesn't provide any context.  What are you doing and why?

Don't assume that the issue is available.  A reader might not have access to the web at the moment of reading.  Github as a company may no longer exist or have changed policy.  Another offering may have become more interesting for some reason.  Say what you need to say in the issue.

The issue tag is important whilst github exists and also permits `git log --grep` on the issue tag.

```
git log --grep '#455'
```

The transition script looks right, but why not add a path to `urls.py` and make the `fill_template_name` function a view accessible only to super users?  That way there's an historical trace of what we did.

JeffAbrahamson:
**Reviewing stage 2**

The git comment doesn't provide enough context.  The first line ends with a period.

Why do we need a stage that doesn't work?  The point of staging is usually to keep the system operational through a change.

Why do you want to change the template name to be called template?  It's a string representing the name of the template, after all.

I think this commit could probably be dropped with no particular loss.  Wdyt?

JeffAbrahamson:
**Reviewing stage 3**

Checkpoint commit.  I'm a bit lost on what you're trying to do here. ;-)

If you think it should be clear, give me a call or tell me more.

I think you only need to stages.  In the first you do what you did in the first.  In the second you add the dicts of dicts that replace the old template model, you remove the foreign keys that point to `TBTemplate`, and you make sure the rest of the code that referred to `item.template` still works if `template = whatever_the_dict_is_called[item.template_name]`.

Finally you remove the fields from `TBContentType` and `TBTemplate` and put in an historical note like at `TBPage`.

Shriukan33:
> **Reviewing stage 1**
> 
> The git comment doesn't provide any context. What are you doing and why?
> 
> Don't assume that the issue is available. A reader might not have access to the web at the moment of reading. Github as a company may no longer exist or have changed policy. Another offering may have become more interesting for some reason. Say what you need to say in the issue.
> 
> The issue tag is important whilst github exists and also permits `git log --grep` on the issue tag.
> 
> ```
> git log --grep '#455'
> ```
> 
> The transition script looks right, but why not add a path to `urls.py` and make the `fill_template_name` function a view accessible only to super users? That way there's an historical trace of what we did.

I edited the post to include the issue in question ! :slightly_smiling_face: 

I assumed that the function itself is an historical trace of what we did ? Besides, the shell interface is already exclusive to admins so I didn't feel there was the need for an url for a one-time function.

Shriukan33:
> **Reviewing stage 2**
> 
> The git comment doesn't provide enough context. The first line ends with a period.
> 
> Why do we need a stage that doesn't work? The point of staging is usually to keep the system operational through a change.
> 
> Why do you want to change the template name to be called template? It's a string representing the name of the template, after all.
> 
> I think this commit could probably be dropped with no particular loss. Wdyt?

The 2nd step (not stage per say) was more a way to make a transition to the 3rd branch, because this 3rd branch assumed a "template" field, and historically that's how we called that field.
Besides it does the job of transitioning between a FK to a populated Charfield so we shouldn't drop entirely this step. 

I could integrate this commit to step 3, but wanted to explain the transition here. 

Will reword the topic and the commit

JeffAbrahamson:
> I assumed that the function itself is an historical trace of what we did ? Besides, the shell interface is already exclusive to admins so I didn't feel there was the need for an url for a one-time function.

It's more about documentation than security.  It also lets us review precisely what we're going to do and to be sure we don't forget a step.  Although copy-pasting a playbook is pretty close to that level of certainty.

JeffAbrahamson:
> > **Reviewing stage 2**
> > The git comment doesn't provide enough context. The first line ends with a period.
> > Why do we need a stage that doesn't work? The point of staging is usually to keep the system operational through a change.
> > Why do you want to change the template name to be called template? It's a string representing the name of the template, after all.
> > I think this commit could probably be dropped with no particular loss. Wdyt?
> 
> The 2nd step (not stage per say) was more a way to make a transition to the 3rd branch, because this 3rd branch assumed a "template" field, and historically that's how we called that field. Besides it does the job of transitioning between a FK to a populated Charfield so we shouldn't drop entirely this step.
> 
> I could integrate this commit to step 3, but wanted to explain the transition here.
> 
> Also this would be merged after step 1 but not deployed on its own, would require step 3 to work
> 
> Will reword the topic and the commit

It's more about not putting commits on master than we know in advance will not work correctly.  Especially where we know in advance they will break serving the main content of the site!

Shriukan33:
> **Reviewing stage 3**
> 
> Checkpoint commit. I'm a bit lost on what you're trying to do here. ;-)
> 
> If you think it should be clear, give me a call or tell me more.
> 
> I think you only need to stages. In the first you do what you did in the first. In the second you add the dicts of dicts that replace the old template model, you remove the foreign keys that point to `TBTemplate`, and you make sure the rest of the code that referred to `item.template` still works if `template = whatever_the_dict_is_called[item.template_name]`.
> 
> Finally you remove the fields from `TBContentType` and `TBTemplate` and put in an historical note like at `TBPage`.

What you are describing is actually what I do in this 3rd branch  ! :slightly_smiling_face: 
Basically functions and tests pass in the current state, you can save items using the form, and it displays normally. You can publish using the listview.
However the form itself doesn't have a list of available templates and content types, you have to write manually "topicblog/content.html" to have it as a template, which is not reasonable from an user standpoint. 

To get this bit I need to fiddle with the ModelForm, Ajax, and the view. 

Hopefully I can make this clear in the 3rd branch !

Shriukan33:
> > > **Reviewing stage 2**
> > > The git comment doesn't provide enough context. The first line ends with a period.
> > > Why do we need a stage that doesn't work? The point of staging is usually to keep the system operational through a change.
> > > Why do you want to change the template name to be called template? It's a string representing the name of the template, after all.
> > > I think this commit could probably be dropped with no particular loss. Wdyt?
> > 
> > 
> > The 2nd step (not stage per say) was more a way to make a transition to the 3rd branch, because this 3rd branch assumed a "template" field, and historically that's how we called that field. Besides it does the job of transitioning between a FK to a populated Charfield so we shouldn't drop entirely this step.
> > I could integrate this commit to step 3, but wanted to explain the transition here.
> > Also this would be merged after step 1 but not deployed on its own, would require step 3 to work
> > Will reword the topic and the commit
> 
> It's more about not putting commits on master than we know in advance will not work correctly. Especially where we know in advance they will break serving the main content of the site!

Can squash commits, but I feel the review is already hard enough to not squeeze more changes in the same commit.

JeffAbrahamson:
If the dropdown has static content (i.e., won't change whilst the form displays), then you can pre-populate in django.  We do this in `survey/`.

JeffAbrahamson:
Let's get the first stage done and sorted and then think about the second.  It will maybe reduce cognitive load for you. ;-)

PR #457, opened 2022-01-26
Step 2 - decouple template model 455

Shriukan33:
:warning:  TO MERGE AFTER STEP 1 IN PR #456 :warning: 

This second step, once the template_name field is filled with the value of template.template_name, consist of deleting the previous template field, as we saved the value we wanted in template_name and delete the views created for the occasion of the Transition.

This step opens the way to complete TBT separation from TBItem in next branch.


JeffAbrahamson:
LGTM.

If I don't miss my mark, however, this is not yet mergeable because it will render all TBItem's undisplayable, no?

Shriukan33:
> LGTM.
> 
> If I don't miss my mark, however, this is not yet mergeable because it will render all TBItem's undisplayable.
> 
> I think the intent, however, is that this PR doesn't get merged, rather we merge step 3 directly.

Should be merged but not deployed until step 3 is also merged.

Step 3 is also based on step 2 so merging step 3 should be equivalent

JeffAbrahamson:
I'm going to close this so we don't accidentally merge it.
The merge will happen on the step 3 close.

PR #458, opened 2022-01-27
Step 3 - decouple template model 455

Shriukan33:
:warning: Draft while step 1 and 2 haven't been processed ! :warning: 

In prevision of further changes, we're deprecating the ContentType
and TBTemplate.
Their features are now transfered to TBItem as attributes.

Every reference to ContentType and TBTemplate have been removed except
the models. Because the table isn't deleted in the database, leaving
the model leaves a trace of the reason of their existences.

Also updated the edit form to reflect the change.


Shriukan33:
Because there are some changes to make after your feedbacks on step 1 & 2, this branch is going to be rebased as well in a soon future, however you have the gist of it here for review.

JeffAbrahamson:
LGTM except for one rather important bug: it forgets the template name on every edit cycle.

JeffAbrahamson:
Reminder to self to GET `/tb/transition` just before deploy.

Shriukan33:
Weird, I tried it and could publish items no problem going through the form, something must be up with the db migration maybe.

Will check that once I'll be able to do so.

If you're trying to publish from a non superuser account, you're displayed a 403.

Shriukan33:
> Reminder to self to GET `/tb/transition` just before deploy.

The transition view has been deleted, iirc

JeffAbrahamson:
> > Reminder to self to GET `/tb/transition` just before deploy.
> 
> The transition view has been deleted, iirc

It has, but before deploying (and so deleting the transition view) on beta or prod, we will need to re-run the transition, because any article edits in the mean time will not reliably have the `template_name` field set.

JeffAbrahamson:
> Weird, I tried it and could publish items no problem going through the form, something must be up with the db migration maybe.
> 
> Will check that once I'll be able to do so.
> 
> If you're trying to publish from a non superuser account, you're displayed a 403.

I don't even get a publish button.

PR #461, opened 2022-02-02
Add permissions 460

### LINKED ISSUE #460: Enforce permissions
Several permissions are already set in TBItem and TBEmail models, we want them to be effective and checked in the views.

There are 3 distinct groups for now : 
Editors which : 
* may create and edit TBItems
Publisher : 
* Editors + may publish other's TBITems
Admin :
* All rights, notably can publish themselves. This isn't an explicit group, as @permission_required will return True for superusers.

Those groups need to be created manually in the Django admin console, the rights are then checked in the views.

Shriukan33:
Mergeable since Step 3 has been merged ! 

* I did add the permissions to TBI, it's been a bit more painful to setup than I thought it would be, mostly because I tried to use method decorator inside a class based view, while I needed a Mixin. Then had to create my own PermissionRequiredMixin for TBIViewOne. Overall it's been mostly about understanding how to make it all work, further permissions should be much quicker to implement ! 
* The automated tests aren't written yet, I manually tested and it seems to react as we expect : You can't self publish your content even if you are a Publisher (only admins can do so), however being a publisher allows you to publish someone else's items. You can try it yourself. 
* Fixed a small problem on the edit form, which would always display None even on edit (while there was a value in db) 
* I fixed TBI List template to show more accurate information, like the Publisher email's (added the field) instead of its author (when talking about who published what) and also fixed the "Brouillon" / 'Publié ... (retiré)' display, which was incorrect.
* Updated tests to pass with permissions

Closes #460


Shriukan33:
> Looks like this fixes the template name saving bug, but I still can't publish.
> 
> A couple minor white space errors in base_topicblog.html.

@JeffAbrahamson I believe the issue was about the body field not completed, as we discussed last week.
I've cleaned the white spaces you mentioned. 

Also updated the tests so they are green with the new permissions system. :slightly_smiling_face:

JeffAbrahamson:
LGTM.

One comment for future:

b72e9e5 aka 503832d - It's usually better UI design to lead users to do the right thing by not even offering for them to do the wrong thing.  Then you don't need 4xx errors or the messages that go with them, just 500 and say "my bad" if bad things happen.

Issue #462, opened 2022-02-03
TBItem quick issues


JeffAbrahamson:
Quick notes of quick things to fix as I'm sorting other details this evening:

* `servable` doesn't exist any more, but it's referenced in `base_topicblog.html`
* the `template_name` pop-down is lost on new edit, though I think that's fixed once the permissions branch is merged
* `views.py` still computes `servable_object`, but it's a bit unclear (to me, this evening) why
* tests are missing (as noted elsewhere)
* there's a bug in the list sort order (I'll send a PR)

JeffAbrahamson:
status @Shriukan33 ?

JeffAbrahamson:
@Shriukan33 Is this closable?

PR #463, opened 2022-02-03
Fiddle with sort order

JeffAbrahamson:
For choosing the publishable page, it doesn't matter a wit in what
order things were modified, only in what order they were published.
So only sort on publication date.

For the list view, it turns out to get confusing when including
publication date in sort order.  Our heads mostly remember when we
modified things, and more recent things should come first.  So do
that.

Part of #462.


Shriukan33:
So the list view doesn't show what item is published first, only the last modified?

JeffAbrahamson:
We _mostly_ proceed in modified order.  When you get a list like [this](https://www.mobilitains.fr/tb/admin/t/list/benevolat-operation-pieton/), it's really disturbing to have the order not be by modification.

Shriukan33:
Yup, sounds good to me

Shriukan33 (approved):

PR #467, opened 2022-02-08
Fix slight time skew on save.

### LINKED ISSUE #435: TBItem publish function has time skew
Publish potentially calls `now()` twice for two values that should, in that case, have the same value.  But they could have different values if the clock ticks during execution.

Shriukan33:
Calling .now() twice would return close but not same values for
values that would be expected to be equal.

Closes #435

### DIFF
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -385,9 +385,12 @@ class TopicBlogItem(TopicBlogObjectBase):
 
         """
         if self.is_publishable():
+            now_timestamp = datetime.now(timezone.utc)
             if self.first_publication_date is None:
-                self.first_publication_date = datetime.now(timezone.utc)
-            self.publication_date = datetime.now(timezone.utc)
+                self.first_publication_date = now_timestamp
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
LGTM.

JeffAbrahamson:
Oh, maybe I am just hungry.  It's right, but it would be _clearer_ if you did what I suggested.

Shriukan33:
I did reuse the now() but maybe it wasn't obvious at first read, does it look better now ?

JeffAbrahamson:
Cool, that gets the logic right.  Do you see how this makes super clear to the reader that only one possible timestamp is being use din all of it?

Now, just one really minor pythonic thing: naming a variable exactly the same as a common function is grammatically correct, python won't be confused, but consider that some readers will be used to seeing `from datetime.datetime import now` and so might misread.  That would argue for calling the variable `now_datetime` or `now_timestamp` or `timestamp` or something like that.

JeffAbrahamson:
LGTM, variable rename or not.

PR #468, opened 2022-02-08
Delete TN-era functions and reference to it

### LINKED ISSUE #367: Rationalise the "M" functions
This is a clean-up issue.

We have a number of templates and functions that duplicate, the key signal being a capital "M" on the end of the names.

We should scan the code and clean these up.  In a first pass, if there is an M function and a non-M function, let's confirm we can remove the non-M function and do so.

Once that is bedded, let's rename all M functions by removing the M.

Shriukan33:
we use mpanels a bit everywhere now, panels were the templatetags
that rendered now inexistants templates used by Transport-Nantes.

Updated a template that still refered to panels instead of mpanels,
although it belongs to a deprecated app.

Couldn't find other references to tn specific code.

Closes #367


JeffAbrahamson:
LGTM.

PR #469, opened 2022-02-08
Add stripe_event_id to donations model

Shriukan33:
Tracking the event_id allows us to link a donation to a Stripe event.
Besides, it allows us to check if we already saved the information
of an event, in the case we re-send some. It avoids duplicates.

Part of #233


JeffAbrahamson:
The log line should be "Creating donation".

This needs tests, all the more so because our current issues concern reliability.

JeffAbrahamson:
LGTM.

PR #471, opened 2022-02-09
Refactor Topicblog tests

mcmikashi:
Fusioning all topicblog tests into one file

Part #460


transport_nantes/topicblog/tests_topic_blog.py
```diff
@@ -0,0 +1,1338 @@
...
+                         number_of_items,
+                         msg="The list of items must be the same length as "
+                         "the number of items with the corresponding slug "
+                         "in the database."
+                         f"\nnumber of items : {number_of_items}"
+                         "\nnumber of items in the list : "
+                         f'{len(response.context["object_list"])}')
+
+
+class TBIModel(TestCase):
```
Shriukan33, line 1301: Cette partie là n'est pas un refactoring, il faudra le mettre dans une autre pull request dédiée à l'amélioration de la couverture de tests

Shriukan33 (changes requested):
TBIModel should be in another model, other than that, the refactoring didn't eliminate that many lines unfortunately. 

Any ideas about what could be withdrawn ? It gets pretty lengthy because there are many cases for various users. 

Maybe we could reduce using a structure like this : 
```Python
user_expected_code = {
    self.user: 302,
    self.unautg_user: 403
    self.permitted_user: 200
}
for client, expected_code in user_expected_code.items():
    response = client.get(...)
    self.assertEqual(response.status_code, expected_code)

```
@JeffAbrahamson What do you think ? Would it improve readability for you ?

JeffAbrahamson:
The diff removes 173 LOC (12%).

@mcmikashi Good observation about short-circuiting.  That said, I agree with the @Shriukan33 proposal.  Why?  Because I don't really think most devs read _all_ the errors, internalise them all, and then go fix things.  Most people look at a couple, maybe get a feel for the whole, and then go fix what fits in their heads.

In addition, a good dev will see through to other errors that are likely and will lazily save himself the effort of rerunning the test if it's obvious the bug applies to multiple cases.

Also, we read code much more than we write it.  Favouring readability is usually the right approach.

JeffAbrahamson:
Rather than commenting at L29, maybe you could assign a variable like `main_page_slug_name` and use that.  Then we don't have to offer excuses for what ever its value is.

For a second commmit, but not that the comments about sort key are no longer correct.

Shriukan33 (changes requested):

Shriukan33:
LGTM
@JeffAbrahamson  ?

JeffAbrahamson:
The PR says it's based on branch `mcmikashi:test-topicblog`, but I also see a branch `origin/test-topicblog` that is not the same.  Why are there two same-named branches that are out of sync?

JeffAbrahamson:
Convention with the word "must" in specifications is to capitalise it.  Avoid complicating the logic with "must only ... if", which can lead to misreading.

So say (L 390)
```
Items which have a slug view page MUST return 200 if:
```
(In addition, the "only" there was superfluous: they can't return more than one code.  Also, English doesn't like spaces before punctuation, unlike French.)

L 397,
```
View page MUST NOT return 200 if:
```

JeffAbrahamson:
Otherwise, LGTM.  And these items may be fixed in a new commit if you prefer to merge this one, since my remarks are all about comments.

JeffAbrahamson:
Fair enough, what was confusing was the request that I review but two non-identical branches.

There's not much benefit to your maintaining two branches like this.  If you want to push branches to the main repo, just do that and drop your clone.  If you prefer to work with your clone, then make PRs on the clone.

But the PR says you're using your clone, so that's what I review and that's what the PR would merge.  The branch in the main repo is just noise in that case.

mcmikashi:
> In `test_item_with_slug_edit` you reuse the same variable name. This is a fragile coding style. It would be better to use different names.
The variable name that i reuse is users_expected  I should write users_expected_1 like this ?

mcmikashi:
I made the change for the comment.

mcmikashi:
I made the change for the variable.
This is good for review.

JeffAbrahamson:
Thanks, LGTM.

PR #477, opened 2022-02-14
Add django github-action

### LINKED ISSUE #343: Automate tests before merge with GitHub Actions
https://www.hacksoft.io/blog/github-actions-in-action-setting-up-django-and-postgres

It sounds like a good idea to have github run the `python manage.py test` command before merging to master, just so no preventable failure slip through the net.

mcmikashi:
Automating test and coverage before push and pull request

Fix #343

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -1,0 +1,46 @@
+name: Django CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.8]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/django.yml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
Shriukan33, line 12: Notre server tourne sur Ubuntu 20.04 (Focal Fossa) si je ne me trompe pas (@JeffAbrahamson ?), est-ce que c'est possible de préciser la version ? Ou pas ?

JeffAbrahamson:
Is the context that the yaml causes code to run in a github sandbox, and so we're just setting up that sandbox and telling it how to run the coverage report?

That is, this is code that we don't run on our hosts, they do on theirs?

If so (and please confirm explicitly), LGTM.

mcmikashi:
> Is the context that the yaml causes code to run in a github sandbox, and so we're just setting up that sandbox and telling it how to run the coverage report?

Yes It will only run on push or PR on master for checks if all tests is good and show the coverage.

> That is, this is code that we don't run on our hosts, they do on theirs?

They do on theirs.

PR #481, opened 2022-02-16
Add google-chrome to Vagrant VM

Shriukan33:
Selenium requires a local install of Google-chrome to run.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -88,6 +88,10 @@ Vagrant.configure("2") do |config|
     sudo mkdir /tmp/django-tn-log/
     sudo touch /tmp/django-tn-log/tn_web.log
     sudo chmod 777 /tmp/django-tn-log/tn_web.log
+    sudo sh -c 'echo "deb [arch=amd64] https://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/google-chrome.list'
+    wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+    sudo apt-get update -y
+    sudo apt-get install google-chrome-stable -y
 SETUP
   # To run in vagrant:
   #  ROOT_URLCONF=transport_nantes.urls_m DEFAULT_SITE_ID=1 \


JeffAbrahamson:
Hmm, so this will need saltstack support as well in order for the PPA and package to be installed on beta and prod?

I presume github will run this without issue?

This looks at first glance like the full chrome browser.  Will this mean tests require a fully functioning X environment to run?

Shriukan33:
> Hmm, so this will need saltstack support as well in order for the PPA and package to be installed on beta and prod?

I honestly don't know how this goes for saltstack, I made it so the vagrant venv runs the tests, are there key differences ? 

> I presume github will run this without issue?

Github Action's Django CI already runs the tests OK, you can see the details of the report [here](https://github.com/transport-nantes/tn_web/runs/5214615454?check_suite_focus=true)
image


> This looks at first glance like the full chrome browser. Will this mean tests require a fully functioning X environment to run?
It is the full chrome browser. 
I looked it up, and providing only the driver wont work, as it's searching for the binary version as well. 
It needs to have a fully working browser, to perform browser actions, sounds coherent. Note that it runs headless though, thanks to the --headless option added in the initialization of the test.

JeffAbrahamson:
If a deploy should run tests, then it would need to be installed on the machine to which we're deploying.

As I think about it, I'm actually not sure.  This is python, once we do the `git pull`, we'd better be sure that the code is good.  Unlike a compiled language where you compile and then install to a separate place.

So I guess we're ok without saltstack support for this in production.

What's curious is that I do have `chromium-browser` installed.  And I have a version of chromium installed via snap.  So now it somehow needs a _third_ version of chrome.

Can we the solenium tests auto-disable themselves if the env isn't ok?  Is that even a reasonable strategy?  I'm thinking of things like wanting to quick check something without bringing up a vagrant instance.

Shriukan33:
> If a deploy should run tests, then it would need to be installed on the machine to which we're deploying.
> 
> As I think about it, I'm actually not sure. This is python, once we do the `git pull`, we'd better be sure that the code is good. Unlike a compiled language where you compile and then install to a separate place.
> 
> So I guess we're ok without saltstack support for this in production.

Guess we could figure this out while deploying to beta then. 

> What's curious is that I do have `chromium-browser` installed. And I have a version of chromium installed via snap. So now it somehow needs a _third_ version of chrome.

It's specifically looking for google-chrome, it even inputs `google-chrome --version` to check. it can also look for firefox. 

> Can we the solenium tests auto-disable themselves if the env isn't ok? Is that even a reasonable strategy? I'm thinking of things like wanting to quick check something without bringing up a vagrant instance.

We can keep the selenium tests for dev, as this is where they are used to sanity check that our forms work. 
What happens currently if tests fail on a deploy  ? Are they ran at all ?

JeffAbrahamson:
I don't think we run them on deploy.  It's probably best left to CI anyway.

JeffAbrahamson:
What's the status of this issue?

Shriukan33:
> What's the status of this issue?

Not sure what's next, are you okay with chrome being installed on the vagrant? Otherwise I don't know how to make it run in the vm

JeffAbrahamson:
Yes, I think the only issue was making sure we didn't need to install it on production systems.  But that was sorted when we confirmed that we don't run tests on beta/prod (since the nature of django means the git pull _is_ the deploy).  So leaving it to CI seems the right thing to do.

Perhaps we could set up CI to actually do deploys for us, too, at some point.

Shriukan33:
Probably, don't know if this is doable from Github Action!

Soo this is mergeable? :)

JeffAbrahamson:
Ah, I see, in all the discussion I didn't LGTM.

LGTM.

PR #482, opened 2022-02-16
Create Temporary view to add missing records

Shriukan33:
Some records were missing because of a hole in the Stripe code.
This view will fire once and 500 if refreshed, effectively creating
the missing records only once.

It's stripped of identifying information on purpose.

It's been tested locally
stripe_fix

We need to give the django username and the stripe_customer_id in the URL, so no information shows in this code.

This view will only work once and on a user with only one existing record of Donation, otherwise it will fail purposefully.

### DIFF
--- a/transport_nantes/stripe_app/admin.py
+++ b/transport_nantes/stripe_app/admin.py
@@ -12,6 +12,10 @@ class DonationAdmin(admin.ModelAdmin):
     def has_change_permission(self, request, obj=None):
         """Makes read only all fields of the model."""
         return False
+    readonly_fields = ("timestamp",)
+
+    class Meta:
+        list_display = [field.name for field in Donation._meta.get_fields()]
 
 
 admin.site.register(TrackingProgression, TrackingProgressionAdmin)
--- a/transport_nantes/stripe_app/urls.py
+++ b/transport_nantes/stripe_app/urls.py
@@ -1,5 +1,6 @@
 from django.urls import path
-from .views import (QuickDonationView, StripeView, SuccessView, get_public_key,
+from .views import (QuickDonationView, StripeView, SuccessView,
+                    TempCreateMissingRecords, get_public_key,
                     create_checkout_session, tracking_progression,
                     stripe_webhook)
 
@@ -14,4 +15,6 @@ urlpatterns = [
     path('success/', SuccessView.as_view(), name="stripe_success"),
     path('tracking/', tracking_progression),
     path('webhook/', stripe_webhook, name="webhook"),
+    path('migrate-my-user/<str:stripe_customer_id>/<str:username>/',
+         TempCreateMissingRecords.as_view(), name="create_missing_records"),
 ]
--- a/transport_nantes/stripe_app/views.py
+++ b/transport_nantes/stripe_app/views.py
@@ -3,21 +3,22 @@ import logging
 import random
 import string
 
-from django.views.generic.base import TemplateView
-from django.shortcuts import render
-from django.http import JsonResponse, HttpResponse
-from django.http.response import HttpResponseServerError
-from django.views.decorators.csrf import csrf_exempt
-from django.contrib.auth.models import User
-
 import stripe
 import user_agents
+from django.contrib.auth.models import User
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
+from django.http import HttpResponse, HttpResponseForbidden, JsonResponse
+from django.http.response import HttpResponseServerError
+from django.shortcuts import render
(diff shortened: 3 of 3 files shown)


JeffAbrahamson:
I'd say we don't give away PII by revealing a pk id, but we can make sure that way that the user id passed in is the correct one.  In other words, it's an extra level of avoiding human error.

You don't check that the pk id and username match what was passed in.  Call this on the wrong username and you'll set it to the stripe id of the customer you have in mind.

Perhaps pass only username on the url and hard-code the pk id, then search on both so that you're sure to have the record you think you have.  And check it after fetch: that you got it and that you got only one thing.  You're being too trusting that the world is proceeding as you expect.

Shriukan33:
> 



> I'd say we don't give away PII by revealing a pk id, but we can make sure that way that the user id passed in is the correct one. In other words, it's an extra level of avoiding human error.
> 
> You don't check that the pk id and username match what was passed in. Call this on the wrong username and you'll set it to the stripe id of the customer you have in mind.
> 
> Perhaps pass only username on the url and hard-code the pk id, then search on both so that you're sure to have the record you think you have. And check it after fetch: that you got it and that you got only one thing. You're being too trusting that the world is proceeding as you expect.

Actually I can't know for sure the ID of the customer in the prod database, I'd need to use the shell to confirm it.
I know their email, their django username (random string), and their Stripe customer ID, those I can confirm.

So what do I do ?

JeffAbrahamson:
pk id = 3

Shriukan33:
Used the pkid instead of username, it's fixed now.

Shriukan33:
> Sorry, there was confusion. I was suggesting that you encode `stripe.id=3` in the view, not user id 3. Then you provide the user name at invocation time, and let django confirm that the user name you provided is, indeed, the user associated with donation 3.

Added !

> I think the url path has since lost its name, too, which means that passing any id as `/donation/17/` or whatever will trigger this view.

Understood, did re add the username in url

> In addition, I don't see the view being protected to admins. No matter how secure you think your code is, you should always restrict it if it concerns sensitive data (i.e., if it's not ok that it displays for everyone who visited some reddit for secret URLs on ecology websites).

Added a check at the very beginning of function !


> Finally, you're return value is too trusting. Perhaps you could display the data you've changed or something like that? Your code is saying "trust me" at every turn, and that's rarely good practice.

Expanded on the context value, displaying the result would require to modify the template, while you can check for yourself in the admin panel, so I didn't proceed this way. However a message will pop if it dodges every checks and exceptions on the way !

> The 500 on not finding the object is fine, there's no need to do better error handling here. But I wouldn't advertise it as the thing that guarantees idempotency. Indeed, I think you still need to handle that case, though it's all a bit confused, because you're querying for user 3, not for donation 3, and then when you query donation you do it by user, and this user will have made more than one donation by the time you're done. I think it will work by chance as a side-effect, but I don't think it your code is specifying that this is what should happen.
> 
> Indeed, your only error condition is whether the `get()` or the `save()` throw. But you're catching a generic exception rather than the very specific exception you expect (`<Model>.DoesNotExist` or `<Model>.MultipleObjectsReturned`). Again, you're being way to trusting that everything will work just like you expect it to.

I improved the error handling, sounds better ? 

Modified the function from get_context_data to get() because get_context data is expected to return a dict, not an httpResponse. In Template view, get is defined like this: 

```python
class TemplateView(TemplateResponseMixin, ContextMixin, View):
    """
    Render a template. Pass keyword arguments from the URLconf to the context.
    """
    def get(self, request, *args, **kwargs):
        context = self.get_context_data(**kwargs)
        return self.render_to_response(context)
```

JeffAbrahamson:
The url path needs a name, not a "match any number that happens to come along".  It doesn't matter what the name is, but you're writing a somewhat sensitive function, you should only invoke it on getting _exactly_ the right input.  The more wildcards you have, the easier it is to make a mistake and have something innocent match it.

```
path('<str:stripe_customer_id>/<str:username>/',
```

```
path('migrate-my-user/<str:username>/',
```

We absolutely do care about people trespassing.  Do you want to find out about a problem with the function when someone finds a bug and hits us with 100 requests per second on this function?  Why is it hard to add an authorisation decorator?  It's one extra word and a comma, no?

I have not checked the user id of this person.  I checked the donation id.  You are checking the user id.

```
wanted_user = User.objects.get(id=3)
```

You are looking up the donation by user.  It _happens_ that that will work, but it's fragile.  If the user comes along between the time you write this and the time you execute and makes a donation, your code will break.  Don't write fragile code.

You have the _primary key id_ of the donation record.  _It_ is unique.  Fetch it by hard-coded value.  Check its associated user, which you've passed in on the GET request.  Make sure they match.  Then check that you've not yet done your thing.  You can tell, because you can search for the records you plan to insert.  If all is well, do your thing.  If not, don't and say why not.

PR #496, opened 2022-02-18
Add filter on newspaper name

mcmikashi:
The first and second commit is there for no merging conflic.
On the last commit i add link on newspaper that open on new tab and filter the table.

### DIFF
--- a/transport_nantes/press/templates/press/press_list_view.html
+++ b/transport_nantes/press/templates/press/press_list_view.html
@@ -13,7 +13,7 @@
           </p>
         </div>
         <div class="card-footer text-center">
-          <a href="{{press_mention.article_link}}" class="btn btn-outline-primary" target="_blank"
+          <a href="{{press_mention.article_link}}" class="btn navigation-button" target="_blank"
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
LGTM.

At some point, could you update the button styles to conform to our [style guide](https://www.mobilitains.fr/tb/t/guide-visuel-mobilitains/)?  There's CSS support for it.  If it's unclear or the CSS is not right, just ask.

The style guide page will move, it wasn't supposed to be a TBItem.  Long story.

Shriukan33:
I don't want to put words in Jeff's mouth, but I think he meant these buttons : 
image
image

JeffAbrahamson:
LGTM.

There are some visual quibbles, like the table header not extending over the last column, but they can be addressed later.

More important is that it's not responsive.  The reason that the world abandoned HTML tables in favour of CSS tables is that HTML tables largely break as screen size changes.

JeffAbrahamson:
(Oh, I was looking at `origin/press-mickael-474`...)

PR #499, opened 2022-02-21
Improve the mailing list subcription

mcmikashi:
I use a get method to send the mail and the mailingslist to the last part but I dont know if this is the best way for do this. I think about session for do this too but I dont know what is the best.

### DIFF
--- a/transport_nantes/mailing_list/forms.py
+++ b/transport_nantes/mailing_list/forms.py
@@ -1,6 +1,6 @@
 from django import forms
 from django.contrib.auth.models import User
-from django.forms import ModelForm
+from django.forms import ModelForm, Form
 from captcha.fields import CaptchaField
 from .models import MailingList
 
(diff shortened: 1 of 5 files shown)


JeffAbrahamson:
Feedback:
* careful about trailing white space
* your commit messages are a bit sparse: _how_ did you improve the list subscription?

JeffAbrahamson:
Otherwise, LGTM.  @Shriukan33 ?

Shriukan33:
Was about to do the same comment about .. well comments. You could expand a bit more on what you are doing, we know because it's recent, but this would be hard to get out of context. 
Otherwise LGTM, the get request doesn't seem to be an issue.

JeffAbrahamson:
Hmm, the final stage is a GET with bare URL.  So a refresh re-submits the captcha, even though the user has successfully completed that stage.

I think after succeeding, the user should 302 to a URL that, even on refresh, is the same place and not a part of the subscribe process.

Eventually (soon!) we'll want to provide a custom page per list with fallback default.  But not today.

JeffAbrahamson:
I'm merging because I'm still swimming in github docs explaining all the options and mechanisms for permitting people to merge, for auto-merge, etc.

It's not hard, and it doesn't reflect a volition to limit @mcmikashi , it's just that I'm time-constrained...

Like any good startup, we often start with things that don't scale...

Issue #486, opened 2022-02-17
Make subscribe and unsubscribe functions


JeffAbrahamson:
Subscriptions are currently explicit object creation methods in the view.  For reliability and readability, make functions that do these things in `events.py`.  Call those functions from the view as needed.

Note that these are probably really simple functions.

PR #501, opened 2022-02-21
Improve mailing list subscrition and unsubscription

mcmikashi:
Add subscribe and unsubscribe functions and use the subscribe fonction on views.

### DIFF
--- a/transport_nantes/mailing_list/views.py
+++ b/transport_nantes/mailing_list/views.py
@@ -13,8 +13,9 @@ from asso_tn.views import AssoView
 from .forms import (MailingListSignupForm, QuickMailingListSignupForm,
                     QuickPetitionSignupForm,
                     FirstStepQuickMailingListSignupForm)
-from .models import MailingList, MailingListEvent, Petition
-from .events import user_subscribe_count, subscriber_count
+from .models import MailingList, Petition
+from .events import (user_subscribe_count, subscriber_count,
+                     subscribe_user_to_list)
 
 logger = logging.getLogger("django")
 
@@ -63,11 +64,7 @@ class MailingListSignup(FormView):
         user.profile.code_postal = form.cleaned_data['code_postal']
         user.profile.save()
         for newsletter in form.cleaned_data['newsletters']:
-            subscription = MailingListEvent.objects.create(
-                user=user,
-                mailing_list=newsletter,
-                event_type=MailingListEvent.EventType.SUBSCRIBE)
-            subscription.save()
+            subscribe_user_to_list(user, newsletter)
             # At some point we should also store the last known
             # subscription state in a table with foreign key user.  If
             # the user is in that table, we use it, otherwise we look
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
Same feedback: commit comments are way too vague

JeffAbrahamson:
What does this commit do?  I mean, I could work through the code, but what do you mean for it to do?  Your code is uncommented (should explain the intent of the code, not the detail, which is in the code itself), your commit has no comments (should explain the why -- why you did this at all).  Remember that the issue is handy but is outside git, so you need to bring git readers up to speed.

Shriukan33:
No linked issue in the PR, think about tagging the issues you close in your PRs so it closes the actual issue. :)

JeffAbrahamson:
You can also tag it in the issue ("Closes #486" instead of "Part of #486").  Github has a page about issue-speak recognised language.

The advantage in the PR is that it only closes when the PR merges.

JeffAbrahamson:
It's important that the issue number is in git commit, of course, because `git log --grep`.

mcmikashi:
For the new push I add some comment and rewrite the commit message.

transport_nantes/mailing_list/views.py
```diff
@@ -226,10 +225,8 @@ def form_valid(self, form):
             mailing_list_token=form.cleaned_data['petition_name'])
         if len(petition) == 0:
             return HttpResponseNotFound("Pétition inconnu")
-        subscription = MailingListEvent.objects.create(
-            user=user,
-            mailing_list=petition[0],
-            event_type=MailingListEvent.EventType.SUBSCRIBE)
+        # Create MailingListEvent that make the user sign the petition
```
Shriukan33, line 228: It doesn't exactly make the user sign, the event is created to account for the petition's signing by the user.
So yeah, it's because the user has signed, we create a MailingListEvent. Your comment kind of make us think the reverse happens somehow. 

Same goes for the previous formulations like : 
```python
# Create MailingListEvent that subscribe the user
# in the  mailing_list
```
You save the user's subscription as a mailing list event. 

I'm quibbling here :wink:

JeffAbrahamson: Code comments that say exactly what the code does aren't particularly useful:

```
# Add two and two.
y = 2 + 2
```

The event is called `SUBSCRIBE`, it's kind of obvious what that one line is doing, no?

JeffAbrahamson: Language niggle: distinguish between "make the user sign" and "make sure the user has signed".  They have drastically different meanings. ;-)

mcmikashi:
I deleted the comments.

Shriukan33:
LGTM

JeffAbrahamson:
I think the function names give the wrong impression.  If the intent is to subscribe or unsubscribe the user, then the functions should says so:

`subscribe_user_to_list`
`unsubscribe_user_from_list`

In that case, what is achieved by not calling save immediately?  If the client is requesting a sub or unsub, then do it and return nothing.  It's far clearer semantics unless you will have utility for the returned data.

mcmikashi:
I made change for function (remane,add the save and return none) and update the view.

JeffAbrahamson:
LGTM.

Could you just change the text in the comments before submit:

You don't want to say "Make user un/subscribe to list", this suggests you go round to the user's house with a gun and tell her/him to subscribe.  You want to say "Subscribe user to list" (or unsubscribe).

JeffAbrahamson:
oh, I guess I or @Shriukan33 still need to merge for you.  Sorry, this reflects my time constraints and nothing more.

mcmikashi:
I update the commit message and made the change for the comment.

JeffAbrahamson:
LGTM.  Merging.

@Shriukan33 is going to hit you over the head with a PEP 8 stick for that comment formatting. ;-)

PR #502, opened 2022-02-23
Make email template 419

### LINKED ISSUE #419: Make responsive email template
When viewing an email in browser, we'll use the view `TopicBlogEmailView`.  But when _sending_ a mail, we'll have to more work.

First, and to the point of this issue, we'll need a template that works in an email client.  See #410 for some tips.

Shriukan33:
Add base template for TBEmails

It brings template tags to ease the creation of new templates,
and improves readability of the code.

The views allows us to send and see TBEmails in browser and mail
clients.

:warning: Let's note that for now, it does only check if the user has the permission to send emails, not if it's actually _published_  before proceeding. We haven't defined how to qualify a TBEmail as publishable / published.

For now no interface allows us to have a list of existing TBEmails (beside django admin) to publish them, and there isn't an implementation of publishable like TBItems.
So this is quite minimal, but we are able to 1) create a TBEmail through web form or Django admin 2) Send it to a mailing list, given you have to right to do so.

To send a TBEmail item, use the 'send_email' url, that is currently at this address : 
`/tb/admin/e/send/<pkid>/<slug>/<mailing_list_token>/`

You can preview the TBEmail on `/tb/admin/e/view/<pkid>/<slug>/` as usual.

Last, the image serving is dependent on your own settings.STATIC_URL and MEDIA_URL. the logo uses Static and the images, because they are ImageFields, use media. 
To have a proper display in your emails you can use prod's static and asset servers as urls.
For local use, simply "/static/" and "/media/" would do.  Just keep in mind that if you leave those defaults, the generated email will embed urls like "/static/path/of/image.jpg" which wont work in the context of mail clients. 

Closes #419

### DIFF
--- a/transport_nantes/authentication/views.py
+++ b/transport_nantes/authentication/views.py
@@ -5,7 +5,6 @@ from django.utils.crypto import get_random_string
 from django.views.generic.edit import FormView
 from django.core.mail import send_mail
 from django.contrib import auth
-from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.auth.models import User
 from django.shortcuts import render, redirect
@@ -13,6 +12,7 @@ from django.template.loader import render_to_string
 from django.contrib.auth.views import LogoutView
 from django.conf import settings
 from django.urls import reverse
+from django.utils.html import strip_tags
 from django.views.generic.base import TemplateView
 
 from authentication.forms import (EmailLoginForm, PasswordLoginForm,
@@ -132,7 +132,11 @@ def send_activation(request, email, remember_me):
 
     """
     subject = 'sujet'
-    message = render_to_string(
+    # Emails can have a HTML version and a plain text alternative.
+    # https://docs.djangoproject.com/en/3.2/topics/email/#send-mail
+    # You can pass html to the send_mail function through the
+    # html_message argument.
+    html_message = render_to_string(
         'authentication/account_activation_email.html',
         {  # request.build_absolute_uri(),
             'scheme': request.scheme,
@@ -140,14 +144,15 @@ def send_activation(request, email, remember_me):
             'token': make_timed_token(email, 20),
             'remember_me': remember_me,
         })
-
+    plain_text_message = strip_tags(html_message)
     if hasattr(settings, 'ROLE') and settings.ROLE in ['beta', 'production']:
         try:
             send_mail(
                 subject,
-                message,
+                plain_text_message,
                 settings.DEFAULT_FROM_EMAIL,
                 [email],
+                html_message=html_message,
                 fail_silently=False)
         except Exception as e:
             logger.error(f"Error while sending mail to {email} : {e}")
@@ -155,7 +160,7 @@ def send_activation(request, email, remember_me):
     elif os.getenv('TEST_MODE', "0") == "0":
         # Only print this in dev mode, which is the only time
         # we'd care.
-        print(f"Sent message : \n{message}")
+        print(f"Sent message : \n{plain_text_message}")
 
 
 class ActivationLoginView(TemplateView):
--- a/transport_nantes/mailing_list/events.py
+++ b/transport_nantes/mailing_list/events.py
@@ -2,7 +2,8 @@ from django.contrib.auth.models import User
 from django.db.models import Max
 from django.db.models import Q
 
-from .models import MailingListEvent
+from .models import MailingList, MailingListEvent
+
 
 
 def subscribe_user_to_list(user, mailing_list) -> MailingListEvent:
@@ -82,3 +83,23 @@ def subscriber_count(mailing_list):
     users_subscribed = user_states.filter(
         Q(event_type=MailingListEvent.EventType.SUBSCRIBE)).count()
     return users_subscribed
+
+
+def get_subcribed_users_email_list(mailing_list: MailingList) -> list:
+    """Return a list of email addresses of users subscribed to this list.
+    """
+    # Returns a set of user PK that subscribed to this list at least once.
+    set_of_once_subscribed_users = set(MailingListEvent.objects.filter(
+        mailing_list__mailing_list_token=mailing_list.mailing_list_token
+        ).values_list("user", flat=True))
+    # Get the list of subscribed users.
+    subscribed_users = []
+    for user_id in set_of_once_subscribed_users:
+        user = User.objects.get(pk=user_id)
+        latest_event = user_current_state(
+            user,
+            mailing_list)
+        if latest_event.event_type == "sub":
+            subscribed_users.append(user.email)
+
+    return subscribed_users
--- a/transport_nantes/mailing_list/models.py
+++ b/transport_nantes/mailing_list/models.py
@@ -1,14 +1,12 @@
 from django.db import models
 from django.contrib.auth.models import User
-from django.db.models.signals import post_save
-from django.dispatch import receiver
 import django.utils.timezone
-import uuid
 
 # This application might have been called newsletter.  Think of this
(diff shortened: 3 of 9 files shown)


JeffAbrahamson:
The variable `plain_message` should be `plain_text_message`, because you are (I think) referring to a message with MIME type `text/plain`.

JeffAbrahamson:
The function `get_subcribed_users_email_list` should surely go in `events.py`.  I agree there's an argument that those functions should live on the model, but history.  And there's maybe also an argument that the model shouldn't get more and more loaded as we need additional such functions.

This is a limitation of python: it has no really good way to say what is private data and what depends on API.

JeffAbrahamson:
You're a bit shy on code comments and tests...

JeffAbrahamson:
I think the last commit here should really be more than one commit.  The commit message gives no hint at its complexity and the 435 line diff spans three separate topics: creating an email template, sending mail, and determining the list of people to whom to send it.

Shriukan33:
> The function `get_subcribed_users_email_list` should surely go in `events.py`. I agree there's an argument that those functions should live on the model, but history. And there's maybe also an argument that the model shouldn't get more and more loaded as we need additional such functions.
> 
> This is a limitation of python: it has no really good way to say what is private data and what depends on API.

Could you clarify what's the scope of events.py ? What are we supposed to find in here ? I don't mind moving it to events.py and pass a MailingList to it instead of self, I'm just curious to learn why it shouldn't be in the model ? Placed it in the model because it relies on one instance of mailingList to run, so might as well place it inside, as it's no use anywhere else.

JeffAbrahamson:
> Could you clarify what's the scope of events.py

Functions that manipulate collections of email event objects.

To my mind, and I'm not sure this is a hard rule accepted by everyone, a model method should operate largely on a single row.  When you start doing queries and operating on qeuerysets, it makes more sense to me to be separate.

Shriukan33:
> You're a bit shy on code comments and tests...

True, I didn't write test for this part yet, will do ! Meanwhile I edited the history to fit your comments, tell me if it looks better.

Also you pointed out I lacked on comments, is there a specific part that you are thinking of ? Maybe it was the commit that was way too long ?

JeffAbrahamson:
The main thing that I think needs testing is `get_subcribed_users_email_list`.

The commit with comment "Add base template for TBEmails" should just add the templates.  The function should be in another commit.  A log reader would have no clue that you added the function here, it's a bit far removed from the commit header.

Otherwise, looking good.

Just a heads up, we're going to need eventually (not too far away) filter by users who have already received this email.  Think about what happens when we want to send the same email to two lists.  Or if we say "let's send to the quarterly list, too", but some of those people received mail a month ago, so aren't yet eligible, in a sense.

Shriukan33:
> The main thing that I think needs testing is `get_subcribed_users_email_list`.
> 

Will come with this one first first then ! Also will do a coverage scan see if I can keep the coverage high enough.  

> The commit with comment "Add base template for TBEmails" should just add the templates. The function should be in another commit. A log reader would have no clue that you added the function here, it's a bit far removed from the commit header.

You mean the get_context_data ? I thought it made sense to have it in this commit because the templates do use the email context var, and I thought is was an easy to understand function. 
i could move the get_context_data function right before the template creation? 

> Otherwise, looking good.

Thanks !

Shriukan33:
> Just a heads up, we're going to need eventually (not too far away) filter by users who have already received this email. Think about what happens when we want to send the same email to two lists. Or if we say "let's send to the quarterly list, too", but some of those people received mail a month ago, so aren't yet eligible, in a sense.

I was thinking, we also want to track the sent newsletters, I was thinking we should have a model that stores : 
* The article sent
* The mailing list it's been sent to
* the timestamp of the send
* Eventually but not yet, performances of this campaign (opening rate, bounces ..) 

Design wise I totally see where you are coming from, but I'm not sure yet how to track to who it's been sent from withing the mailing list. Another model with the user as FK, a timestamp, and the mailing list ?
this way we query the list of already-sent-to emails, and remove them from sending list by comparison.

Maybe there is a better approach?

JeffAbrahamson:
>> The commit with comment "Add base template for TBEmails" should just add the templates. 
>
> You mean the get_context_data ?

Looks like I was looking at the wrong version of the branch.  It's good now.

JeffAbrahamson:
> > Just a heads up, we're going to need eventually (not too far away) filter by users who have already received this email. Think about what happens when we want to send the same email to two lists. Or if we say "let's send to the quarterly list, too", but some of those people received mail a month ago, so aren't yet eligible, in a sense.
> 
> I was thinking, we also want to track the sent newsletters, I was thinking we should have a model that stores :
> 
>     * The article sent
> 
>     * The mailing list it's been sent to
> 
>     * the timestamp of the send
> 
>     * Eventually but not yet, performances of this campaign (opening rate, bounces ..)
> 
> 
> Design wise I totally see where you are coming from, but I'm not sure yet how to track to who it's been sent from withing the mailing list. Another model with the user as FK, a timestamp, and the mailing list ? this way we query the list of already-sent-to emails, and remove them from sending list by comparison.
> 
> Maybe there is a better approach?

Are you talking about `TopicBlogEmailSendRecord` and `TopicBlogEmailClicks` (cf. also my pedagical branch `origin/TBEmail-395-views-but-needs-ajax`, which may still contain traces of things not yet done)?

Shriukan33:
> You're a bit shy on code comments and tests...

I added tests regarding View and View One, and the function we talked about. 

I still have the send email view to test, just found this part of the doc :
https://docs.djangoproject.com/en/dev/topics/testing/tools/#email-services
Guess I'll add this next :)

JeffAbrahamson:
Aside: you're now 7 commits from master with 648 lines of diff.  This isn't the idea complexity level for PRs. ;-)

JeffAbrahamson:
It would be really good to test as you go, not all at the end.  So, `e6085e` is missing a test.  It looks pretty straight-forward, but it's still a commit without a test.

JeffAbrahamson:
In `c24871`, your templatetag has no test.

Why does TBEmail get a template subdirectory but TBItem does not?  I'm open to having a subdir for TBEmail (singular), but then we should have a previous commit that puts all the TBItem templates in a subdir.

JeffAbrahamson:
`24b9726ea` has no test (in that commit).

Writing the test separately is bad practice.  First, you might have to change history if your test finds bugs.  Second, it makes history harder to read.

JeffAbrahamson:
In `c8db99a293` you don't explain what's going on.  Yes, you're sending an email, and you're specifying it by pkid and slug, but how does this happen, in what context, what triggers it?  How does anything else about the mail get specified (the subject, the addresses, the From address)?

JeffAbrahamson:
In `c8db99a`, you appear to be sending the mail to all recipients at once.  This poses multiple problems:
* Can recipients now see each other (GDPR violation)?
* Can we detect bounces when you send them together (i.e., who bounced?)
* When you include an unsub link in the mail, who precisely will it unsub?
* Why do all mails come from the same entity?  OK, admittedly, that comes from `9476337b` 18 months ago, but still, did you think about this as what you wanted, something temporary?
* Some of your comments are goofy.  HTML mail is not "the mail to display in the email client", it is the text/html version of the mail.  You are also sending a text/plain version.  It is up to the email client which of those two it shows and what software it uses to render them.  And that is RFC-level stuff, so doesn't need to be commented here, your variables names should be enough to make clear that they are the same content, just different MIME types.
* You aren't logging anything at all, not even on error.  What do you expect to happen when you 500 and talk about pkid or slug?  Who will read that?  And if it happens in production, how will we know anything at all about what happened?

Shriukan33:
Saving the feedbacks, will close this PR to do smaller ones, this has got out of the initial intended scope. :slightly_smiling_face:

PR #504, opened 2022-02-23
Add the user status mailing list page

mcmikashi:
Add status view on mailing list where the user can manage subscription.

### DIFF
--- a/transport_nantes/mailing_list/forms.py
+++ b/transport_nantes/mailing_list/forms.py
@@ -56,7 +56,16 @@ class MailingListSignupForm(ModelForm):
         }
 
 
-# Fist step of subscire to the mailing list
+# Form for subscribe or unsubscribe auth required
+class SubscribeUpdateForm(Form):
+    mailinglist = forms.CharField(
+        max_length=100,
+        required=True,
+        widget=forms.HiddenInput())
+
+# First step of quick signup to the newsletter
+
+
 class FirstStepQuickMailingListSignupForm(Form):
     email = forms.EmailField(label="Adresse mél", required=True)
     mailinglist = forms.CharField(
--- a/transport_nantes/mailing_list/templates/mailing_list/mailing_list_user_status.html
+++ b/transport_nantes/mailing_list/templates/mailing_list/mailing_list_user_status.html
@@ -1,0 +1,12 @@
+{% extends 'asso_tn/base_mobilitain.html' %}
+
+{% block content %}
+<div class="container-fluid p-3">
+    <div class="row">
+        <div class="col-sm-12 col-md-10 mx-auto">
+            {% comment %} Create a list with mailing_list {% endcomment %}
+            {% include "mailing_list/status/list_m.html"  with div_title="Mes newsletters" base_list=mailing_lists%}
+        </div>
+    </div>
+</div>
+{% endblock %}
--- a/transport_nantes/mailing_list/templates/mailing_list/status/hidden_form.html
+++ b/transport_nantes/mailing_list/templates/mailing_list/status/hidden_form.html
@@ -1,0 +1,17 @@
+<form method="post" action="{% url 'mailing_list:toggle_subscription' %}">
+    {% csrf_token %}
+    <input type="hidden" id="mailinglist" name="mailinglist" value="{{mailing_list.id}}">
+    {% comment %} If this is a unsub form show the back button {% endcomment %}
+    {% if unsubscribe %}
+        <div class="row">
+            <div class="col-md-6">
+                    <button type="submit" class="btn navigation-button btn-lg btn-block">{{button_label}}</button>
+            </div>
+            <div class="col-md-6">
+                <a href="{% url 'mailing_list:user_status' %}" class="btn navigation-button btn-lg btn-block">Retour</a>
+            </div>
+        </div>
+    {% else %}
+        <button type="submit" class="btn donation-button m-1">{{button_label}}</button>
+    {% endif %}
+</form>
--- a/transport_nantes/mailing_list/views.py
+++ b/transport_nantes/mailing_list/views.py
@@ -11,11 +11,13 @@ from django.views.generic import ListView, FormView
 
 from asso_tn.views import AssoView
 from .forms import (MailingListSignupForm, QuickMailingListSignupForm,
-                    QuickPetitionSignupForm,
+                    QuickPetitionSignupForm, SubscribeUpdateForm,
                     FirstStepQuickMailingListSignupForm)
 from .models import MailingList, Petition
 from .events import (user_subscribe_count, subscriber_count,
-                     subscribe_user_to_list)
+                     subscribe_user_to_list, user_current_state,
+                     unsubscribe_user_from_list)
+from django.core.paginator import Paginator
 
 logger = logging.getLogger("django")
 
@@ -233,4 +235,67 @@ class MailingListListView(LoginRequiredMixin,
                                     for list in lists if not list.is_petition]
         context['petitions_lists'] = [(list, subscriber_count(list))
                                       for list in lists if list.is_petition]
+        return context
+
+
+class UserStatusView(LoginRequiredMixin, ListView):
+    model = MailingList
+    template_name = "mailing_list/mailing_list_user_status.html"
+    context_object_name = 'mailing_lists'
+    paginate_by = 10
+
+    def get_queryset(self):
+        """Create a list with all active mailing list
+        with the current state of the user"""
+        object_list = MailingList.objects.filter(
+            list_active=True, is_petition=False).order_by(
+            'mailing_list_name')
+        user = self.request.user
+        object_list_with_state = \
+            [(mailing_list, user_current_state(user, mailing_list).event_type,)
+             for mailing_list in object_list]
+        return object_list_with_state
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        """Create a list that contain the current page
+           the current page number with 4 numbers if
+           this is possible before and after the current
+           number."""
+        context["number_pagination_list"] = \
+            context["paginator"].get_elided_page_range(
+            number=context["page_obj"].number,
+            on_each_side=4, on_ends=0)
+        return context
+
+
+class MailingListToggleSubscription(LoginRequiredMixin, FormView):
+    template_name = 'mailing_list/validate_form.html'
+    form_class = SubscribeUpdateForm
+    success_url = reverse_lazy('mailing_list:user_status')
+
+    def get(self, request):
+        # redirect if the get request don't have a mailinglist
+        if not self.request.GET.get("mailinglist"):
+            return redirect(reverse_lazy("mailing_list:user_status"))
+        return super().get(request)
+
+    def form_valid(self, form):
+        mailing_list_id = form.cleaned_data['mailinglist']
+        user = self.request.user
+        mailing_list = get_object_or_404(MailingList, id=mailing_list_id)
+        current_state = user_current_state(user, mailing_list)
+        # check the current state and toggle the result
+        if current_state.event_type == "unsub":
+            subscribe_user_to_list(user, mailing_list)
+        else:
+            unsubscribe_user_from_list(user, mailing_list)
+        return super().form_valid(form)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if self.request.GET.get("mailinglist"):
+            mailing_list_id = self.request.GET.get("mailinglist")
+            context["mailing_list"] = get_object_or_404(
+                MailingList, id=mailing_list_id)
         return context
--- a/transport_nantes/mailing_list/templates/mailing_list/status/list_m.html
+++ b/transport_nantes/mailing_list/templates/mailing_list/status/list_m.html
@@ -1,0 +1,21 @@
+<div>
+    <h2>{{div_title}}</h2>
+    <ul>
+        {% for mailing_list,state in base_list %}
+            <li id="id-ml-{{mailing_list.id}}" >
+                <div class="d-flex align-items-center">
+                    {{ mailing_list.mailing_list_name }}
+                        {% if state == "sub" %}
+                            <a href="{% url 'mailing_list:toggle_subscription' %}?mailinglist={{mailing_list.id}}" class="m-1">Se désabonner</a>
+                        {% else %}
+                            {% comment %} Generate a hidden form for subscribe to a mailing list no confirm needed {% endcomment %}
+                            {% include "mailing_list/status/hidden_form.html"  with button_label="S'abonner" %}
+                        {% endif %}
+                </div>
+            </li>
+        {% empty %}
+            <li>Aucune</li>
+        {% endfor %}
+      </ul>
+      {% include "mailing_list/status/pagination.html" %}
+</div>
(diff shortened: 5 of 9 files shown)


Shriukan33:
## Front end

image

I think the whole table thing is way too big for the amount of information that is present, could even be scary for some users (if it's that big, it's important and complex !)

i think something less wide and simpler would do the trick, maybe with either checkboxes and a submit button, or a link to sub/unsub aside the maling list's name.  The checkbox option would be the more conventional approach, in my opinion. 


(An example of something simpler)
image

I don't think we will have that many newsletter to justify having a whole table of it

Petitions in the other hand can get more numerous, so I'm fine with it being a table !

There are a lot of conjugation mistakes in this, it's no big deal in code but be more careful when you're writing on a user-visible template, as it can damage our credibility. (_Pétition**s** signé**es**_)

mcmikashi:
I add the view with the token for unsubscribe.

JeffAbrahamson:
You're missing issue tags on your commit messages.

JeffAbrahamson:
I don't understand "Create status user view on that view the user". ;-)

JeffAbrahamson:
This is missing test.

mcmikashi:
I rebase on master and add the mail list event historic.

mcmikashi:
I rebase on master and add the mail list event historic.

mcmikashi:
I rebase on master and add the mail list event historic.

JeffAbrahamson:
This PR is a bit complicated to treat.  I'm going to review it commit by commit.  Once one commit is approved, let's move to the next.

So I'm looking at what for me is `e0abba7`, "Add the status user view".  (That might more naturally be "Add user list subscription status page" -- after all, what status are you talking about?

We want to show users the lists to which they are subscribed, period.  We don't want to _encourage_ unsubscription.  Think about what your interface encourages: put a big button there and people will sometimes think they should click it.  So just a list (`<ul>`) of lists is fine.  You might put something in parentheses afterwards indicating that frequency, but you have to say this in a user-friendly way.  Short of thinking that through, don't talk about it.  That's worth an issue in itself, how do we communicate to users about how often they can receive mail on a given list.

Click-through to find out more about the list can be left to another issue.

Using up the whole page with your text is never a good idea.  Look at this:

image

You've used nearly a million pixels on my screen to tell me I'm subscribed to one list and not to another?  Couldn't you just have said
```
<b>Mes newsletter</b>
<p>Je ne suis pas encore abonné à une newsletter.</p>
```
or
```
<b>Mes newsletter</b>
<ul><li>Nouvelles trimestrielles</li>
<li>La vie piétonne</li>
</ul>
```

For petitions, the information isn't actionable.  You can't unsign (short of noting an error in signing, but that's a separate issue, how do we talk about that?).

It's also completely imaginable that we will have tens or hundreds of lists or peittions, and we don't want to present that.  In addition, some lists may not be meant for everyone.  So I don't think that's a useful thing to display to a user.

We will certainly want to think of how to encourage the user to sign up to more newsletters, sign more petitions.  But that's another issue requiring thought on how the interface works and how we communicate it.

I'm mostly saying that these things require thought that hasn't happened yet, neither by me nor (apparently ;-) by you.  So let's make issues for what we'd like to do, but think it through before doing it.

In this particular case, this particular issue, we need tests.  How do you know you're displaying the right thing?

One final issue, remember that the ultimate goal is to have the status page be multi-function.  Look at how other sites to settings.  Until we have enough to say that it warrants multiple pages, we're surely going to display different things together.  But how we do that (templatetags?) can await another discussion and another issue.  For now, a separate page is fine.  But just keep in mind that you're not alone in the world in wanting to display status.

I suggest you branch at this commit and PR it separately.  You can still have your main branch based at this branch (makes it easy to rebase the other as this one evolves).  Once this one is sorted, PR the next piece, etc.  That way you don't slow down your dev, but we can review these things more easily.

mcmikashi:
I update the commit message,update the front of status page and add test.

JeffAbrahamson:
Hey, looking better.  But you're still in "this screen is free, let me use lots of it" mode.

I see this:
image

But I think I should see something closer to this:

image

JeffAbrahamson:
Note that we don't need to show the user the petitions they've signed.  Maybe someday we'll have to do that, or offer to do that based on information transparency rules.  But it's definitely not something the average user is going to be looking for, and it's just likely to confuse.  It's also not actionable.

Again, please think in terms of more data.  Imagine that whatever type of object you're dealing with, there exists tens of thousands of them.  That will help you think about your design.  You certainly aren't going to show the user all 17000 petitions s/he could potentially sign, right?  And do you really want to show the 37538 mailing lists?

Of course, the numbers today are smaller, but design for growth, not for being small.

mcmikashi:
> Again, please think in terms of more data. Imagine that whatever type of object you're dealing with, there exists tens of thousands of them. That will help you think about your design. You certainly aren't going to show the user all 17000 petitions s/he could potentially sign, right? And do you really want to show the 37538 mailing lists?

I think we can separate petition and mailing and use the list view pagination for this or turn the data into json (Jsonresonse, [Django REST framework](https://www.django-rest-framework.org/), [Django Ninja](https://django-ninja.rest-framework.com/) ) and use a js library or framework (Jquery, [ListeJS](https://listjs.com/), [HTMX](https://htmx.org/)) for show the result.

JeffAbrahamson:
I just reread the issue, and I think I understand the source of the confusion.

**Case 1.**  Single user single list unsubscribe.  The context here is that the user received an email at the bottom of which is a link offering to unsubscribe.  On clicking that link, the user finds himself on a page that either says "You aren't subscribed to this list" (say because the user has already clicked the link) or else says "this is the list, this is what we believe your email address is, this is some properties of the list (but skip that for now) and here's a button that lets you unsubscribe if that's really what you want".

Or maybe it says something went wrong and we politely apologise that we can't do anything here.

Note here that we have to _ignore_ the user's current logged in status.  If you get a mail for someone else and click unsubscribe, you unsubscribe _that_ user.  Why?  Because it's really frustrating to receive mail for addresses you no longer control or because you're in charge of someone's mail after they've left a company and you can't unsub them from lists.

**Case 2.** User subscribed list status.  Show the user the lists to which they're subscribed.  Provide unsub buttons (fa rubbish bins?) that take you to the case 1 page.

**Case 3.** Lets us do case 2 as though we were someone else.  So assumes we pass a username.  Don't worry for now about where the username comes from, just implement it assuming it's passed in.

Is that clearer?  I think this is much simpler that the what you've done.

I'd suggest doing each of the three, in that order, as separate PRs.

mcmikashi:
> 2\. User status page.  Show the user all lists to which they are currently subscribed with buttons for unsubscribe.  Unsubscribe takes you to a second page that politely asks for confirmation.  Below, show a second list of newsletters to which the user is not subscribed and offer (much more easily) for the user to subscribe.  This page requires.

This PR is the part  2  of the issue.

JeffAbrahamson:
Does part 1 exist?
If not, where does the unsub take you?

mcmikashi:
Yes it's exist on the previous PR with the token.I need this to be valide for add other it will have a lot of merge conflit.

The unsub take me to page where the user is asked for confirmation and unsub event if the user confirm.

mcmikashi:
Update the design and don't show the pétition that is signed.

transport_nantes/mailing_list/templates/mailing_list/status/hidden_form.html
```diff
@@ -0,0 +1,22 @@
+<form method="post"
+    {% if unsubscribe %}
+        action="{% url 'mailing_list:status_unsubscribe' mailing_list.id %}"
+    {% else %}
+        action="{% url 'mailing_list:status_subscribe' %}"
+    {% endif %}>
+    {% csrf_token %}
+    <input type="hidden" id="user" name="user" value="{{user_id}}">
```
Shriukan33, line 8: Why include the user id in the form while you can retrieve it in the view ? Which you do next btw, there is not plus side, as so far as I'm aware, to add this id to the form ?

transport_nantes/mailing_list/views.py
```diff
@@ -252,9 +254,78 @@ class MailingListListView(ListView):
...
+            if not mailing_list.is_petition:
+                context['mailing_lists'].append((mailing_list, state,))
+            else:
+                if state != "sub":
+                    context['petitions_lists'].append((mailing_list, state,))
+        context['user_id'] = self.request.user.id
+        return context
+
+
+class MailingListSubscribeFromStatus(LoginRequiredMixin, FormView):
```
Shriukan33, line 290: I feel like subscribre and unsuscribe could be in the same view, where depending on the current status, you either invoke `subscribe_user_to_list` or `unsubscribe_user_from_list`.
The two views are extremely similar : They both check, quite unnecessarily (?), the user id of the form against the one of the request, and then get user and ML objects to call the approriate function. 
Why wouldn't you use `user_current_state` with user and ML as args to call one of the two function ? This way you could have a unique "MailingListToggleSubscription" instead of two separate views.

Shriukan33 (changes requested):

mcmikashi:
I made all the change you was asking for @Shriukan33

transport_nantes/mailing_list/templates/mailing_list/mailing_list_user_status.html
```diff
@@ -0,0 +1,14 @@
+{% extends 'asso_tn/base_mobilitain.html' %}
+
+{% block content %}
+<div class="container-fluid p-3">
+    <div class="row">
+        <div class="col-sm-12 col-md-10 mx-auto">
+            {% comment %} Create a list with mailing_list {% endcomment %}
+            {% include "mailing_list/status/list_mp.html"  with div_title="Mes newsletters" base_list=mailing_lists%}
+            {% comment %} Create a list with petition {% endcomment %}
```
Shriukan33, line 9: image

Mes Petition is still present,  we discussed and set on petitions aren't actionable for users, so we don't need to include them. 
beside, the mailing_list/mailing_list_user_status.html is still pushed to the very left, could you give a shot to a centered version please ? :slightly_smiling_face:

transport_nantes/mailing_list/views.py
```diff
@@ -252,9 +254,57 @@ class MailingListListView(ListView):
...
+    template_name = 'mailing_list/validate_form.html'
+    form_class = SubscribeUpdateForm
+    success_url = reverse_lazy('mailing_list:user_status')
+
+    def form_valid(self, form):
+        mailing_list_id = form.cleaned_data['mailinglist']
+        user = self.request.user
+        mailing_list = get_object_or_404(MailingList, id=mailing_list_id)
+        current_state = user_current_state(user, mailing_list)
+        print(current_state)
```
Shriukan33, line 296: Dev print statement to clean here

transport_nantes/mailing_list/views.py
```diff
@@ -252,9 +254,57 @@ class MailingListListView(ListView):
...
+        if current_state.event_type == "unsub":
+            subscribe_user_to_list(user, mailing_list)
+        else:
+            unsubscribe_user_from_list(user, mailing_list)
+        return super().form_valid(form)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        if (self.request.GET.get("mailinglist")):
```
Shriukan33, line 306: Why parenthesis around the if ? Not especially important, just surprised :slightly_smiling_face:

transport_nantes/mailing_list/templates/mailing_list/status/list_mp.html
```diff
@@ -0,0 +1,25 @@
+<div>
+    <h2>{{div_title}}</h2>
+    <ul>
+        {% for mailing_list,state in base_list %}
+            <li id="id-ml-{{mailing_list.id}}" >
+                <div class="d-flex align-items-center">
+                    {{ mailing_list.mailing_list_name }}
+                    {% if mailing_list.is_petition %}
+                            {% include "mailing_list/status/hidden_form.html"  with button="Signer" %}
```
Shriukan33, line 9: The "button" variable should be renamed "button_label" I think, because this is all it is, a label for the button. Same at Line 15

transport_nantes/mailing_list/forms.py
```diff
@@ -55,8 +55,18 @@ class Meta:
...
+# Form for subscribe or subscribe auth required
+
+
+class SubscribeUpdateForm(Form):
+    mailinglist = forms.CharField(
+        max_length=100,
+        required=True,
+        widget=forms.HiddenInput())
+
+# Fisrt step of subscire to the mailing list
```
Shriukan33, line 67: > # Fisrt step of subscire
Some typos here ? :sweat_smile:

transport_nantes/mailing_list/views.py
```diff
@@ -252,9 +254,57 @@ class MailingListListView(ListView):
...
+                self.request.user, mailing_list).event_type
+            if not mailing_list.is_petition:
+                context['mailing_lists'].append((mailing_list, state,))
+            else:
+                if state != "sub":
+                    context['petitions_lists'].append((mailing_list, state,))
+        return context
+
+
+class MailingListToggleSubscription(LoginRequiredMixin, FormView):
```
Shriukan33, line 286: Doing a GET on this URL (ml/toggle_subscription) gets you to a page that looks like this : 
image

Maybe you want to redirect to /ml/status if no mailing list is given. :slightly_smiling_face:

Shriukan33 (changes requested):
Overall, you're being a bit too conservative about Docstrings and comments in your code, try to explain more the intent of your code, it's not very hard to read, but it still is longer to understand and review when you don't have the easy path to read the comment :)

mcmikashi:
I rebase and made the update you asking for.

JeffAbrahamson:
This PR, when rebased to master and diff'd, appears to make changes to the signup process.  I don't believe that is intended.  Could you please rebase to origin/master, review what you've done, and make sure it does what you expect?

This is a relatively small feature, I'm suspicious when I see a nearly 700 line diff.

mcmikashi:
I rebase and change :
- On template change the varaible name "button" is now "button_label"
- Add comment on view and template
- Add redirection if the get request on (ml/toggle_subscription) don't have mailling redirect to  (/ml/status)
- Only show petition that hasn't been signed

JeffAbrahamson:
The documentation on this commit is lacking.  The commit comment isn't telling me clearly what to expect and is garbled.  The issue (487) is about mailing lists, but both the commit comment and the code continues to treat petitions.

The page at `http://localhost:8000/ml/status` is pretty user-unfriendly.  Why are you referring to me in the third person, for example?  Can you find some examples on the internet of what such pages look like?

Also, you show a list of every list we have.  Imagine that there are thousands.  Are you going to show me that?

Assuming you have access to a bash-like shell, could you please run the following:

```
for n in {2..100}; do echo "INSERT INTO mailing_list_mailinglist (mailing_list_name, mailing_list_token, contact_frequency_weeks, list_active, is_petition, list_created) VALUES ('list_$n', 'list-$n', $n, 1, 0, '2022-03-02');" | sqlite3 db.sqlite3; done
```

or else write yourself a short python script in the django python shell to do the same.

JeffAbrahamson:
image

mcmikashi:
I rebase and make some change :

- Delete petiton from the status view ( and update the test)
- Add pagination to status view

JeffAbrahamson:
I'm going to LGTM and merge this, but it's still fundamentally wrong on the user side, just relatively harmless.  We can fix it later.  The code is fine.

What's wrong:
* It shows an alphabetical list of all mailing lists ever, whether subscribed or not.  This is rarely what a user wants and it's something we explicitly didn't want to do here.
* The user experience is downright weird, from the strange kerning between text and buttons to the vertical wtf by having the buttons not line up.  (This is why, if you copy other people, you discover that people often put buttons on the left, since they are fixed width.)

Issue #497, opened 2022-02-18
tn_markdown can thow


JeffAbrahamson:
If someone types an invalid link, the view after TBItem edit will 500.

PR #507, opened 2022-02-27
Fix order issue on slug: markdown directive

JeffAbrahamson:
I'd inversed the label and slug.

Part of #497.

### DIFF
--- a/transport_nantes/topicblog/test_tn_links.py
+++ b/transport_nantes/topicblog/test_tn_links.py
@@ -100,8 +100,8 @@ class TnLinkParserTest(TestCase):
     def test_internal_url(self):
         the_slug = 'my-slug'
         the_label = 'my-label-text'
-        markdown = f'[[slug:{the_slug}]](({the_label}))'
-        html = slug.tbi_slug(the_slug, the_label)
+        markdown = f'[[slug:{the_label}]](({the_slug}))'
+        html = slug.tbi_slug(the_label, the_slug)
         self.assertEqual(self.parser.transform(markdown), html)
 
     def test_petition_url(self):
--- a/transport_nantes/topicblog/templatetags/slug.py
+++ b/transport_nantes/topicblog/templatetags/slug.py
@@ -6,7 +6,7 @@ register = template.Library()
 
 
 @register.simple_tag
-def tbi_slug(slug, label):
+def tbi_slug(label, slug):
     """Render an internal TBItem link from a slug.
 
     The value must include everything after "/tb/t/".


transport_nantes/topicblog/test_tn_links.py
```diff
@@ -100,8 +100,8 @@ def test_external_url_button(self):
     def test_internal_url(self):
         the_slug = 'my-slug'
         the_label = 'my-label-text'
-        markdown = f'[[slug:{the_slug}]](({the_label}))'
-        html = slug.tbi_slug(the_slug, the_label)
+        markdown = f'[[slug:{the_label}]](({the_slug}))'
```
Shriukan33, line 103: Then you put the label next to the slug, it's counterintuitive. 
I do agree that it makes more sense to have the [[label]]((slug)) because it fits the usual markdown pattern, but I think you should also update the parser to not use "slug" arg for label 
It's line 229 in tn_links.py

JeffAbrahamson: The pattern (and the "don" directives are an unfortunate exception) is `[[verb:user-text]]((direct-object))`, for some very permissive definition of "verb" and "direct-object".

Are you suggesting a different user interface than `[[slug:text]]((item_slug))`?  What do you think the right thing to do is?

JeffAbrahamson: BTW, I'm perfectly fine with the idea that we should migrate our markdown extensions to a more logical schema.  It's not yet hard to do.

If we want to go that way, we should make a design doc.  But we can start that conversation here.

Shriukan33: Was thinking about the code, mostly, because reading it and seeing "slug:" I'd expect to have the slug argument right next to it, and it's now the label. 
but I see what you mean in "verb" now that you say it, maybe it's just the "slug" verb that is just not very clear from a user (user as in someone who writes articles for mobilitains) standpoint. 
`[[bouton_vers_article: {the_label}]](({the_slug}))`
would make more sense, maybe ? 
Though it wasn't the objective of this PR initially :grimacing:

JeffAbrahamson: As I said, I'm open to a global rethink and design doc. ;-)

What do you think on this PR in particular?

JeffAbrahamson: No fixed idea on the design doc, just that if we think the vocabulary or syntax needs modifying, better now than a year from now.

So do you lgtm?

Shriukan33: PR: LGTM 

Agreed with the need of design doc ! Will you open the issue  ?

JeffAbrahamson: Done.

PR #510, opened 2022-02-28
Add html_message argument to send_mail function

Shriukan33:
This allows us to send both plain text and html versions of a message
in an email.

For instance, the magic token email. 

The template doesn't contain actual html tags in it (yet), the objective is to be consistent with the rest of the code regarding emails.

Part of #419


JeffAbrahamson:
LGTM.

Should the print be a log.info for good form?  I don't have a strong opinion, do as you like on that point and rebase+merge.

Shriukan33:
No point in making it a log, as it's only in test mode, aka dev. :slightly_smiling_face:

PR #511, opened 2022-02-28
Add email template newsletter 419

### LINKED ISSUE #419: Make responsive email template

Shriukan33:
Add base template for TBEmails alongside a view to preview them.
This template displays properly on common mail clients (Tested on Thunderbird, gmail, and outlook).

Closes #419

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/TBEmail/base_email.html
+++ b/transport_nantes/topicblog/templates/topicblog/TBEmail/base_email.html
@@ -1,0 +1,130 @@
+<!DOCTYPE html>
+<html lang="en" xmlns="https://www.w3.org/1999/xhtml" xmlns:o="urn:schemas-microsoft-com:office:office">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <meta name="x-apple-disable-message-reformatting">
+    <title></title>
+    {% comment %}
+        Some CSS and PNG sizing resets for Microsoft Outlook on Windows
+        hidden inside conditional comments (the <!--[if mso]> bits)
+        src : https://webdesign.tutsplus.com/articles/creating-a-simple-responsive-html-email--webdesign-12978
+    {% endcomment %}
+    <!--[if mso]>
+    <style>
+        table {border-collapse:collapse;border-spacing:0;border:none;margin:0;}
+        div, td {padding:0;}
+        div {margin:0 !important;}
+    </style>
+    <noscript>
+        <xml>
+            <o:OfficeDocumentSettings>
+            <o:PixelsPerInch>96</o:PixelsPerInch>
+            </o:OfficeDocumentSettings>
+        </xml>
+    </noscript>
+    <![endif]-->
+    <style>
+        table, td, div, h1, p {
+            font-family: Arial, sans-serif;
+        }
+        @media screen and (max-width: 530px) {
+            .unsub {
+                display: block;
+                padding: 8px;
+                margin-top: 14px;
+                border-radius: 6px;
+                background-color: #555555;
+                text-decoration: none !important;
+                font-weight: bold;
+            }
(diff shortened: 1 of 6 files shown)


JeffAbrahamson:
Why do you implement special email template tags without any comment in the code about why you are doing this?  What is the problem you are solving?  According to your commit message, you are improving readability, but that doesn't quite explain why these different template tags are needed.

What is the intent of `email_cta_button` that is different from `action_button`?

What will cause markup used in an email message to use these functions instead of the ones in `asso_tn/templatetags/don.py`?

Shriukan33:
> Why do you implement special email template tags without any comment in the code about why you are doing this? What is the problem you are solving? According to your commit message, you are improving readability, but that doesn't quite explain why these different template tags are needed.
> 

The email template relies heavily on HTML tables (<td>, <tr> ...) which makes it really heavy to use and hard to read at glance. 
If, in the future, we want to alter the email template or simply create another one, I figured it would be easier to just have blocks you can pile and have a nice looking page without the pain of using the HTML table. Those blocks are the template tags ! You can reorder them to your liking, making it easier to edit and create new templates, in spite of its bad readability under the hood. 

> What is the intent of `email_cta_button` that is different from `action_button`?
> 

The email_cta_button will use HTML tables that are compatible with the email templates, because they are built differently compared to our normal templates, for example they don't bring bootstrap classes.

> What will cause markup used in an email message to use these functions instead of the ones in `asso_tn/templatetags/don.py`?

Not sure to understand this question properly, but basically the reason I created these templates is because Email clients, as they act differently and are limited in features, use an HTML structure that is harder to read than our pages, so having templatetags as building blocks makes it easier to edit. 

Attempting to use our existing template tags will result in html code that wont fit in the structure because: 
* Email clients wont use bootstrap classes or run any JS
* Use the wrong html structure to fit the rest of the document

Shriukan33:
(Rebase because of a miss in my commit commands, no changes otherwise)

JeffAbrahamson:
>> What will cause markup used in an email message to use these functions instead of the ones in asso_tn/templatetags/don.py?
>
> Not sure to understand this question properly, but basically the reason I created these templates is because Email clients, as they act differently and are limited in features, use an HTML structure that is harder to read than our pages, so having templatetags as building blocks makes it easier to edit.
>
> Attempting to use our existing template tags will result in html code that wont fit in the structure because:
>
> * Email clients wont use bootstrap classes or run any JS
> * Use the wrong html structure to fit the rest of the document

Hmm, so the fundamental assumption is that the same markdown can generate both the email we sent to someone and the email we display in a web page.  So this may require some rethink.

How were you imagining users would format emails if not via markdown?

JeffAbrahamson:
Can we use a library like [bootstrapemail](https://bootstrapemail.com/docs/introduction) to do this?

N.B.: I haven't analysed it, just googled "responsive email bootstrap" to read what people way about getting responsive emails and the fact that you can't count on bootstrap loading.

Shriukan33:
>Can we use a library like [bootstrapemail](https://bootstrapemail.com/docs/introduction) to do this?
>
>N.B.: I haven't analysed it, just googled "responsive email bootstrap" to read what people way about getting responsive emails and the fact that you can't count on bootstrap loading.

So I've looked into the link you mention, it's unfortunately compatible only with Ruby on rails.
I've looked for alternatives, and consensus is "Email template is a bitch, you need to write tables and inline CSS to make it look good, en even then, you can't be sure."
For now I could make a basic template that seem to work, and isn't too hard to edit if we keep with simple stuff (text, images, and buttons).
I tried to make it simpler with template tags.

Shriukan33:
Will make another PR based on master and which integrates all newer changes

PR #513, opened 2022-02-28
Fix the quick mailing list on production

mcmikashi:
Redirect all user on production who use get method on the QuickMailingListSignup and MailingListMerci view to the FirstStepQuickMailingListSignup view.

### DIFF
--- a/transport_nantes/mailing_list/templates/mailing_list/first_step_sign_up_m.html
+++ b/transport_nantes/mailing_list/templates/mailing_list/first_step_sign_up_m.html
@@ -1,0 +1,32 @@
+{% extends 'asso_tn/base_mobilitain.html' %}
+{% load static %}
+{% load don %}
+{% load mpanels %}
+{% load newsletter %}
+{% load crispy_forms_tags %}
+
+{% block og_title %}Inscrivez-vous à la newsletter des Mobilitains !{% endblock og_title %}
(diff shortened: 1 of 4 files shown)


JeffAbrahamson:
I think the logic you want is really, really simple.

On `QuickMailingListSignup`, you have a `get()` function.  If it's called, that means someone has arrived without already trying to signup to a list, say a URL share.  Expect an argument like `?list=xyz` and ask for email and captcha to sign up to xyz.  (This means that things that send us here probably want to provide those args.  Though I have a doubt as to whether they play nicely with get/post transitions.)

Or maybe you arrive on `post()`, and then you know that a subscription is in process.  So check to see if you have an email and a successful captcha result in the form.  If you do, then you're done and you render the thank-you page.  If you don't, then you're not done and your render the captcha page, optionally with an error message if you have a failed captcha.

In all cases, you stay at the say URL -- the only thing that changes is the GET or POST data.  Indeed, if the GET case causes problems, we can probably drop it and merely display a message saying "sorry, we're not sure what you were looking for".

Am I confused?

mcmikashi:
The first form` FirstStepQuickMailingListSignup` don't have capcha if i send a post to request direcly to the second form `QuickMailingListSignup` it will show error capcha is required this is why i use get method to send the mailing list and the mail to  the second part and the other things work like you say.

PR #517, opened 2022-03-02
Add the mailing list   unsubscribe with token page

mcmikashi:
Add the first part of the 487 issue : 
User unsubscribe page for one list. (Note: not for petitions.) If the user is not subscribed (note that we already have a function to check), just say so. Otherwise, offer to unsubscribe. Present a button that, on click, writes an unsubscribe event and confirms for the user that this is done. Unsubscribing more than once is not an error (if, somehow, the user prepares a bunch of pages to unsubscribe).

### DIFF
--- a/transport_nantes/mailing_list/templates/mailing_list/last_part_unsubscribe.html
+++ b/transport_nantes/mailing_list/templates/mailing_list/last_part_unsubscribe.html
@@ -1,0 +1,16 @@
+{% extends 'asso_tn/base_mobilitain.html' %}
+{% load mpanels %}
+{% block app_content %}
+<div class="container pt-5">
+    <div class="row">
+        <div class="col-md-12">
+            <p>Vous avez bien été désinscrit de la newsletter.</p>
+        </div>
+    </div>
+    <div class="row pt-4">
+	{% show_projects %}
+    </div>
+</div>
+{% show_donate %}
+{% show_volunteer %}
+{% endblock %}
--- a/transport_nantes/mailing_list/tests.py
+++ b/transport_nantes/mailing_list/tests.py
@@ -2,15 +2,16 @@ from django.test import Client, LiveServerTestCase, TestCase
 from selenium.webdriver.chrome.webdriver import WebDriver
 from webdriver_manager.chrome import ChromeDriverManager
 from selenium.webdriver.chrome.options import Options
-from topicblog.models import TopicBlogItem
+from topicblog.models import TopicBlogItem, TopicBlogEmailSendRecord
 from datetime import datetime, timezone
 from .models import MailingList, MailingListEvent, Petition
 from django.contrib.auth.models import User
 from django.urls import reverse
 from .events import user_current_state
+from asso_tn.utils import make_timed_token
 
 
-class MailingListIntegrationTestCase(LiveServerTestCase):
+class MailingListStatusCodeTest(TestCase):
     def setUp(self):
         # Create MailingList (newsletter) object
         self.mailing_list_1 = MailingList.objects.create(
@@ -48,15 +49,16 @@ class MailingListIntegrationTestCase(LiveServerTestCase):
             slug="slug2",
             petition1_md="First sentence 2",
         )
+
         self.user = User.objects.create_user(username='ml-staff',
                                                       password='ml-staff',
-                                                      email="staff@mail.com")
+                                                      email="duponun@test.fr")
         self.user.save()
         self.logged_client = Client()
         self.logged_client.login(username='ml-staff', password='ml-staff')
         self.superuser = User.objects.create_superuser(username='ml-admin',
                                                        password='ml-admin',
-                                                       email="admin@mail.com")
+                                                       email="admin@test.fr")
         self.superuser.save()
         self.admin_client = Client()
         self.admin_client.login(username='ml-admin', password='ml-admin')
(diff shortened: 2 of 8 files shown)


JeffAbrahamson:
This branch does not currently rebase to master, could you please update so it becomes reviewable?

Thanks!

JeffAbrahamson:
This doesn't rebase cleanly to master, could you clean up?  Thanks!

mcmikashi:
I rebase on master.

JeffAbrahamson:
The file `sorry.html` appears to be a generic 404 page, why does it need to exist?  If it does need to exist, why isn't it a separate issue and a well-thought-through design using django's `404.html` mechanism?

JeffAbrahamson:
Given that `make_timed_token` is in `asso_tn/utils.py`, perhaps `make_not_timed_token` should go there, too?  (More natural would be `make_untimed_token`.)  They share a lot of code, could they be the same function with a flag?  Or might we simply establish that no token is forever and provide a long timeout, like a week?

mcmikashi:
I make some change :

- I delete the utils on mailing_list 
- On asso tn add the flag on make timed token and token valid for accepting the required input for the mailing list token 
- add test in asso tn and mailing list
- Delete the sorry page and redirect to the default 404.

JeffAbrahamson:
You test with an email address from `nomail.fr`.  This could lead to unstable results depending on future code and how that domain resolves under DNS.  Please use established reserved top level DNS names, cf. [RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606).

JeffAbrahamson:
You make a rather significant change to `make_timed_token`, but the function comment strangely stays the same, as if the function does pretty much what it did before.  In addition, you don't elaborate on it in commit comments.  And its tests barely change.  If this function evolves so much, it seems worthy of explanation.  Are you fixing a bug?  Adding features?

JeffAbrahamson:
The file `sorry.html` still exists despite your saying you 404 instead.  I agree you appear not to make reference to it.

mcmikashi:
- I rebase on master 
- Delete the sorry.html
- Update the commit message and add comment on make_timed_token and valid_token
- Update the commit message

mcmikashi:
- I add  make_mailing_list_token on asso_tn util
- update test on asso tn
- update the commit message
- Add the update of TopicBlogEmailSendRecord unsubscribe_time

mcmikashi:
I think TopicBlogEmailSendRecord should allow open_time,click_time and unsubscribe_time to be blank or none.

Shriukan33:
> I think TopicBlogEmailSendRecord should allow open_time,click_time and unsubscribe_time to be blank or none.

Added this in my latest branch, will make a PR soon, still need to write tests

transport_nantes/mailing_list/views.py
```diff
@@ -299,3 +300,61 @@ def get_context_data(self, **kwargs):
...
+
+class MailingListUnsubscribeWithToken(FormView):
+    form_class = SubscribeUpdateForm
+    success_url = reverse_lazy('mailing_list:unsubscribe_finish')
+    template_name = 'mailing_list/validate_form.html'
+
+    def get(self, *args, **kwargs):
+        token = self.kwargs['token']
+        if not token:
+            logger.info("Try to unbscrucribe to mailist without token.")
```
Shriukan33, line 313: unbscrucribe x3

transport_nantes/mailing_list/views.py
```diff
@@ -299,3 +300,61 @@ def get_context_data(self, **kwargs):
...
+        context['user'] = get_object_or_404(User, email=email)
+        context['tb_email_send_record'] = get_object_or_404(
+            TopicBlogEmailSendRecord, id=the_tb_email_send_record_id)
+        context['mailing_list'] = context['tb_email_send_record'].mailinglist
+        # Check the current state and set context if the user is unsub
+        current_state = user_current_state(
+            context['user'], context['mailing_list'])
+        if current_state.event_type == "unsub":
+            context['is_unsub'] = True
+        context['is_from_token'] = True
```
Shriukan33, line 344: What's the point of `context['is_from_token'] = True` ? If this is always True, why does it exist anyway ?

transport_nantes/mailing_list/views.py
```diff
@@ -299,3 +300,61 @@ def get_context_data(self, **kwargs):
...
+        return context
+
+    def form_valid(self, form):
+        context = self.get_context_data()
+        mailing_list_id = form.cleaned_data['mailinglist']
+        user = context['user']
+        mailing_list = get_object_or_404(MailingList, id=mailing_list_id)
+        unsubscribe_user_from_list(user, mailing_list)
+        # save the new TopicBlogEmailSendRecord unsubscribe date
+        context['tb_email_send_record'].unsubscribe_time = datetime.now()
```
Shriukan33, line 354: Naive datetime (timezone.utc)

transport_nantes/mailing_list/templates/mailing_list/last_part_unsubscribe.html
```diff
@@ -0,0 +1,22 @@
...
+{% load static %}
+{% load don %}
+{% load mpanels %}
+{% load crispy_forms_tags %}
+{% block app_content %}
+
+<div class="container pt-5">
+    <div class="row">
+        <div class="col-md-12">
+            <p>Vous avez bien été désinscrie de la newsletter.</p>
```
Shriukan33, line 11: https://la-conjugaison.nouvelobs.com/du/verbe/desinscrire.php

transport_nantes/mailing_list/templates/mailing_list/last_part_unsubscribe.html
```diff
@@ -0,0 +1,22 @@
+{% extends 'asso_tn/base_mobilitain.html' %}
+{% load static %}
+{% load don %}
+{% load mpanels %}
+{% load crispy_forms_tags %}
```
Shriukan33, line 5: Do you need to load all those ? I get it for mpanels, but the rest ?

transport_nantes/mailing_list/tests.py
```diff
@@ -370,3 +255,187 @@ def test_status_page_context(self):
...
+                                    reverse("mailing_list:unsubscribe_token",
+                                            kwargs={
+                                                "token": token
+                                            })))
+        title_text = self.selenium.find_element_by_css_selector(
+            ".container-fluid h5.alert").get_attribute(
+            'innerHTML')
+        self.assertEqual(title_text,
+                         "Vous êtes déja désabonné à cette newsletter",
+                         msg="User can't unsuscibe again")
```
Shriukan33, line 441: unsuscibe

Shriukan33 (changes requested):

transport_nantes/mailing_list/views.py
mcmikashi: I use same template for status and token i need to know where the request is from for the render.

Shriukan33: Could you please add a comment to explain this please ? It's not immediately obvious :slightly_smiling_face:

mcmikashi:
- fix the gramar and missspelling
- roll back the minute comment
- remove the unused load on last_part_unsubscribe.html template
- use the date time datetime (timezone.utc) for all date.
- add comment for ['is_from_token']

PR #518, opened 2022-03-02
Add email send view 402

### LINKED ISSUE #402: Make TBEmail sendable to a mailing list
The send function, visible to people authorized to send mails, should :
* show the mailing list's available and permit making a choice. 
* show the number of people to whom the mail will be sent and ask for confirmation. 
* Before any of that, it should verify that the email is published. 

Note that an email could be sent more than once (but probably not to the same list).

Shriukan33:
This PR is based on the PR #511 

Adds a view that, if the user has the permission, sends to a mailing list the given TBEmail.

For now the view renders the Email (because it's easier to see the result straight away), but we could just return a 200 code and be done with it. 
I guess we will think of other ways to send mails (through a dashboard for example), but as we didn't this part yet, the view will work in the meantime. 

Note : For test purpose, only @mobilitain.fr addresses are valid as recipients.

Closes #402

### DIFF
--- a/transport_nantes/mailing_list/events.py
+++ b/transport_nantes/mailing_list/events.py
@@ -2,7 +2,8 @@ from django.contrib.auth.models import User
 from django.db.models import Max
 from django.db.models import Q
 
-from .models import MailingListEvent
+from .models import MailingList, MailingListEvent
+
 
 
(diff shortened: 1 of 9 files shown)


Shriukan33:
Rebased !

Shriukan33:
This branch will be implemented in different branches

PR #519, opened 2022-03-02
Update the originating view and params on sub cycles.

Shriukan33:
Because the origin isn't a form we created but a renewal event sent from stripe, we might just save it as such.


JeffAbrahamson:
LGTM.

JeffAbrahamson:
I'm going to merge and close this, because it looks straight-forward and harmless, and I'd rather it not take our time by passing into merge conflict in a couple days.

PR #529, opened 2022-03-06
Jeff tb stuff

JeffAbrahamson:
In an attempt to accelerate development, I've implemented TBEmail, TBPress, and TBLauncher this afternoon.

A handful of points overlap with things on which you guys have been working for a while.  I tried to pull in your code when I could.

I'm presenting this as a PR to make conversations easier (maybe), but I don't want you to think in terms of approving a standard PR.  Among other things, it's way too long (16 commits) to do that efficiently.

What I'd like you to is to take this commit by commit, in order, and ask yourselves if at _that commit_ you see any blocking issues.  As you LGTM the commits, I'll merge them onto master.  To the extent that you find need for improvement or bugs, let's improve or fix them.  But be really careful to distinguish between bugs vs "not everything is perfect yet".

@Shriukan33 You can probably make a good start in the  morning.  Once @mcmikashi comes online mid-day for you, I'd like you guys to pair program the rest of it.  I'd expect that you should be able to work through the 15 commits by EOD and that the remaining functionality should be doable tomorrow (Tuesday).  Once it's working, we'll see things that need cleanup and improvement, and we can do those things.

A couple notes:
* I've left the tests for you guys to sort out, although I've run it through its paces and I think it's in good shape.  Be careful about testing in too much detail right away, as interface and workflow may change slightly.
* I didn't implement edit for any of these.  To start with, just make scratch entries via `/admin/`.  Again, once we get all of this in approved and merged, then go back and implement edit.
* I've left detailed notes for you on how I think you should implement the send routines for TBEmail and TBPress.  Pleaes work though and let's get merged what's here, then go back and fill in the send functions.
* Once all of that is done, let's think about aesthetics.  There's a huge different between otherworldly user experience  (a point on which I sometimes harp) vs just getting spacing, colour, text a bit wrong.  The perfect is the enemy of the good.


transport_nantes/topicblog/views.py
```diff
@@ -376,9 +390,16 @@ class TopicBlogEmailView(TopicBlogBaseView):
...
 
 
-class TopicBlogEmailViewOne(TopicBlogBaseViewOne):
+class TopicBlogEmailViewOne(TopicBlogEmailViewOnePermissions,
+                            TopicBlogBaseViewOne):
     model = TopicBlogEmail
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['context_appropriate_base_template'] = 'asso_tn/base_mobilitain.html'
```
Shriukan33: Why displaying emails with base_mobilitain ? It's quite different from the original template and the template the user is supposed to see in his client. I feel like viewing an email object should be seen as if if was in a mail client. Unless you have some motives I haven't seen yet ?

Shriukan33 (commented):
On commit e76be538ed1b9c3fa705c93dbd7e8ca322185c18, other than that, LGTM

Shriukan33 (commented):
Commit 6e71bb08d615ebef4096bf5811738c2a32602826 LGTM

transport_nantes/topicblog/views.py
JeffAbrahamson: The view is in the context that the user has clicked the "display in my browser", so they are now displaying as part of our website.  The should have a standard web experience.

The other rendering is when we send the content as email, then we use `base_email_mobilitain` to render for an email client.

transport_nantes/topicblog/views.py
```diff
@@ -389,6 +389,11 @@ class TopicBlogEmailEdit(TopicBlogBaseEdit):
 class TopicBlogEmailView(TopicBlogBaseView):
     model = TopicBlogEmail
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['context_appropriate_base_template'] = 'asso_tn/base_mobilitain.html'
```
Shriukan33: Same comment as previous commit : 

> Why displaying emails with base_mobilitain ? It's quite different from the original template and the template the user is supposed to see in his client. I feel like viewing an email object should be seen as if if was in a mail client. Unless you have some motives I haven't seen yet ?

Shriukan33 (commented):
Commit 8d566170c4fc654732d190991f1ac877c0477b3a LGTM % the question

transport_nantes/topicblog/views.py
Shriukan33: Ha, didn't  see base-email_mobilitain yet, this explains it ! I get it, SGTM

JeffAbrahamson: That said, we can fiddle with it later to get it how we want.  Maybe we'll want reduced headers/menus for displaying emails.  I was envisioning something akin to how we display press releases or blog posts.

Shriukan33 (commented):
LGTM 6064ff1267acb42dc840f65d21b3312837bf5693

transport_nantes/topicblog/templates/topicblog/base_email.html
```diff
@@ -0,0 +1,67 @@
...
+			{% endif %}
+	    {% endif %}
+
+	    <a type="button" class="btn btn-outline-info btn-sm"
+	       href="{{ page.get_edit_url }}">Modifier</a>
+
+	    {% if page.publication_date %}
+	    <a type="button" class="btn btn-outline-info btn-sm"
+	       href="{% url 'topic_blog:view_email_by_slug' page.slug|default:'--sans-slug--' %}">Visualiser (usager)</a>
+	    {% if page.servable %}
```
Shriukan33, line 47: Servable is some dirt we forgot to clean from the early versions of TBItem, this will never show up.

transport_nantes/topicblog/templates/topicblog/base_email.html
```diff
@@ -0,0 +1,67 @@
...
+	    {% endif %}{% endif %}
+	</p>
+    </div>
+</div>
+  {% elif sendable %}{% comment %}
+Sendable and not topicblog_admin, so viewing published page, so can
+offer to send.
+{% endcomment %}<div class="container-fluid pl-5 py-3">
+    <div class="row">
+	<a type="button" class="btn btn-warning" area-disabled="true" href="{% url 'topic_blog:send_email' page.slug %}">Envoyer</a>
```
Shriukan33, line 62: This button has the interesting particularity to be eyes burning, we will definitely want to fix that, I think we will fix the overall css once the logic is set :)

Shriukan33 (commented):
Commit 442f8f51a590666290f51ebe594f76b5460a4e18 has some cleaning to do, but logic wise it LGTM

Shriukan33 (commented):
I see someone had to halt vagrant to get to another 8000 ! :laughing: 
LGTM, no problems with that

transport_nantes/topicblog/templates/topicblog/base_email.html
JeffAbrahamson: Yes, I agree whole-heartedly!
I couldn't see right off what colour to make the button, so I decided to make sure we'd revisit it!

transport_nantes/topicblog/templates/topicblog/base_email.html
JeffAbrahamson: Yeah, I think I even made an issue to come back to it.  I didn't want to refactor that bit at the same time.

There's also the references to "retired" that need to die at the same time.

JeffAbrahamson:
> I see someone had to halt vagrant to get to another 8000 !

So many ports, so little time!

Shriukan33 (commented):
Commit ef9a4ee801013fd53df67c3bea418e3f48a73b40
TopicBlogEmailView : LGTM
The EmailSend : Thanks for the specs, will definitely help us writing the code.

transport_nantes/topicblog/models.py
```diff
@@ -664,3 +664,121 @@ class TopicBlogEmailClicks(models.Model):
...
+    """Represent the fact that we sent an press.
+    """
+    slug = models.SlugField(max_length=90, allow_unicode=True, blank=True)
+    mailinglist = models.ForeignKey(MailingList, on_delete=models.PROTECT)
+    recipient = models.ForeignKey(User, on_delete=models.PROTECT)
+    send_time = models.DateTimeField(auto_now=True)
+    # Open time is the time of the first instance of a beacon responding.
+    open_time = models.DateTimeField()
+    # Click time is the time of the first instance of a link being clicked.
+    click_time = models.DateTimeField()
```
Shriukan33, line 783: When you say "of a link" you mean any link contained into the article ?

transport_nantes/topicblog/models.py
```diff
@@ -664,3 +664,121 @@ class TopicBlogEmailClicks(models.Model):
...
+class TopicBlogPressSendRecord(models.Model):
+
+    """Represent the fact that we sent an press.
+    """
+    slug = models.SlugField(max_length=90, allow_unicode=True, blank=True)
+    mailinglist = models.ForeignKey(MailingList, on_delete=models.PROTECT)
+    recipient = models.ForeignKey(User, on_delete=models.PROTECT)
+    send_time = models.DateTimeField(auto_now=True)
+    # Open time is the time of the first instance of a beacon responding.
+    open_time = models.DateTimeField()
```
Shriukan33, line 781: If I understood correctly, this is set when one opens the email ? Because it loads an image at a certain URL, and the get triggers the update of open_time, if it wasn't already set ?

transport_nantes/topicblog/models.py
```diff
@@ -664,3 +664,121 @@ class TopicBlogEmailClicks(models.Model):
...
+    """Represent the fact that an press was clicked.
+
+    Note that TopicBlogPressSendRecord will record the first click,
+    but here we want to record all clicks that happen and where they
+    lead.
+
+    """
+    press = models.ForeignKey(TopicBlogPressSendRecord,
+                              on_delete=models.PROTECT)
+    click_time = models.DateTimeField()
```
Shriukan33, line 804: What's the difference in click_time in both classes ? 
The Record only register the first time while a TBPClicks is created on each click, is that right ?

Why saving specifically the first click in records while we can lookup the first occurrence with the FK ?

Shriukan33 (commented):
Some questions on commit e55a00cd3ed152708e52b1ad7d5c992d48997a10

Shriukan33 (commented):
commit fc4722fc81668b149b5a7fc64f2fbf4ed1226ba9 LGTM

Shriukan33 (approved):
LGTM commit e08c42f2a723873717fce3e02a0d1bd29fcd82d6

Shriukan33:
LGTM'd commits skipping the implementation of TBPress in e55a00cd3ed152708e52b1ad7d5c992d48997a10, let me rewind this

transport_nantes/topicblog/models.py
JeffAbrahamson: When we provide links (URLs) in emails, we will want to automatically decorate them with something that allows us to know they've been clicked.  How we do that is for tomorrow (well, later this week), not today.

One way is to add an argument (`?something=value`).  Another way is to wrap it, `https://www.mobilitains.fr/tb/redirect/<str:value>/<str:url-encoded-destination>`.  The redirect (need better name) path reads the value and knows it refers to a given send record, and it reads the next argument and 302's to it (assuming first arg is valid and the redirect goes to us).

Shriukan33: The get argument sounds simpler, although fakeable, but it shouldn't be a matter of concern, would it ?

transport_nantes/topicblog/models.py
Shriukan33: I miss worded it, I meant why save the first click in the PressSendRecord and save a PressClick at the same time, while you could query Pressclick to get the first record existing for a given PressSendRecord. 

I guess it's to not have to query, but it could be a FK, for what it's worth.

transport_nantes/topicblog/views.py
```diff
@@ -500,3 +500,156 @@ class TopicBlogEmailSend(LoginRequiredMixin, TemplateView):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['context_appropriate_base_template'] = 'topicblog/base_press.html'
+        return context
+
+
+class TopicBlogPressList(TopicBlogBaseList):
+    model = TopicBlogPress
+    permission_required = 'topicblog.tbp.may_view'
```
Shriukan33, line 556: There is no PermisionRequiredMixin to enforce this permission. BaseList will ask for Login (LoginRequiredMixin) but no permission is set at this point. So a logged in user would be able to access this view :slightly_smiling_face:

Shriukan33 (changes requested):

transport_nantes/topicblog/views.py
JeffAbrahamson: It's enough to add `PermisionRequiredMixin` in the inheritance list, right?

JeffAbrahamson:
> commit [7f5ebb3](https://github.com/transport-nantes/tn_web/commit/7f5ebb3384a2524f81293da00bf5c1af29c2b4f0) So no migration file for this change ?
> 
> Not sure about what problem you're trying to solve:
> 
> > Thus far all TB instantiations have been directly servable.  We'd like
> > to move beyond that constraint.  This commit lets us do that.
> 
> Why being servable is a constraint ?

There's no migration because it's abstract model inheritance, it doesn't lead to a database change.

TBLaunchers normally render as rectangles in the four "projects" panels.  So in that sense, they don't have social media data, they aren't independently servable.

transport_nantes/topicblog/views.py
Shriukan33: Yes, no need to have extra permissions checks for this like you did with the [...]One views, imho.

JeffAbrahamson: Done.

transport_nantes/topicblog/models.py
JeffAbrahamson: The principle convenience of the second approach, I think, is that the first requires intercepting _all_ requests to check for the flag whilst the second only engages on specific request.

Shriukan33:
> > commit [7f5ebb3](https://github.com/transport-nantes/tn_web/commit/7f5ebb3384a2524f81293da00bf5c1af29c2b4f0) So no migration file for this change ?
> > Not sure about what problem you're trying to solve:
> > > Thus far all TB instantiations have been directly servable.  We'd like
> > > to move beyond that constraint.  This commit lets us do that.
> > 
> > 
> > Why being servable is a constraint ?
> 
> There's no migration because it's abstract model inheritance, it doesn't lead to a database change.
> 
> TBLaunchers normally render as rectangles in the four "projects" panels. So in that sense, they don't have social media data, they aren't independently servable.

Alright, so it was prep work for TBLaunchers. I thought that launchers would be quite different from TBObjects, but I'm just arriving on the code, I'll make feedbacks once I had a chance to actually review your implementation

Shriukan33:
d55d44f27b9f9f60b27ec72caceb541ea13a5155 LGTM

Shriukan33:
dc9364b9b0c2b5f816b2926d2578ada4fd6c8d00 LGTM

transport_nantes/topicblog/models.py
Shriukan33: So your advice is the make a token, embed it in email urls (so all urls will point to `https://www.mobilitains.fr/tb/redirect/<str:value>/<str:url-encoded-destination>`, reverse the token and extract the information from there.

transport_nantes/topicblog/views.py
```diff
@@ -654,3 +654,64 @@ class TopicBlogPressSend(LoginRequiredMixin, TemplateView):
...
+    """
+    def has_permission(self) -> bool:
+        user = self.request.user
+        if self.request.method == 'POST':
+            return user.has_perm('topicblog.tbla.may_publish')
+        elif self.request.method == 'GET':
+            return user.has_perm('topicblog.tbla.may_view')
+        return super().has_permission()
+
+class TopicBlogLauncherEdit(TopicBlogBaseEdit):
```
Shriukan33, line 677: No permissions set

transport_nantes/topicblog/views.py
```diff
@@ -654,3 +654,64 @@ class TopicBlogPressSend(LoginRequiredMixin, TemplateView):
...
+class TopicBlogLauncherViewOne(TopicBlogLauncherViewOnePermissions,
+                            TopicBlogBaseViewOne):
+    model = TopicBlogLauncher
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        return context
+
+
+class TopicBlogLauncherList(TopicBlogBaseList):
```
Shriukan33, line 713: No permissions set here, like Press, TBOList doesn't set permissions, we need to add the appropriate permissions here

transport_nantes/topicblog/views.py
```diff
@@ -654,3 +654,64 @@ class TopicBlogPressSend(LoginRequiredMixin, TemplateView):
...
+
+class TopicBlogLauncherView(TopicBlogBaseView):
+    model = TopicBlogLauncher
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['context_appropriate_base_template'] = 'topicblog/base_launcher.html'
+        tb_object = context['page']
+        user = self.request.user
+        if user.has_perm('topicblog.tbp.may_send_self') or \
```
Shriukan33, line 697: This look like a copy paste, sendable is absent from the launchers templates, and it refers to tbp permissions.

Shriukan33 (changes requested):

transport_nantes/topicblog/views.py
```diff
@@ -681,6 +681,12 @@ class TopicBlogLauncherEdit(TopicBlogBaseEdit):
 
 
 class TopicBlogLauncherView(TopicBlogBaseView):
+    """This view is a bit curious.  In real life, TBLauncher's will be
```
Shriukan33, line 684: image

It seems that the display is broken

Shriukan33 (changes requested):
Commit a45373ecf923121f7995ee490888e0687606e56b

Shriukan33:
noticed actually that in commit e76be53, there is a double definition of get_context_data : 

```python
class TopicBlogEmailViewOne(PermissionRequiredMixin, TopicBlogBaseViewOne):
    model = TopicBlogEmail
    permission_required = 'topicblog.tbe.may_view'

    def get_context_data(self, **kwargs):
        pkid = self.kwargs.get('pkid', -1)
        the_slug = self.kwargs.get('the_slug', None)
        tb_email = TopicBlogEmail.objects.get(pk=pkid, slug=the_slug)
        context["email"] = tb_email
        return context

    def get_context_data(self, **kwargs):
        context['context_appropriate_base_template'] = 'topicblog/base_email.html'
        return context
```

transport_nantes/topicblog/views.py
JeffAbrahamson: Indeed, good catch, thanks!

transport_nantes/topicblog/views.py
JeffAbrahamson: Done.

mcmikashi:
- e55a00cd3ed152708e52b1ad7d5c992d48997a10 LGTM
- fc4722fc81668b149b5a7fc64f2fbf4ed1226ba9 LGTM
- e08c42f2a723873717fce3e02a0d1bd29fcd82d6 LGTM
- 6c79ecf477f61512c2a9d434d038e26cf015ce47 LGTM
- 6aaefe21be463c9fec11929c7548920290f78652 LGTM
- d55d44f27b9f9f60b27ec72caceb541ea13a5155 LGTM
- dc9364b9b0c2b5f816b2926d2578ada4fd6c8d00 LGTM
- 2677353b238cf05678ccdb85cdda4a1f47e068cd LGTM
- a104a2329f79fd417e071318d97c92c9bdccd30b (need to fix the url of the image on [content_launcher.html](https://github.com/transport-nantes/tn_web/pull/529/commits/a104a2329f79fd417e071318d97c92c9bdccd30b#diff-574a58bf890f469769ed37bbd319ffba953c9ac64df9246f481d86e581bbb518))

transport_nantes/topicblog/views.py
Shriukan33: I thought that launchers had images ? In place of "SEE OUR NEW LAUNCHER!"

JeffAbrahamson: In my test, I didn't upload an image.  I guess I'm taking on faith that the image works, I shouldn't do that. ;-)

transport_nantes/topicblog/models.py
JeffAbrahamson: That's correct.

Let's check in in when you start implementing this (at the level that you've had a rough stab at it so that your head's in it) and we can make sure we're thinking the same things.

PR #535, opened 2022-03-08
Tbe send implementation 532

Shriukan33:
This is very WIP but here is where I stand with this PR : 

> 1.  On GET, display a form.  That form should (for now) just show
the mailing_lists available in a dropdown list and let the user
choose one.  Once a mailing_list is chosen, enable a send button.
Pushing the send button will POST to the same url.

GET on the send URL will display a drop-down with all our mailing lists as described there 

>2.  On POST, send the mail.  This means you do the following:
    (i)  Get the list of users from the mailing_list by calling
         subscribed_users() from mailing_list/events.py.  Note that
         that function doesn't exist yet, but it should be a really,
         really simple function for you to write based on the other
         functions in the file.  You should make a single commit
         with that function and tests for that function.

This is done, I created a function `get_subcribed_users_email_list` for this purpose, alongside its tests.

> (ii) For each user in the list:
Compute a timed_token (function already exists,
make_timed_token() in asso_tn/utils.py).  Given it a
three-week timeout so as not to over-think.  Pass the
TopicBlogEmailSendRecord pk_id in persistent. 

This is @mcmikashi 's PR #517, so while this isn't merged, I started to create a function `_set_email_context` which I plan to make set the context, meaning the unsubscribe link at first and any other useful context variable. Because it's not finished yet, I set context in the `prepare_email` function for now.

> In the footer of the
base email template (note: this is 30 seconds, you just
write "<div><div><p><a href={% url
... %}>unsubscribe</a></div></div>" with maybe some
arguments to the divs and such.  DO NOT spend time now
fiddling with making it look just right.  There's a
difference between unworldly user interaction paradigms
and simply not being pretty yet.  The latter is easily
remedied in a second commit, the former is trickier.
This is a commit.

I happen to have already created a template that has a placeholder for the unsubscribe link : 
image

So once the function to have an unsubscribe link will be up, I'll be able to replace it in context.

>  Also compute the url (the path already exists) to view
this email on the web.  Put that in the context, too, and
make sure you add a link at the top of the email base
template, just above the content.  That's another commit.

This isn't integrated, both backend and template. It's TODO.

>Render the email, pass it off to SES for sending, and
write a record to TopicBlogEmailSendRecord.  This is
another commit.

I currently pass the email to SES and it gets send, but the TBESendRecord isn't created yet. TODO

All the rest isn't started yet.

I do have a question for @JeffAbrahamson , in your PR you created a lot of templates (base_xxx, content_xxx), I'm not sure about why we need a base_email and a content_email yet, I added the two templates needed to display in may clients, but the e/view/<slug> will display like a normal TBItem, I think.

### DIFF
--- a/transport_nantes/mailing_list/events.py
+++ b/transport_nantes/mailing_list/events.py
@@ -2,7 +2,7 @@ from django.contrib.auth.models import User
 from django.db.models import Max
 from django.db.models import Q
 
-from .models import MailingListEvent
+from .models import MailingList, MailingListEvent
 
 
 def subscribe_user_to_list(user, mailing_list) -> MailingListEvent:
@@ -13,7 +13,6 @@ def subscribe_user_to_list(user, mailing_list) -> MailingListEvent:
         mailing_list=mailing_list,
         event_type=MailingListEvent.EventType.SUBSCRIBE)
     subscribe.save()
-
 
 
 def unsubscribe_user_from_list(user, mailing_list) -> MailingListEvent:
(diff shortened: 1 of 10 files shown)


JeffAbrahamson:
You're already four commits off of master.  Why is this still a draft PR instead of several PRs?

Your description of WIP is good, but surely some of that is reviewable and mergeable?

Shriukan33:
Closing this PR to make smaller ones

PR #538, opened 2022-03-08
Add test for TopicBlogLauncher

mcmikashi:
Last part of 530.
Need some fix

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/content_launcher.html
+++ b/transport_nantes/topicblog/templates/topicblog/content_launcher.html
@@ -53,7 +53,14 @@
      <div id="app_content" class="d-flex flex-column col-sm-12 col-md-10 col-lg-9 m-auto">
     <div class="col-md-3">
 	<div class="thumbnail">
-	    <a href="{% url 'topic_blog:view_item_by_slug' page.slug %}">
+        {% comment %}
+        This page only happens during launcher development.
+        The slug that is used for the link is the article slug
+        and this field is required before publication.
+        {% endcomment %}
+        {% if page.article_slug %}
+            <a href="{% url 'topic_blog:view_item_by_slug' page.article_slug %}">
+        {% endif %}
 		<div class="shadow-md">
 		    <img class="rounded"
 			 src="{{ page.launcher_image.url }}"
@@ -68,7 +75,9 @@
 			{% tn_markdown page.launcher_text_md %}
 		    </p>
 		</div>
-	    </a>
+        {% if page.article_slug %}
+	        </a>
+        {% endif %}
 	</div>
     </div>
      </div>
--- a/transport_nantes/topicblog/test_topic_blog_part_2.py
+++ b/transport_nantes/topicblog/test_topic_blog_part_2.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timezone
 from django.test import TestCase, Client
-from .models import TopicBlogEmail, TopicBlogPress
+from .models import TopicBlogEmail, TopicBlogPress, TopicBlogLauncher
 from django.contrib.auth.models import Permission, User
 from django.urls import reverse
 
@@ -12,8 +12,9 @@ class TBETest(TestCase):
                                              password='test-pass')
         self.user.save()
         # Create a user with all permission
-        self.staff = User.objects.create_superuser(username='test-staff',
-                                                   password='test-staff')
+        self.staff = User.objects.create_user(username='test-staff',
+                                              password='test-staff')
+
         edit_permission = Permission.objects.get(codename="tbe.may_edit")
         view_permission = Permission.objects.get(codename="tbe.may_view")
         publish_permission = Permission.objects.get(codename="tbe.may_publish")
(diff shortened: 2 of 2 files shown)


mcmikashi:
Got some weird error on TB Launcher when testing the page that use topicblog:view_launcher_by_pkid_only as name with the staff user:
 ```
   url = reverse(view_name, args=args, kwargs=kwargs, current_app=current_app)
  File "M:\studi\mobilitains\tnbonenv\lib\site-packages\django\urls\base.py", line 86, in reverse
    return resolver._reverse_with_prefix(view, prefix, *args, **kwargs)
  File "M:\studi\mobilitains\tnbonenv\lib\site-packages\django\urls\resolvers.py", line 698, in _reverse_with_prefix
    raise NoReverseMatch(msg)
django.urls.exceptions.NoReverseMatch: Reverse for 'view_item_by_slug' with arguments '('',)' not found. 1 pattern(s) tried:
['tb/t/(?P<the_slug>[-a-zA-Z0-9_]+)/\\Z']
```

**Review by Shriukan33 — CHANGES_REQUESTED** [review]
Figured out the origin of this bug : 
Notice the test mentions "view_item_by_slug", while you are testing "view_launcher_by_pkid_only"
Why would this happen only on staff ? What different when you are staff ? An anonymous visitor is redirected to login, and if you don't have the permission you 403.The staff user has the appropriate permission so it actually renders the template ! 

If you scan the template, you can see this :
https://github.com/transport-nantes/tn_web/blob/0e154fa9ac5ccdbbce8b75a8d730cf19ef1982dd/transport_nantes/topicblog/templates/topicblog/content_launcher.html#L56

In this test, you're testing what happens if we try to access a TBLA item that has no slug : 
https://github.com/transport-nantes/tn_web/blob/0e154fa9ac5ccdbbce8b75a8d730cf19ef1982dd/transport_nantes/topicblog/tests_topic_blog_epl.py#L890

So once it gets to the rendering part, it has no argument to give in the template resulting in the error you describe. 

I wrote this to fix it : 
```
        {% if page.slug %}
            <a href="{% url 'topic_blog:view_item_by_slug' page.slug %}">
        {% else %}
            <a href="#">
        {% endif %}	    
            <div class="shadow-md">
                <img class="rounded"
                {% if page.launcher_image %}
                    src="{{ page.launcher_image.url }}"
                    alt="{{ page.launcher_image_alt_text }}"
                {% else %}
                    {% comment %}Puts a grey placeholder image 667x667 px{% endcomment %}
                    src="https://via.placeholder.com/667"
                {% endif %}
                
                style="width:100%">
            </div>
            <div class="caption pt-2">
                <h3 class="font-weight-light"
                style="font-variant: all-small-caps;
                    font-size: 150%;">{{ page.headline }}</h3>
                <p>
                {% tn_markdown page.launcher_text_md %}
                </p>
            </div>
```

Notice that another error would occur because the launcher is slug less but also image-less, so calling `page.launcher_image.url` would also have failed.

Shriukan33 (commented):
I think we can use tb_edit_template.html ?

mcmikashi:
Yes I use this file on my next PR.

JeffAbrahamson:
9087bf7 adds a template but doesn't use it, what is this for?  The commit comment gives no clue.

JeffAbrahamson:
eb80ba846 commit comment doesn't explain what was wrong, only that you are fixing it.  Could you give a proper explanation of the problem and the solution?

If there is no slug, then your link will go nowhere.  Why do you render it at all?  Also, this should never happen for unprivileged users, only for editors whilst making a new launcher but not done yet.  That is, why would we want a launcher with no slug?  (I'm assuming `page.slug` is being set somewhere to `article_slug`.)

Is this not a reasonable constraint for the "ready to publish" function?

JeffAbrahamson:
2a5b37ce7 - given that you're on your own branch, might this not have been squashed into `2a5b37ce7^`?

mcmikashi:
> If there is no slug, then your link will go nowhere. Why do you render it at all? Also, this should never happen for unprivileged users, only for editors whilst making a new launcher but not done yet. That is, why would we want a launcher with no slug? (I'm assuming `page.slug` is being set somewhere to `article_slug`.)

Yes this error only happen on staff user no auth will be redirected to auth and normal user get a 403.

mcmikashi:
- I update the message of all commit 
- I delete the link when the object doesn't have a slug
- I sqash the fix comment commit into the last commit

JeffAbrahamson:
Come on, six months from now no one will understand your comment in 59aac0e3.  What test fails?  Why does it fail?  How does adding this template change anything?

JeffAbrahamson:
a06502e309a - Do you want to add a comment to the template that explains that this view only happens during launcher development and that the slug is really `article_slug` here and that it is required for publication?

mcmikashi:
- Update the commit message of first and second commit.
- add comment on the template content_launcher.html

JeffAbrahamson:
This appears to be a single commit based on code that is under review elsewhere, so I'll hold off on review until the other code is merged and this rebases.

What does "epl" mean?

mcmikashi:
> What does "epl" mean?

epl mean Email Press Launcher

JeffAbrahamson:
> > What does "epl" mean?
> 
> epl mean Email Press Launcher

How will anyone ever know that?

mcmikashi:
- I update the first and second commit message
- I rename the file tests_topic_blog_epl.py  to tests_topic_blog_part_2.py
Maybe I can split the file into 3 files ?

JeffAbrahamson:
> Last part of 530. Need some fix

You probably don't want to create PRs whose first line says "please don't look at me".  If it needs fixing, just mark it as a draft.

PR #544, opened 2022-03-10
Tbe view templates 532

Shriukan33:
We're going to use 2 sorts of templates to display TBEmails:
- One to be displayed in mail clients, which heavily relies on inline
styles and HTML tables
- One to be displayed on our website, with a more article looking way,
that will allow the user to browse the website normally.

The 1st kind is introduced with these commits, and the 2nd has already been added in #529, and updated here to fit the current implementation of TBEmails.

On the note of templatetags, it's true that for now, one might not be able to use all those available for TBItem (mostly thinking about custom markdown ones like `[[verb]]((arg))`) although I haven't tried to display them to see how it reacts.
So far I've provided fields that I know will display properly in both templates, but lack of freedom in the _md fields.

I think we could make it better by either providing alternatives of processing the markdown in different ways depending on what's the purpose : `[[CTA: "Je signe"]]((the-slug))` would produce different results depending if we are rendering a mail or a classic content.

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/base_email_client.html
+++ b/transport_nantes/topicblog/templates/topicblog/base_email_client.html
@@ -1,0 +1,130 @@
+<!DOCTYPE html>
+<html lang="en" xmlns="https://www.w3.org/1999/xhtml" xmlns:o="urn:schemas-microsoft-com:office:office">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <meta name="x-apple-disable-message-reformatting">
+    <title></title>
+    {% comment %}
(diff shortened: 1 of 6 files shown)


mcmikashi (approved):
LGTM

JeffAbrahamson:
It mostly looks good.

I find it strange that the tags that markdown will call live in files in totally different applications.  I'd also think it clearer to rename the existing tags so that they are called `web_*` to make the naming clearer and more parallel.  That can come later.

I'm not immediately clear how we will pass the information to markdown, but I'd think initially that we'll set something in `context`, since it's the view that knows what the usage is and this is the standard way in django to communicate between view and renderer.

Shriukan33:
Opened https://github.com/transport-nantes/tn_web/issues/547

PR #548, opened 2022-03-16
Add a form to generate signatures for emails

Shriukan33:
Default values are user's info if any, they are overridden by whatever the user would input in the form.

Simply copy paste following the [guide](https://docs.google.com/document/d/13HbNW_dRQElrP3fgEcYDrmw08mqQ6ETWwzj25NNOzr8/edit)


JeffAbrahamson:
LGTM.

Curious effect in browser, but maybe it works in mail!

image

Issue #543, opened 2022-03-09
Update our publicaction tracking model for all TBObjects


JeffAbrahamson:
We unintentionally created a bit of a mess.  When we publish an object, we unpublish all others with the same slug.  This is reasonable.  But `first_publication_date` tracks when _this particular slug_ was first published.  As a result, we have no trace of which versions of a slug were previously published.  And this makes clean-up hard, because we can be relatively aggressive in cleaning up "old" things that have never been published, but things that were once live we might want to keep longer.

Now, admittedly, we have the `publisher` field, and this at least tells us that the object was once live.  But it doesn't tell us when.

Consider that some objects already have dozens of versions.  Fast forward a couple years.  This will be a mess.

A proposal is that we add a field `unpublish_date` that tells when this particular object stopped being live.  Then in `TopicBlogBaseViewOne.post()` we add one extra update command:

```
    self.model.objects.filter(
        id=tb_object.id).update(unpublish_date=datetime.datetime.now())

```

Then in the dashboard we drop the word about _retiré_ and instead say "unpublished on ..." (dépublié ?).

JeffAbrahamson:
Can template_name be a dropdown of available templates like for TBItem?

Shriukan33:
All items that have a first_publication_date have been once published, those who don't never have. We could clean up those who don't have one, as it's never reset, the code goes like this : 
```python
self.model.objects.filter(
                    slug=tb_object.slug).exclude(
                        id=tb_object.id).update(publication_date=None)
                tb_object.save()
```
Only publication_date is reset. So I think it would be safe to delete those who don't have a 1st publication date, but the risk is that older pages might not have one, as I'm not sure about when we introduced that field.

To solve that, we could delete those who don't have a first_publication_date and have a created_at > [date we introduced that field] 

What do you think ?

JeffAbrahamson:
Quick taxonomy:
*  **Page is active**
   * `publication_date` is non-null
   * `publisher` is non-null
* **Page was active**
   * `publication_date` is null
   * `publisher` is non-null
* **Page has never been active**
   * `publication_date` is null
   * `publisher` is null

So, except for quite old articles, we can already tell the difference.

I think the main point is that it would be helpful to know how long it's been since something was unpublished.

New taxonomy:
*  **Page is active**
   * `publication_date` is non-null
   * `publisher` is non-null
   * `unpublish_date` is null
* **Page was active**
   * `publication_date` is null
   * `publisher` is non-null
[...]

Issue #364, opened 2022-01-02
Track all users who touch (edit) TBv2 items


JeffAbrahamson:
If a user edits a TBItem, we should enter them in a table so that we can email all affected users when a page changes status.  A status change is publish, withdrawal, or new draft.

The link should be to the slug.  So this is a table with slug pointing to TBItem and a user and a timestamp.

We might also want people to be able to unsubscribe from announcements, but leave that for now.  We'll just prepare it by adding a nullable timestamp field for unsubscribe.  That way at least your tests can confirm that you're only selecting people who have not yet unsubscribed.

Ignore for now actually sending anyone mail.

Shriukan33:
Can you confirm this is the kind of workflow you want ? 

Gaelle goes to /tb/admin/new and creates a content. Slug : "mySlug"
-> Who should receive a mail ? 
I edit Gaelle's "mySlug" item, so it creates a new copy. 
-> Who should receive a mail ? 
Gaelle withdraws the previous version and publishes the new one. 
-> Same question 

Also, you suggest creating a new table containing the following fields : 
slug (charfield) 
user (ForeignKey)
timestamp (DateField, blank=True)

I guess we should also have a table that list the evbent we're listenning with fields like this : 
TBI_slug (CharField) / listenner (ForeignKey with user ID) / Active (Bool) 
So we can check that table and filter Active = true and the appropriate TBI slug to send emails  ?

JeffAbrahamson:
User U1 creates a new TBItem T with slug my-slug.  This issue is only about tracking, not about emailing, but we'd record in the TBItem tracking table that U1 touched my-slug at timestamp.  No one is mailed.

Later U2 edits T (creating a new copy).  The tracking table gets a record "U2 touched my-slug at timestamp".  U1 gets mail (not this issue).

Now U1 makes some other change (withdraws, publishes, edits).  Again, tracking says "U1 modified my-slug at timestamp".  U2 gets mail (not this issue).

The table: yes, you've got it.  But see next:

I complicated things with my initial description.  I think we should do this like we do with mailing lists: just the events.  If someone unsubscribes (not this issue), we just write an unsub *event*.  This lets us get a feel for edit activity even after unpublished drafts are cleaned up.

So then the table (TopicBlogEvents) is

```
class TBEventType -- SUBSCIRIBE, UNSUBSCRIBE

slug = CharField
user = ForeignKey
event_type = CharField(choices=EventType)
[...]

PR #551, opened 2022-03-16
Add possibility to retire TBItem from TBIViewOne

Shriukan33:
This is a sort of proof of concept for a retire view.
This view allows to retire a TBItem from a TBIViewOne, basically
it sets a given item's publication_date to None.

Part of #462

As for the other parts of #462 :

> Quick notes of quick things to fix as I'm sorting other details this evening:
> 
> - servable doesn't exist any more, but it's referenced in base_topicblog.html

Fixed here with the addition of "Retirer" button.

> - the template_name pop-down is lost on new edit, though I think that's fixed once the permissions branch is merged

I believe this must be old because the template drowpdown is currently active

> - views.py still computes servable_object, but it's a bit unclear (to me, this evening) why

This is because it's used to tweak the display in ListOne, it's says "publié depuis", add boldness on appropriate ID ...

> -  tests are missing (as noted elsewhere)

I think It's fine now since Mickael joined, I added more tests for the retire view.

> -  there's a bug in the list sort order (I'll send a PR)

You fixed that since then :)


Shriukan33:
Yeah it's sort of an old issue about what's published or not, while it's true it's not really relevant now as we might change it in   #543 , it deletes the "republier" button that never worked and the servable reference :laughing: 
Not sure what to do with this PR now

JeffAbrahamson:
I'd say let's implement #543 and #364 and then look back at this and decide if there's anything left to consider.

PR #555, opened 2022-03-17
Fix press

mcmikashi:
-  fix the list at top appears not to be limited
- fix the warning on test


JeffAbrahamson:
LGTM.

Why is the ordering different in the model
```
ordering = ['-article_publication_date', 'newspaper_name']
```

vs in the view
```
ordering = ['-article_publication_date']
```

PR #556, opened 2022-03-18
Improve press with open graph and twitter

mcmikashi:
- I was thinking about the data. Should I take only the image since we got all the other data in the press mention table ? 
- Should I keep og and twitter ? (i think all use og)
-  On edit PressMention should I keep the previous OpenGraphTwitter data and create a new one ?
- Did you see other things that I could improve on this ?


JeffAbrahamson:
* I think we can just call this open graph, even if we fall back on twitter if OG is for some reason not available.  But I'd expect OG always to be available if either are.
* Using the selenium driver in the model (that is, outside of testing) is the wrong way to do this.  We absolutely don't want a chrome instance on a production host.  Python has many good http parsing and request tools, you should use them.

PR #561, opened 2022-03-21
Tbe create send record 532

Shriukan33:
Create send records on email sending, one for each recipient.

This PR also tweaks the model that had incoherent database requirements (like open time that can't be null), and adds the send record to admin.
The send records are searchable / order-able in Django admin.

PR #563, opened 2022-03-21
Add unsub link 406

### LINKED ISSUE #406: Create unsub links and page
Sent TBEmails should carry an unsub link that allows us to identify from what email it's been clicked.
We should make a POST request to validate the unsub action (with a button saying "I want to unsuscribe from X mailing list").

Shriukan33:
This PR is based on #561 (TBE-create-SendRecord-532)

It adds an URL that allows the user to unsub from a mailing list by parsing the token.
A confirmation is asked from the user, and they are then shown a confirmation page.

The link is embedded inside TBEmails when they are rendered to send via emails.

Closes #406

I leave it as a draft because the tests aren't written yet.

### DIFF
--- a/transport_nantes/topicblog/views.py
+++ b/transport_nantes/topicblog/views.py
@@ -22,7 +22,7 @@ from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.urls import reverse, reverse_lazy
 from django.utils.html import strip_tags
 
-from asso_tn.utils import StaffRequired, token_valid
+from asso_tn.utils import StaffRequired, make_timed_token, token_valid
 from mailing_list.events import (get_subcribed_users_email_list,
                                  unsubscribe_user_from_list)
 from mailing_list.models import MailingList
@@ -613,7 +613,7 @@ class TopicBlogEmailSend(PermissionRequiredMixin, LoginRequiredMixin,
                 the_slug=tbe_slug,
                 recipient=[recipient],
                 mailing_list=mailing_list,
-                send_record=send_record.id)
+                send_record_id=send_record.id)
 
             logger.info(f"Successfully prepared email to {recipient}")
             custom_email.send(fail_silently=False)
@@ -622,11 +622,13 @@ class TopicBlogEmailSend(PermissionRequiredMixin, LoginRequiredMixin,
         return super().form_valid(form)
 
     def prepare_email(self, pkid: int, the_slug: str, recipient: list,
-                      mailing_list: MailingList, send_record: int) \
+                      mailing_list: MailingList, send_record_id: int) \
             -> mail.EmailMultiAlternatives:
         """
         Creates a sendable email object from a TBEmail with a mail
         client-friendly template, given a pkid, a slug and a mail adress.
+
+        send_record is the pk_id of the TopicBlogEmailSendRecord
         """
         if pkid < 0 or the_slug is None:
             logger.info(f"pkid < 0 ({pkid}) or slug is none ({the_slug})")
@@ -635,17 +637,8 @@ class TopicBlogEmailSend(PermissionRequiredMixin, LoginRequiredMixin,
         # Preparing the email
         tb_email = TopicBlogEmail.objects.get(pk=pkid, slug=the_slug)
         self.template_name = tb_email.template_name
-        # The context statements will be moved to a _set_context method
-        # once all the features are implemented.
-        context = dict()
-        context["context_appropriate_base_template"] = \
-            "topicblog/base_email.html"
-        context["email"] = tb_email
-        context["host"] = get_current_site(self.request).domain
 
-        # The unsubscribe link is created with the send_record and the
-        # user's email hidden in a token.
-        context["unsub_link"] = self.get_unsubscribe_link()
+        context = self._set_email_context(recipient, send_record_id, tb_email)
 
         try:
             email = self._create_email_object(tb_email, context, recipient)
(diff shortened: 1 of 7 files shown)


JeffAbrahamson:
I don't understand what the point of 1e5cd26 is.

The function `make_timed_token()` takes a string (email), and an int (persistent).

How is what you've done different from writing
```
def make_very_special_benja_timed_token(email, minutes, send_record_id, test_value_now=None):
    return make_timed_token(email, minutes, send_record_id, test_value)
```
?

transport_nantes/topicblog/views.py
```diff
@@ -728,6 +728,34 @@ def get_unsubscribe_link(self) -> str:
         """
         pass
 
+
+class UnsubscribeFromMailingListView(TemplateView):
+    """
+    View to unsubscribe a user from a mailing list.
+    """
```
JeffAbrahamson, line 744: In view of the name of the view, this comment doesn't appear to be adding much value. ;-)

Shriukan33:
> I don't understand what the point of [1e5cd26](https://github.com/transport-nantes/tn_web/commit/1e5cd26afea9553edb9ebb420b63efd04ef0a1e4) is.
> 
> The function `make_timed_token()` takes a string (email), and an int (persistent).
> 
> How is what you've done different from writing
> 
> ```
> def make_very_special_benja_timed_token(email, minutes, send_record_id, test_value_now=None):
>     return make_timed_token(email, minutes, send_record_id, test_value)
> ```
> 
> ?
> 
> And if you are going to do that, you don't have to do that, because you can just call `make_timed_token()` directly.

Why do we need persistent flag then ? It's supposed to be 0 or 1, but now you drop it to let it be any number, wasn't it important for the login initially ? I didn't track its role in the code, preferred adding a dedicated variable instead of "if this is superior to 1 this is a send record and else it's a persistent flag", which sounds clunky

Shriukan33:
It seems that the jobs (GitHub Action and CodeCI) aren't running, maybe it's related to the problems we had with GitHub earlier.
So just as a sanity check, here are the tests : 
image

JeffAbrahamson:
In the commit where you ask for unsub confirmation, you don't appear actually to ask a question:

> Vous vous êtes désabonné(e) de la liste de diffusion {{ ... }}

There's no "?" and then there are yes/no buttons.

Why do you want/need javascript on the "no" button instead of marking it as a cancel action?  Every time you add javascript, you're adding a potential browser incompatibility, so there should be something you're getting in return.

Shriukan33:
> In the commit where you ask for unsub confirmation, you don't appear actually to ask a question:
> 
> > Vous vous êtes désabonné(e) de la liste de diffusion {{ ... }}
> 
> There's no "?" and then there are yes/no buttons.
> 
> Why do you want/need javascript on the "no" button instead of marking it as a cancel action? Every time you add javascript, you're adding a potential browser incompatibility, so there should be something you're getting in return.

I actually replaced the JS button with a normal button looking link, I think I messed up during my rebase, let me fix this :)

Shriukan33:
Added tests ! Also merged the context setting in a single function

JeffAbrahamson:
Please avoid magic numbers, even magic numbers with comments before them.  Instead, use a constant, something like `kMinutesInSixMonths = 60 * 30 * 6` and then use `kMinutesInSixMonths` in your function call.

JeffAbrahamson:
The function `get_unsubscribe_link()` takes an int (line 719), but you call it with a SendRecord (line 700).  Why does this work?

It would make more sense to call it with a SendRecord and use that value.

JeffAbrahamson:
In general, when a function does not access class members, it's better written as a stand-alone function than as a class method.  So by this test, `get_unsubscribe_link()` doesn't really merit being a class method.

Shriukan33:
> Please avoid magic numbers, even magic numbers with comments before them. Instead, use a constant, something like `kMinutesInSixMonths = 60 * 30 * 6` and then use `kMinutesInSixMonths` in your function call.

Fixed



> The function `get_unsubscribe_link()` takes an int (line 719), but you call it with a SendRecord (line 700). Why does this work?
> 
> It would make more sense to call it with a SendRecord and use that value.

It wasn't a SendRecord Object that was passed, only its ID. It wasn't clear in the code, so I added a _id suffix to clear the confusion !



> In general, when a function does not access class members, it's better written as a stand-alone function than as a class method. So by this test, `get_unsubscribe_link()` doesn't really merit being a class method.

Actually it accessed `self.request` but I could just pass self.request as any other arg, which also works. So I did that to get `get_unsubscribe_link()` out of the class.


<hr>
Also modified a URL path name in transport_nantes/urls.py because of a duplicate name ("index") that was already taken. 
It made a test using `reverse("index")` fail.

JeffAbrahamson:
> It wasn't a SendRecord Object that was passed, only its ID.

Self-documenting variables names are important!

> Actually it accessed self.request but I could just pass self.request as any other arg

Oh, I missed that. In that case, what you had before made more sense, unless you anticipate the function being useful from multiple classes.

It's all about communication.  Declaring it class communicates the range of utility and dependence.  Declaring it independent says something similar but in the opposite direction.

JeffAbrahamson:
5167421 - LGTM

JeffAbrahamson:
e3facda7 - LGTM

Remember that you introduce blank lines in your HTML with comments.  You can either strip the nl (I think, if I'm not thinking of jinja2) or else not include the nl at all (comment on same line as directive that follows).  Although, to be fair, we should also be minimising our html, css, and js, and we're not.

JeffAbrahamson:
e334ce21e - In `get_unsubscribe_link()`, the argument `pk_id` clearly points to a table, but you don't say which one.  This makes your code harder to understand.

As noted in slack, it probably makes sense for this to be a member function given that it refers to `self.request`, especially since you define it immediately after the class and the url manipulation you do is clearly specific to this view class.  And given that you are only using it here, then the variable is `send_record_id` or similar.

JeffAbrahamson:
1c3cd722 - good point ;-)

Shriukan33:
> [e334ce2](https://github.com/transport-nantes/tn_web/commit/e334ce21e6bf4320109d321486d346ce106acd71) - In `get_unsubscribe_link()`, the argument `pk_id` clearly points to a table, but you don't say which one. This makes your code harder to understand.
> 
> As noted in slack, it probably makes sense for this to be a member function given that it refers to `self.request`, especially since you define it immediately after the class and the url manipulation you do is clearly specific to this view class. And given that you are only using it here, then the variable is `send_record_id` or similar.

I've fixed this by replacing "pk_id" with "send_record_id" so we know it's a send record, and it's more consistent with the rest of this class' code.

JeffAbrahamson:
LGTM.

PR #564, opened 2022-03-22
Improve timed token security

JeffAbrahamson:
Our timed tokens were emminently forgeable, although it seemed quite a
lot of work for an adversary to do so for the very little gain.  This
commit addresses that issue by upping (significantly) the amount of
work necessary to forge a timed token.  We believe the difficulty of
forging a timed token is now on a par with the difficulty of breaking
SHA1 or SHA256.

This increases the length of our timed tokens, but we don't currently
consider that to be a problem.

### DIFF
--- a/transport_nantes/asso_tn/utils.py
+++ b/transport_nantes/asso_tn/utils.py
@@ -10,8 +10,17 @@ from django.utils.crypto import secrets
 from django.utils.http import base36_to_int, int_to_base36
 from django.utils.http import urlsafe_base64_encode, urlsafe_base64_decode
 from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
+from hashlib import sha1, sha256
+
 
 logger = logging.getLogger("django")
+
+
+# Don't reveal our secret key, but we'd like something that is
+# nonetheless our secret so that we can validate timed tokens we
+# created.  SHA1 is broken (with great effort), but we'll wrap it in
+# SHA256 below.
+hashed_secret = sha1(settings.SECRET_KEY.encode()).hexdigest()
 
 
 def make_timed_token(string_key, minutes, int_key=0, test_value_now=None):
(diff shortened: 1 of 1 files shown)


transport_nantes/asso_tn/utils.py
```diff
@@ -56,8 +70,9 @@ def make_timed_token(string_key, minutes, int_key=0, test_value_now=None):
     int_key_str = int_to_base36(int(int_key))
     hmac = salted_hmac(soon_seconds + int_key_str, rand_value + str(string_key)
                        ).hexdigest()[:20]
-    token = '{rnd}|{e}|{t}|{p}|{h}'.format(
+    token = '{rnd}|{hr}|{e}|{t}|{p}|{h}'.format(
```
Shriukan33, line 73: Adding another element to the token will raise a ValueError when other functions using this token will unpack the resulting token.

JeffAbrahamson: Isn't the only function to do that `token_validate`?

JeffAbrahamson: Yes, there's a race condition.  All those pending tokens out there...  ;-)

JeffAbrahamson: Though that's a reminder that we need to make sure we're always robust against a bad token coming in, offering something reasonably logical to help the user recover.

Shriukan33: The token_valid function will return a variable number of values, one for each value seperated by a pipe.
And we need to store the values out of token_valid, to extract the email, and send_record_id if any.

JeffAbrahamson: The job of `token_valid()` is to extract the two values we put in.  No other entity should be trying to parse the token and expecting not to be fragile.

Can you construct a scenario and show me in code how you could call `make_timed_token()` and then pass the token to `valid_token()` and have the string and int that you recover not be what you put in?

Shriukan33:
LGTM

Shriukan33 (approved):

Issue #560, opened 2022-03-21
Enable failing tests


Shriukan33:
Some tests will fail because of a technical limitation of sqlite3 used as a test database and using selenium.
You can find a ticket on django dev project relating this bug, it's open as I write those lines : https://code.djangoproject.com/ticket/29062#no2

While we figure out a solution (using pgsql in dev env), Some tests are commented out until we can run them without this limitation.

# Mailing_list app
## tests.py : MailingListIntegrationTestCase
* testing_quick_form_logged_out
* testing_quick_form_fail_captcha_logged_out
* testing_quick_form_fail_captcha_logged_in

Issue #383, opened 2022-01-09
Pre-populate email if logged in


JeffAbrahamson:
If the user is authenticated (and so we know her/his email address), don't even ask for the email address in MLP forms, nor the captcha.  Just do signup directly.

Shriukan33:
Issue #391 is the first step to make it work

jyoost:
@Shriukan33. Are you going to do anything with the django_cms integration?

Shriukan33:
One-click subscriptions are implemented now, closing

PR #565, opened 2022-03-22
Enable easier mailinglist signup for authenticated users

JeffAbrahamson:
Just taking advantage of github's automated tests before merging.  This is the code we agreed yesterday (Benja+Jeff), squashed.  But the rebase had conflicts, and whilst I don't _think_  I made any mistakes, let's let github's tests opine further.

--- 

If the user is authenticated, we needn't ask for email address or
present a captcha.  Otherwise, ask for email and require captcha.

In addition, remove kludge in which we checked form inheritance to
decide in what case we found ourselves.  Instead, just read the fields
we expect from the form and behave accordingly.

Part of #383.

Commented out failing tests (see #560)
Some parts of the views didn't work as intended.

### DIFF
--- a/transport_nantes/mailing_list/forms.py
+++ b/transport_nantes/mailing_list/forms.py
@@ -67,7 +67,7 @@ class SubscribeUpdateForm(Form):
 
 
 class FirstStepQuickMailingListSignupForm(Form):
-    email = forms.EmailField(label="Adresse mél", required=True)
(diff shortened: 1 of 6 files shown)


JeffAbrahamson:
The failing tests are in mailing_list.

Shriukan33:
`user = User.objects.filter(email="admin@mail.com")[0]`

Out of range, remember :)

PR #568, opened 2022-03-22
Add title and description on edit page

mcmikashi:


JeffAbrahamson:
LOL.  Thanks.  This is the right thing from a programming standpoint, but misses the mark slightly for the humans.

Non-dev editors don't really care what id they are modifying, and the title/subject/slug are visible in the form.  All that's missing is a reminder that this is an article, an email, a press release, etc.

Also, it's not an emergency, it's just background information to guide people on their way, so the class alert is perhaps overkill.

Could you modify just this part:
```
+ <h2 class="text-center">Emails</h2>
+ {% if tb_object.id %}
+         <div class="alert alert-primary text-center" role="alert">
+                 <h5>Vous modifiez l'élément suivant :</h5>
+                 <p>id : {{tb_object.id}}</p>
+                 <p>slug : {{tb_object.slug}}</p>
+                 <p>titre : {{tb_object.title}}</p>
+                 <p>sujet : {{tb_object.subject}}</p>
+         </div>
+ {% endif %}
```

The h2 is fine, the rest is perhaps useful for devs, so just put it in an `{% if dev %}` (I think there's a flag for that, I don't remember, else if admin or something and we'll fiddle with it as we go.  And in that case, don't use so much vertical space (the bullets), just put it all on one line, separated with commas or "/" or something.

Thanks!

JeffAbrahamson:
Merged.

PR #569, opened 2022-03-23
Fix failing test mailinglist

Shriukan33:
The test initially made a few mistakes :
The quick form doesn't have a mail field when the user is connected,
and the title of the test "testing_quick_form_logged_in" implies that
the user is logged in. Instead of that simply a "s'inscrire" button
is displayed.
There isn't a captcha to fill either, as we don't ask for one for logged
in users.
The test was checking for existence of a user that never was created
"admin@mail.com"

This commit fixes the above :
- logs in a user (by adding a session cookie to the browser)
- clicks on the button without trying to fill an absent email field
- Checks for presence and correct value of the subsequently created
MailingListEvent


JeffAbrahamson:
LGTM.

PR #570, opened 2022-03-23
Remove immutable status of slug field in TopicBlog Item

mcmikashi:
Add a warning if the slug field  change

If this is good i will do the same on other TB

### DIFF
--- a/transport_nantes/topicblog/forms.py
+++ b/transport_nantes/topicblog/forms.py
@@ -20,10 +20,6 @@ class TopicBlogItemForm(ModelForm):
         super().__init__(*args, **kwargs)
         self.instance: TopicBlogItem
 
-        # When an item is published, its slug becomes frozen (unmodifiable).
-        if self.instance.publication_date:
-            self.fields['slug'].widget.attrs['readonly'] = True
-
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
On saving form with modified slug, this happens:
image

JeffAbrahamson:
Otherwise, generally looks good.  The js feedback is not quite perfect, in that modifying the slug and then going back to unmodified is not quite the same as not modifying.  For example, it warns me about how many items there are.  Either fix now (if it's easy) or make an issue to remember to fix later (if it will take a few hours).

mcmikashi:
> On saving form with modified slug, this happens: image

I dont have this error  the slug you used was "go!" ?

mcmikashi:
> For example, it warns me about how many items there are

This is not the same file, did you want to hide this if it was the previous slug ?

mcmikashi:
- Update the logic

JeffAbrahamson:
Please fix your spelling in `edit_slug_warning.js`: s/warnning/warning/;

JeffAbrahamson:
LGTM.

PR #573, opened 2022-03-24
Add launcher template tags

mcmikashi:
Add the launcher template tags

### DIFF
--- a/transport_nantes/topicblog/templatetags/test_templatags.py
+++ b/transport_nantes/topicblog/templatetags/test_templatags.py
@@ -2,7 +2,9 @@ from django.conf import settings
 from django.template import Template, Context
 from django.test import TestCase
 from django.urls import reverse_lazy
-
+from django.contrib.auth.models import User
+from topicblog.models import TopicBlogItem,TopicBlogLauncher
+from datetime import datetime, timezone
 
 class TBEmailTemplateTagsTests(TestCase):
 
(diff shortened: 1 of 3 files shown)


transport_nantes/topicblog/templatetags/test_templatags.py
```diff
@@ -94,3 +96,62 @@ def test_email_cta_button(self):
...
+            user=self.admin,
+            article_slug=self.item.slug,
+            template_name="topicblog/content_launcher.html",
+            headline="Headline")
+        self.launcher.save()
+
+    def test_launcher(self):
+        url = reverse_lazy("topic_blog:view_item_by_slug",
+                           args=[self.launcher.article_slug])
+        exepted_template =\
```
Shriukan33, line 133: excepted

Shriukan33 (approved):
Except for the typo, LGTM

JeffAbrahamson:
Mostly looks good, but two issues jump out for me:

1) The template doesn't seem to resemble what we're currently doing (cf `asso_tn/templates/asso_tn/mpanel/projects.html`).

2) The test is checking for the precise HTML in the template.  This is overly specific.  If we change something about the presentation, the test will fail.  But we want to check that the thing is doing _functionally_ what we want, not that it's bit-for-bit the same forever.  So pick out things like the title and the image and such and make sure that they are present in elements that are tagged as you expect them to be tagged.  Just don't do a full string match on the whole block of HTML.

In general, you want to test _functionality_ and not _implementation_.

JeffAbrahamson:
LGTM.

mcmikashi:
- update the test

PR #574, opened 2022-03-24
Tbe config 401

Shriukan33:
PR based on #563 (which itself is based of #561)

[Add publish feature to TBEmail](https://github.com/transport-nantes/tn_web/commit/2a82904f07cf2a8ec9d375ed45af63b0d6baf2ca)

Because the TBItem class is very similar on this, we can use the same
methods for this.

[Adjust ViewOne template to delete unused buttons and add another](https://github.com/transport-nantes/tn_web/commit/41c4ed5bf346c07656d95628836b247f368d007e)

Retirer et Republier buttons never were implemented and are outdated in
our current design, so we are removing them.

Added a button that allow the user to see the currently published
version of a slug if any, even from drafts.
It does 404 on click if no slug is published, but it doesn't seem to be
a big issue for now.

Maybe we could have a "PublishableItems" class to set the functions there instead of importing them from TBItem. 
Although for now it doesn't seem to be too much clutter, so it's low priority. wdyt ?


JeffAbrahamson:
LGTM.
Note that this is based on a branch that's not yet approved.

JeffAbrahamson:
BTW, your commit comment doesn't really parse:
```
    Because the TBItem class is very similar on this, we can use the same
    methods for this.
```
The repetition of "this" is pretty confusing.

Shriukan33:
Rebased on add-unsub-link (which is rebased on master), and reworded the commit message which I agree was unclear !

Shriukan33:
I rebased on now master (with the unsub link change), can merge ? :)

PR #575, opened 2022-03-24
Add warning before send

Shriukan33:
based on #563 

Add a confirmation window before sending a TBEmail to a mailing list.

It informs you about how many people are subscribed to a mailing list, and makes an extra step to prevent accidental sending.

### DIFF
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -626,7 +626,7 @@ class TopicBlogEmail(TopicBlogObjectSocialBase):
         ],
     }
     template_config = {
-        'topicblog/content_email.html': {
+        'topicblog/content_email_client.html': {
             'user_template_name': 'Classic',
             'active': True,
             "fields": {
@@ -690,6 +690,18 @@ class TopicBlogEmail(TopicBlogObjectSocialBase):
             return reverse("topicblog:edit_email",
                            kwargs={"pkid": self.pk,
                                    "the_slug": self.slug})
+
+    def get_participating_field_names(self) -> set:
+        return TopicBlogItem.get_participating_field_names(self)
+
+    def get_missing_publication_field_names(self) -> set:
+        return TopicBlogItem.get_missing_publication_field_names(self)
+
+    def is_publishable(self) -> bool:
+        return TopicBlogItem.is_publishable(self)
+
+    def publish(self):
+        return TopicBlogItem.publish(self)
 
 
 class TopicBlogEmailSendRecord(models.Model):
--- a/transport_nantes/topicblog/static/topicblog/email-sending-confirm.js
+++ b/transport_nantes/topicblog/static/topicblog/email-sending-confirm.js
@@ -3,6 +3,9 @@ This script is used to display the number of recipients before
 sending a TBEmail object to a mailing list, and ask for confirmation
 */
 
+// Initial state
+document.getElementById('nb-recipients').innerHTML = "0";
+document.getElementById("send-email-btn").disabled = true;
 
 async function get_number_of_recipients(mailing_list_token) {
     let response = await $.ajax({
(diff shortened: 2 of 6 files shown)


JeffAbrahamson:
Wouldn't it make more sense to display the number of recipients from the outset rather than waiting for the user to hit the send button?  You can still have a chicken box, but providing information up front is usually a better experience.  It satisfies the condition of least surprise, for example.

JeffAbrahamson:
Otherwise, LGTM.

Note that this is based on a branch that isn't yet approved.

Shriukan33:
Added a small line under the dropdown, informing you about the number of recipients : 
image

JeffAbrahamson:
In ab3af08b, it's somewhere between weird and wrong to have function calls that are lateral in the inheritance tree.  If there is commonality like this, it should be refactored to a base class.

Indeed, you'll need this function for all four TB* subclasses, might it not be easier to write them all at once so that you see what's common all at the same time?  Sorting all of these together should be quite fast, I'd think.

JeffAbrahamson:
5a5feda4 - LGTM

JeffAbrahamson:
97983cf1 - A bit of an aside, but to be usable, this is going to have to be async.  But that doesn't matter when we have no one subscribed.

Shriukan33:
> 97983cf1 - A bit of an aside, but to be usable, this is going to have to be async. But that doesn't matter when we have no one subscribed.

Weirdly your commit shas aren't parsed by GitHub but I think I know what you're talking about, just not sure. 
What's not async here ?

JeffAbrahamson:
97983cf1 - I think the logic in `get_number_of_recipients()` could be improved.  Why wait until the use clicks send in order to provide feedback?

This is a conversation we have often.  _Before_ the user acts, make sure, if possible, that you present the user with all the information s/he needs to make an informed decision.  If you still want a chicken box, that's fine, but don't surprise the user with its contents.

So when the user chooses the list from the dropdown, that's the time to fetch and display the number of recipients.

Shriukan33:
> 97983cf1 - I think the logic in `get_number_of_recipients()` could be improved. Why wait until the use clicks send in order to provide feedback?
> 
> This is a conversation we have often. _Before_ the user acts, make sure, if possible, that you present the user with all the information s/he needs to make an informed decision. If you still want a chicken box, that's fine, but don't surprise the user with its contents.
> 
> So when the user chooses the list from the dropdown, that's the time to fetch and display the number of recipients.

This is not what's happening, did you fetch the branch ?
I made it so the number of recipients is shown when you select a mailing list, even before you click send.

JeffAbrahamson:
> > 97983cf1 - I think the logic in `get_number_of_recipients()` could be improved. Why wait until the use clicks send in order to provide feedback?
> > This is a conversation we have often. _Before_ the user acts, make sure, if possible, that you present the user with all the information s/he needs to make an informed decision. If you still want a chicken box, that's fine, but don't surprise the user with its contents.
> > So when the user chooses the list from the dropdown, that's the time to fetch and display the number of recipients.
> 
> This is not what's happening, did you fetch the branch ? I made it so the number of recipients is shown when you select a mailing list, even before you click send.

I fetched cd7c202 which became 6db92f8 when rebased to master.

I'm talking about this:

```
+// This fires only when the "Envoyer" button is clicked
+document.getElementById('send-email-btn').addEventListener('click', async function (event) {
+    // Prevent the button to POST
+    event.preventDefault();
+    // Get the selected value in the dropdown
+    let mailing_list_token = document.getElementById('id_mailing_list').value;
+    // Result of the Ajax request is a dict {"count" : <int:number_of_recipients>}
+    let number_of_recipients = await get_number_of_recipients(mailing_list_token);
+    number_of_recipients = number_of_recipients["count"];
     ...
```
which, if I'm reading correctly, puts a click handler on the send button, and when that handler is called, it asks for the number of recipients.  If that number is positive, it presents a chicken box.  If it is zero, strangely, the user will be told, only now, that she has been wasting her time.

JeffAbrahamson:
6db92f8 - listen**n**er is misspelled.

Here you do provide feedback on changing the dropdown menu, I think.  And you store it in the DOM.  So, on clicking the send button, why do you need to fetch it again instead of grabbing it from the DOM?  You'll save a few hundred milliseconds.

JeffAbrahamson:
LGTM

JeffAbrahamson:
```
+@StaffRequired
+def get_number_of_recipients(request, *args, **kwargs):
```
Why staff only?

Shriukan33:
> ```
> +@StaffRequired
> +def get_number_of_recipients(request, *args, **kwargs):
> ```
> 
> Why staff only?

Changed for `tbe.may_send` permission check !

PR #578, opened 2022-03-25
Update the launcher templatetags test

mcmikashi:
I said that i update the test but i dont push the commit on [PR 573](https://github.com/transport-nantes/tn_web/pull/573)


JeffAbrahamson:
LGTM.  Merging.

Isn't there something in the django test suite for inspecting the DOM rather than just doing string matching on the HTML blog?

Issue #559, opened 2022-03-20
Make TBItem / TBLauncher teaser template tag


JeffAbrahamson:
We want something that will let us stop pointing index to various concrete pages.

So we'd like the index page not to have an article text or anything, just  a template tag, something like this:
```
{% item_teaser 'launcher-slug' %}
```
and what that would do is render the top image and the first 50 words of the article, an ellipsis, and a "read more" button.  Note that there is a django filter tag, [truncatewords](https://www.djangotemplatetagsandfilters.com/filters/truncatewords/).

![image](https://user-images.githubusercontent.com/1220683/159179086-d3e7688f-94f9-468c-86ca-562e5271bfbf.png)

Here we're assuming, of course, that the above launcher is parameterised in the obvious way.  We provide a launcher, but what renders is a part of the underlying TBItem.

We also want (keep this in mind, but make a separate issue for it and only do it later) a second templatetag that doesn't display the image.  That way the index page can become
```
{% item_teaser 'plus-tramway' %}
<h2>Les projets Mobilitains</h2>
{% launcher 'mobilite-regionale' %}
{% launcher 'grande-mobilite' %}
{% launcher 'velopolitain' %}
{% launcher 'ligne-johanna-rolland' %}

{% item_teaser_no_image 'something-else' %}
```

Obviously these tags should share code, which is why it's worth keeping in your head early on.  Make sure it doesn't slow you down, though.  If this starts taking long (e.g., more than an hour or two), start dropping things.

These slugs are all constants, but once it's working we'll make them be dynamic, provided by some function.  Note that we are doing things one step at a time...

Issue #582, opened 2022-03-29
Add teaser_words field to TBLauncher


JeffAbrahamson:
Add an int field to `TBLauncher` called `teaser_words` that defaults to 50.

The `item_teaser` templatetag should use this int rather than always using 50.  This will let us fiddle with it when that seems the right thing to do, say the cutoff changes meaning subtly.

mcmikashi:
The` item_teaser` uses a TBItem object if we want to use teaser_words we should add this to TBItem or add a new argument to the item_teaser templatetags function teaser_words that lets us control the number of truncwords ?

JeffAbrahamson:
Aha, I missed this in the code review.  The teaser should use the TBLauncher and access the TBItem from there.  So this is a bug in the teaser template.

Recall that the eventual logic is that some function will provide an ordered list of (probably) six TBLauncher slugs, and the dynamic index page will render them as a teaser, four standard launchers, and a second teaser.

JeffAbrahamson:
LGTM.

PR #583, opened 2022-03-29
Slight improvement homepage

Shriukan33:
Edit : Failing test is the test_item_teaser that has hardcoded html, going to update this) 
Edit 2 : Solved !

This PR includes 2 improvements, one about the visual aspect of the new index, the other about flexibility of the teasing feature.

The visual improvement has been inspired by the news website I browse, and social medias : You are usually shown just enough to make you want to click, while the text fades out. 

Examples:
image
image
image

So my proposition for this :

image
image

The next improvement is about how much you reveal before fading out. 
For this I leave you to the commit message : 
>It's a proposition to make the number of words displayed adjustable
depending on what we want to display.
I think it could be used to show just enough to tease curisity from the
user, but that line depends on the item we apply the teasing to.
>
>For now it's hardcoded, but so is the item so it's still a git action to
edit this value.
However, I think we could add an optional field "teasing_length" to the
TBItem model to make it configurable in the future.

In code it translate to simply change this variable : 
image

This is part of #559

### DIFF
--- a/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
+++ b/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
@@ -435,7 +435,10 @@ span.centered {
     margin: 2px;
     padding: 4px;
 }
-
+.teaser-text {
+    -webkit-mask-image: linear-gradient(to bottom, black 50%, transparent 100%);
+    mask-image: linear-gradient(to bottom, black 50%, transparent 100%);
+}
 .higlighted-blue {
     font-weight: bold;
     background-color: var(--mobilitains-bleu);
--- a/transport_nantes/asso_tn/templates/asso_tn/index.html
+++ b/transport_nantes/asso_tn/templates/asso_tn/index.html
@@ -5,7 +5,7 @@
 {% load launcher %}
 {% load item_teaser %}
 {% block content %}
-{% item_teaser 'ligne-johanna-rolland-pour-plus-de-mobilite' %}
+{% item_teaser 'ligne-johanna-rolland-pour-plus-de-mobilite' 100 %}
 <div class="background-ad-hoc-blue">
     <div class="d-flex flex-column p-4 p-xl-5">
         <h2>Les projets Mobilitains</h2>
--- a/transport_nantes/topicblog/templates/topicblog/template_tags/item_teaser.html
+++ b/transport_nantes/topicblog/templates/topicblog/template_tags/item_teaser.html
@@ -6,8 +6,8 @@
            <p class="card-text text-white">{{ item.header_description }}</p>
          </div>
     </div>
-    <div class="p-4">
-         <p>{{ item.body_text_1_md|truncatewords:50 }}</p>
+    <div class="p-4 d-flex flex-column">
+         <p class='teaser-text'>{{ item.body_text_1_md|truncatewords:50 }}</p>
 		{% comment %}
 	    If the launcher slug does not reference a launcher in the
 	    database, the templatetag will pass None to this template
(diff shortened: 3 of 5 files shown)


JeffAbrahamson:
e15cc63 - 

* Why is issue #582 not tagged?
* Where does `truncate_after` come from?  I don't see a model change.  Are you expecting the number to be fixed for a template, regardless of what TBItem is being displayed, but maybe variable by template?
* Question answered, maybe, in wondering why you explicitly set the value in `asso_tn/index.html` to 100.  You do know from your exchanges with Mickael that the slug is a variable, which we are setting to be constant in a first iteration?

@mcmikashi - I think that @Shriukan33 didn't grok your design intent and our discussions on the `item_teaser` and `launcher` tags.  This is a direct comment on your git and code documentation and a reminder why documentation is important.  Code must communicate.  It is the programmers equivalent of the Hippocratic Oath: first, explain what you're doing, because no one will understand later.

JeffAbrahamson:
da3c574 - We really need to stop doing hard-coded full text compares as tests.  The result is that if we change the least bit of formatting in a template, the test will fail.  But we never meant to test the minutia of formatting (or, if we do, we should explicitly test that and nothing else in that particular test).

The right way to do these tests is to write the python that renders the view, then query the DOM returned to make sure that what you want to be there is there.

This is a general issue I've started noticing in our tests, and it makes them fragile.

Shriukan33:
> [e15cc63](https://github.com/transport-nantes/tn_web/commit/e15cc634ca30c7ce648fb204d85a64d5dc8f5009) -
> 
>     * Why is issue [Add teaser_words field to TBLauncher #582](https://github.com/transport-nantes/tn_web/issues/582) not tagged?

Wasn't aware of this particular issue, will reword commit message to tag it.

>     * Where does `truncate_after` come from?  I don't see a model change.  Are you expecting the number to be fixed for a template, regardless of what TBItem is being displayed, but maybe variable by template?

It's simply hardcoded into the templatetag, you can either precise the length (in my case, I used 100 but could be anything) or leave it blank (default is 50)
So yes, it's not dependent on the slug you're inputting, and that's why I suggested we could add a field in the model to make it per-item appropriate here : 
```
For now it's hardcoded, but so is the item so it's still a git action to
edit this value.
However, I think we could add an optional field "teasing_length" to the
TBItem model to make it configurable in the future.
```

>     * Question answered, maybe, in wondering why you explicitly set the value in `asso_tn/index.html` to 100.  You do know from your exchanges with Mickael that the slug is a variable, which we are setting to be constant in a first iteration?
> 
> 
> @mcmikashi - I think that @Shriukan33 didn't grok your design intent and our discussions on the `item_teaser` and `launcher` tags. This is a direct comment on your git and code documentation and a reminder why documentation is important. Code must communicate. It is the programmers equivalent of the Hippocratic Oath: first, explain what you're doing, because no one will understand later.

I'm very aware that this is a temporary situation and that eventually we will not hard code the slug into the template for a more db-based approach.
I've added this little improvement in the meantime, and suggested the field idea for the time where this will be db-driven. 

Unless I'm missing something ?

Shriukan33:
> [da3c574](https://github.com/transport-nantes/tn_web/commit/da3c574d0c2ed2fa61e0e5c346afbafb9fa5ca18) - We really need to stop doing hard-coded full text compares as tests. The result is that if we change the least bit of formatting in a template, the test will fail. But we never meant to test the minutia of formatting (or, if we do, we should explicitly test that and nothing else in that particular test).
> 
> The right way to do these tests is to write the python that renders the view, then query the DOM returned to make sure that what you want to be there is there.
> 
> This is a general issue I've started noticing in our tests, and it makes them fragile.

I agree this is very rigid. I simply have no idea about how to test that a templatetag works without checking it against its output with `assertIn()` ...

JeffAbrahamson:
In light of @mcmikashi's PR #585 on #582 perhaps you could create a fade-out issue and rework this PR just to do that?

JeffAbrahamson:
I'm going to close this and let you create a new PR with just the fade-out.

PR #584, opened 2022-03-29
Improve TBLauncher comments

JeffAbrahamson:
They were gibberish, copied from TBPress.  Somehow no one noticed. :-(


Shriukan33 (approved):
LGTM
Comments haven't been cleaned yet, the whole pedagogical docstring could be erased at this point, as we already transposed that into issues :slightly_smiling_face:

PR #585, opened 2022-03-29
item_teaser

mcmikashi:
- Fix item_teaser templatags now use TBlauncher to get the data
- Add teaser_words on TBLauncher Model


transport_nantes/topicblog/templatetags/launcher.py
```diff
@@ -13,3 +13,15 @@ def launcher(slug):
     if launcher is None:
         logger.error(f"item_teaser failed to find slug: \"{slug}\".")
     return {'launcher': launcher}
+
+@register.inclusion_tag('topicblog/template_tags/item_teaser.html')
+def item_teaser(slug):
+    launcher = TopicBlogLauncher.objects.filter(
+        slug__iexact=slug, publication_date__isnull=False).first()
+    if launcher is None:
+        logger.error(f"item_teaser failed to find slug: \"{slug}\".")
```
JeffAbrahamson, line 22: Good to log the error, but then falling through to the next line will fail violently, no?

JeffAbrahamson:
Putting the two templatetags in a single file is good, saves an extra file load, thanks!

JeffAbrahamson:
Could you please tag the issue?

JeffAbrahamson:
You still have a bug in your first commit:
```
+def item_teaser(slug):
+    launcher = TopicBlogLauncher.objects.filter(
+        slug__iexact=slug, publication_date__isnull=False).first()
+    if launcher is None:
+        logger.error(f"item_teaser failed to find slug: \"{slug}\".")
+    item = TopicBlogItem.objects.filter(
+        slug__iexact=launcher.article_slug, publication_date__isnull=False).first()
+    if item is None:
+        logger.error(f"item_teaser failed to find slug: \"{slug}\".")
+    return {'launcher': launcher, 'item': item}
```
If `launcher is None` you will log an error, but then you will immediately dereference `None.article_slug` in the next line, which will throw.

Did you test this?

JeffAbrahamson:
Instead of writing
```
+      {% if launcher.teaser_words %}
+         <p>{{ item.body_text_1_md|truncatewords:launcher.teaser_words }}</p>
+        {% else %}
+        <p>{{ item.body_text_1_md|truncatewords:50 }}</p>
+      {%endif %}
```
why don't you use the `default` filter?

mcmikashi:
> why don't you use the `default` [filter](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#default)?

I try with default and default_if_none but both doesn't work

mcmikashi:
- Now both commit pass the test

Shriukan33:
I suggest this for the truncatewords : 
```jinja2
      {% with launcher.teaser_words|default:50 as teaser_words %}
        <p>{{ item.body_text_1_md|truncatewords:teaser_words }}</p>
      {% endwith %}
```

Also the end of the file is hard to read without proper indent, to my taste : 
```jinja2
          {% if item.slug %}
            <a href="{% url 'topic_blog:view_item_by_slug' item.slug %}" class="btn donation-button btn-lg" >en lire plus <i class="fa fa-arrow-right"></i></a>
          {% else %}
            404
          {% endif %}
```
Instead of 

```jinja2
 {% endcomment %} {% if item.slug %}
               <a href="{% url 'topic_blog:view_item_by_slug' item.slug %}" class="btn donation-button btn-lg" >en lire plus <i class="fa fa-arrow-right"></i></a>{% else %}404{%endif %}
    ```

**Comment by Shriukan33** [comment]
> > I suggest this for the truncatewords :
> > ```
> >       {% with launcher.teaser_words|default:50 as teaser_words %}
> >         <p>{{ item.body_text_1_md|truncatewords:teaser_words }}</p>
> >       {% endwith %}
> > ```
> 
> What happens if you write
> 
> ```
> {{ item.body_text_1_md|truncatewords:(teaser_words|default:50) }}
> ```
> 
> ?
> 
> I can't currently get that code to run due to the exception in creating launchers.

It throws an error. 

But for the sake of it, I tried this : 

```jinja2
{{ item.body_text_1_md|truncatewords:launcher.teaser_words|default:50 }}
```
And it works perfectly fine. I wonder what went wrong with @mcmikashi

mcmikashi:
Test fail when i try this : 
new_fail-31-03-2022
new_fail-n2-31-03-2022

mcmikashi:
- indent the end of item_teaser

JeffAbrahamson:
LGTM, kinda.  I'm merging, having made the following changes:
* Modified the projects line html to  bring it into line with the current html, most notably requesting 4 columns or 1.  This may be revisited separately, but it is a separate design decision and so should be thought through and discussed as such.
* Made the teaser_words value non-nullable in the model, which causes old launcher instances to be assigned the value 50.  I didn't remove the default value in the template, but I think we could do that now that I'm writing this.
* Actually generated and committed the necessary migration.
* Tagged the issue in the commit comments.  Slightly reworded the commit comments to be more understandable.

PR #587, opened 2022-03-29
Add open graph on PressMention

mcmikashi:


JeffAbrahamson:
LGTM

JeffAbrahamson:
The failing test appears to be because someone introduced a test that connects to journal.com.

JeffAbrahamson:
Why is it appropriate to use a receiver instead of simply calling a function from the view's `form_valid()` method?

What will happen if the fetch fails in the receiver?  Is the model saved without the additional data?  What is the user experience?

JeffAbrahamson:
Perhaps you should comment why you have set the `og_*` fields to blank.

JeffAbrahamson:
What sets `created` in `add_og_twitter_data`?

Given that it is provided as a function argument, how could the first condition be handled differently than the second?

JeffAbrahamson:
It appears that you save the object three times:
1. in `ModelFormMixin`'s `form_valid()` method
2. at the end of each of the `if created` blocks

JeffAbrahamson:
In `add_og_twitter_data()` you appear to repeat twice the following block of code:
```
    og_title = tree.xpath('//meta[@property="og:title"]/@content')
    og_description = tree.xpath('//meta[@property="og:description"]/@content')
    og_image = tree.xpath('//meta[@property="og:image"]/@content')
    if created:
        instance.og_title=og_title[0]
        instance.og_description=og_description[0]
        instance.og_image=og_image[0]
```

Did you proofread before committing?

JeffAbrahamson:
You are missing spaces around "=" signs when you set `instance.og_*` in the receiver.

Do you have a linter?

JeffAbrahamson:
You are storing the image URL but not the image.

JeffAbrahamson:
It's still not reasonable for tests to fail because of a distant failure on the internet.  What, precisely, are you testing that requires you to depend on the availability and services provided by a remote host?

That is, what is the invariant whose veracity you are asserting in this test?

JeffAbrahamson:
Oh, I see, you want to verify that the open graph fetches work.  Google around for how to do this, but I suspect that the right way is to open a local port 80 listener that serves the data you want.

JeffAbrahamson:
Note that this is wrong:
```
+        if resp.status_code != requests.codes.ok:
+            logger.error(f"The url of the open graph image is not good.")
```
It could be that the URL is just temporarily unavailable.  Cf. #619.

JeffAbrahamson:
Otherwise, LGTM.  Merging.

PR #608, opened 2022-03-30
Improve list signup with placeholder and text size

mcmikashi:
- title from h2 to h4
- remove the label and add placeholder


JeffAbrahamson:
HTML header tags communicate semantic information about the document.  They aren't the right way to fiddle with size information.

There's perhaps a question whether list signup should be h2 or h3 or not a header at all, but just changing the semantic structure of documents should be motivated by structure, not by font size.

JeffAbrahamson:
h4->h3 edited
merged

PR #611, opened 2022-03-31
Fix navbar is overlapped by header images

mcmikashi:
- Remove the mt-n3 class

### DIFF
--- a/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
+++ b/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
@@ -210,7 +210,7 @@
 				{% endif %}
 			{% endblock navbar %}
 			{% block content %}
-				<div class="hero-image mt-n3">
+				<div class="hero-image">
 					<div class="hero-text">
 						<h1 style="color: white;" class="hero-title">{{ page.hero_title }}</h1>
 						<p style="color: white;" class="hero-description">{{ page.hero_description }}</p>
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
I'm confused why this is the right thing to do, and you don't explain how you concluded that this was correct and not, say, just some random thing you tried that happened to work at the SHA1 and browser and such where you tried it.  This particular bit of CSS has been present for quite some time and we haven't had problems.  Moreover, the overlap phenomenon isn't seen everywhere.

I'm pretty sure the point of the `mt-n3` is to remove an unsightly white line between the header image and the navbar.

I'm not saying this isn't the right solution, but I think it needs to be justified: problem reproduced on the following browsers, solution tested on same, etc., plus some hypothesis why this only affects us now.

Shriukan33:
I tried to solve it in my own branch and came to the same conclusion it had to do with mt-n3, I explained in my PR #620 the origin of it and why this fixes it. 
We could either : 
- re-enable the empty `ul` tag under the navbar this would revert commit 70d41a8262a67edb5437e49e9098d1e67f84e614 change, and therefore, we would be back on initial state) 
- Clean the code from its empty / commented out parts and layout (my solution) 

I think it doesn't make so much sense to have an empty div holding and empty ul tag that only contains long-ago-commented code. 
Moreover it made the layout deal with negative margins to deal with white gap, and I don't think it's especially good practice (although I can't think of rules stating you shouldn't use negative margins to cover empty elements ... )

JeffAbrahamson:
Part of the confusion is that the inclusion of this delta in 70d41a8 made it appear at first glance that it was needed there, but it is looking like it was just an independent bit of clean-up.

This is a good reminder to do clean-up separately from other changes and to comment (explain) why one is doing it.  It makes resolving later bugs clearer.

Shriukan33:
I propose we continue this issue with #620 , closing this one for now

PR #612, opened 2022-03-31
Fix CTA slug on content_email_client template

mcmikashi:
Fix the problem, but maybe there's a cleaner way to do it.


JeffAbrahamson:
You don't explain in your commit what problem you are solving nor why this is the right solution.

--- 

This is a clever approach, but I fear it's not the right way to do it.

A red flag for me is that our template tags generally take slugs, but here you're using the view to provide the templatetag with an alternate syntax.  This seems like technical debt introduced as such, since it requires humans who use our template tags to remember which ones use different syntax.

I'd rather we explore this more and try to find a better solution.

PR #613, opened 2022-04-01
Fix hardcoded s

JeffAbrahamson:
We are temporarily hard-coding the slugs to use on the index page.
Use slugs slugs that actually exist in production and that we
currently expect on the index page so that we can test this and make
sure we're happy with the result.


Shriukan33 (approved):
LGTM

PR #614, opened 2022-04-01
Refactor from TopicBlogItem to TopicBlogObjectSocialBase

mcmikashi:
* Move is_publishable from TBItem to TopicBlogObjectSocialBase
* Move publish from TBItem to TopicBlogObjectSocialBase
* Move get_missing_publication_field_names from TBItem to TopicBlogObjectSocialBase
* Move get_participating_field_names from TBItem to TopicBlogObjectSocialBase
* Update the field list on get_participating_field_names

### DIFF
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -237,7 +237,134 @@ class TopicBlogObjectSocialBase(TopicBlogObjectBase):
 
         context['social'] = social
         return context
+    
+    def is_publishable(self) -> bool:
+        """
+        Return True if the object may be published.
 
+        An object may be published if it has no missing required fields,
+        as defined in self.template_config
+        """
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
This is almost right, but a couple points that would be good to improve, I think:

* In your commit comment, mention why you're doing any of this.
* `get_participating_field_names()` includes field names that aren't part of the base class -- it should just handle the base class, then let subclasses override when necessary for additional fields.

mcmikashi:
- Update the commit message
- Update get_participating_field_names() is override on all subclass
- Add PEP-8 compliance

JeffAbrahamson:
LGTM, merging after rewording the commit comment.

(There's no point of saying in a commit comment what is absolutely obvious from the code diff: things like "moved function from here to there".  What is useful is the context, why you did this at all.)

PR #615, opened 2022-04-01
Fix image on content_email template

mcmikashi:
- fix the name of the field image
- add some indent


JeffAbrahamson:
Your commit comment doesn't say what you're fixing besides that it's something image content.  You're having a conversation with the future.  You get one shot in the present to get it right.  Don't waste your shot.

Also, you say you're fixing images, but your commit also changes buttons.  This is only spacing, but perhaps should be a separate commit to avoid confusion.

That said, was the length tag failing or was this merely clean-up?  (I agree that it's useful clean-up, but your commit message doesn't address this.)

Shriukan33:
LGTM

JeffAbrahamson:
LGTM, merging.

PR #620, opened 2022-04-04
Fix image banners overlapping the Navbar

### LINKED ISSUE #599: Navbar is overlapped by header images
Example from beta.mobilitains.fr : 
image

Shriukan33:
In commit 70d41a8262a67edb5437e49e9098d1e67f84e614, the ul tag was
removed from a div separating the navbar from the content.
The UL tag inside it forced a small gap that was then covered by a
negative top margin ("mt-n3"). The removal of this tag made the negative
margin overlap the navbar, because it didn't have the ul space to cover.

Instead of messing with the CSS, I removed the useless ul tag alongside
its parent div (that was also useless because it had no properties).
Removing this div also removes the need to have a negative margin that
covers the gap left by the ul tag.

This negative margin was present in base_mobilitain.html but also in
most base templates, because they came from base_mobilitain.html.

Closes #599

### DIFF
--- a/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
+++ b/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
@@ -180,26 +180,9 @@
 					</div>
 				</nav>
 
-				<div>
-					{% comment %}
-					<ul>
-								<li class="nav-item dropdown">
-								<a class="nav-link dropdown-toggle" href="#" data-toggle="dropdown"><i class="fa fa-user" aria-hidden="true"></i></a>
-								<div class="dropdown-menu">
-								{% if user.is_authenticated %}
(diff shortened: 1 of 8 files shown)


Shriukan33:
Converting to draft while I fix a failing test that has to do with the layout.

Shriukan33:
Fixed ! Used more stable reference instead of xpath (Used css selector).  These tests shouldn't break if we change the layout now.

JeffAbrahamson:
> Fixed ! Used more stable reference instead of xpath (Used css selector). These tests shouldn't break if we change the layout now.

That's good.  We generally don't mean to be testing the invariance of spaces in our HTML, after all.

Shriukan33:
Is that a LGTM ? :)

JeffAbrahamson:
Excellent commit message, it very nicely explains why you did what you did and encapsulates the research you had to do to get there.

JeffAbrahamson:
LGTM.

JeffAbrahamson:
I'm sufficiently confident you mean for this to be merged, I'm going to do it here just to clean up.

PR #622, opened 2022-04-04
Fix CTA links in mail-client rendered TBEmails

### LINKED ISSUE #596: Fix CTA links inside template for Email clients
CTA links are broken in Email clients because they don't have the host name inside them, only the /url/here part.

Shriukan33:
Links used to not have scheme and host and would look like this:
`/tb/t/<the_slug>/`
This doesn't parse outside of our website.
Adding the scheme and host allows us to have complete links that are
understandable outside of our website's context e.g.:
`https://www.mobilitains.fr/tb/t/<the_slug>/`

Closes #596

tested on beta, works fine !

### DIFF
--- a/transport_nantes/topicblog/templatetags/email_tags.py
+++ b/transport_nantes/topicblog/templatetags/email_tags.py
@@ -40,14 +40,20 @@ def email_body_text_md(text: str) -> str:
     return mark_safe(html_template)
 
 
-@register.simple_tag
-def email_cta_button(slug: str, label: str) -> str:
+@register.simple_tag(takes_context=True)
+def email_cta_button(context: dict, slug: str, label: str) -> str:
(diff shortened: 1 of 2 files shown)


Shriukan33:
> Rather than explicitly passing context, would it not be better to register the templatetag with `takes_context=True`?
> 
> Cf. `topicblog/templatetags/markdown.py`
> 
> ```
> @register.simple_tag(takes_context=True, name="tn_markdown")
> def tn_markdown(context, value):
>     parser = TNLinkParser(context, verbose=False)
>     return mark_safe(markdown(parser.transform(escape(value))))
> ```

That works well, did it with takes_context and updated the branch + rebased

JeffAbrahamson:
LGTM.

PR #624, opened 2022-04-04
Fix user_subscribe_count doesn't return a correct number

mcmikashi:
- Fix the user_subscribe_count
- Update the user_subscribe_count test

### DIFF
--- a/transport_nantes/mailing_list/events.py
+++ b/transport_nantes/mailing_list/events.py
@@ -1,5 +1,5 @@
 from django.contrib.auth.models import User
-from django.db.models import Max
+from django.db.models import Max, OuterRef, Subquery
 from django.db.models import Q
 
 from .models import MailingList, MailingListEvent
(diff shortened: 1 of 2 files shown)


transport_nantes/mailing_list/events.py
```diff
@@ -55,14 +55,13 @@ def user_subscribe_count(mailing_list):
...
-    users_subscribed = user_states.filter(
-        Q(event_type=MailingListEvent.EventType.SUBSCRIBE)).count()
-    return users_subscribed
+    user_list = MailingListEvent.objects.filter(mailing_list=mailing_list).values_list('user',flat=True).distinct()
+    users_subscribed = list()
+    for user in user_list:
+        user_state = user_current_state(User.objects.get(pk=user),mailing_list)
+        if user_state.event_type == MailingListEvent.EventType.SUBSCRIBE:
+           users_subscribed.append(user_state)
+    return len(users_subscribed)
```
Shriukan33, line 64: Could have been a mere counter (total += 1), but that works too !

Shriukan33 (approved):
LGTM

JeffAbrahamson:
Think at scale.  What happens when fetching the list of potentially subscribed users from the database server takes a second?  Or ten seconds?

Can't you do the count on the server and only transfer the number you need?

Your code looks like it should work, but it is highly inefficient.

Shriukan33:
Asked the question on StackOverflow : [here](https://stackoverflow.com/questions/71763363/django-orm-query-to-get-number-of-currently-subbed-user)
The guy who answered proposed this : 
```python
User.objects.alias(
    latest_action=Subquery(
        MailingListEvent.objects.filter(
            mailing_list=mailing_list,
            user=OuterRef('pk')
        ).order_by('-event_timestamp').values('event_type')[:1]
    )
).filter(latest_action='sub').count()
```
From what I understand, we get a Queryset that queries Users and decorates the output with "latest_event", which will be either "sub", "unsub" or "bounce". It then checks the value of this decoration and and filters only the subs. 

As the comment specifies, it means that if the latest event isn't "sub", it wont be counted as a subbed user. However the next function that sends email to out mailing list will get the correct list anyway. 
Implementing this solution might work and be close to reality (because bounces hopefully shouldn't be that common to be found, in latest on top of that), and is Database-friendly. I think it could be good enough for its use (warning user about number of subscribed users). 
On the down side, it's not 100% accurate.

WDYT @JeffAbrahamson ?

mcmikashi:
- Update with the StackOverflow answer
- Clean the import

JeffAbrahamson:
LGTM.

PR #625, opened 2022-04-04
Fix the launcher_image when creating a launcher

mcmikashi:
- Fix the launcher_image when creating a launcher
- Remove headers banner of content_launcher template

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/content_launcher.html
+++ b/transport_nantes/topicblog/templates/topicblog/content_launcher.html
@@ -38,13 +38,6 @@
 {% endblock styles %}
 
 {% block content %}
-<!-- Banner with the image above the written content -->
-<div class="hero-image">
-     <div class="hero-text">
-          <h1 class="hero-title">{{ page.header_title }}</h1>
-          <p class="hero-description">{{ page.header_description }}</p>
-     </div>
-</div>
 <div class="container-fluid">
     <p>Cette page n'est pas privée mais n'est pas destinée au public.
 	Elle peut disparaître du jour au lendemain.
--- a/transport_nantes/topicblog/templates/topicblog/content_launcher.html
+++ b/transport_nantes/topicblog/templates/topicblog/content_launcher.html
@@ -61,12 +61,18 @@
         {% if page.article_slug %}
             <a href="{% url 'topic_blog:view_item_by_slug' page.article_slug %}" class="text-decoration-none">
         {% endif %}
-		<div class="shadow-md">
-		    <img class="rounded"
-			 src="{{ page.launcher_image.url }}"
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
Is this still pertinent in light of ace334c?

JeffAbrahamson:
Launchers are weird, because they are indirections.  They need a concept of active/inactive, which is akin to published, but not quite.  And I see that's causing you guys confusion.

But think through the workflow.

1.  It's completely imaginable that someone starts to make a launcher, realises he's got everything but the image, and saves at that point.  So the image can't be mandatory.
2. It's also imaginable that someone makes a launcher but isn't permissioned to make it active.  So we need a concept of "not active".
3. It's also imaginable that we make a launcher for a campaign that hasn't started yet.  It has to be ready, but it shouldn't be served.  So it's inactive.
4. We might also turn off a campaign for a while and reactivate it later.  Or never.

So this is slightly different than publish, because there's a clear idea of "un(publish)".  We wouldn't do that with a TB{Item|Mail|Press}, because once we publish, we have a sort of public commitment not to 404.  But with launchers, there's no idea of 404.

So there should be a button called "activate" which is pretty much the same as publish (and might be the `publish()` function under the bonnet).

And then there's a button (not a priority for deploy to prod) called "deactivate" that marks it as not active.

We can probably share many fields with `TopicBlogObjectBase` here, since those base publication fields still make sense.  And in that sense you've got most of what you need already written, it's just the human interface needs to hook into it.

It's just that there's also an "active" boolean (to add to the model) that also gets toggled by the activate/deactivate button.

Does this help?

JeffAbrahamson:
Remember, the ultimate goal is that there can be more than 4 or 6 launchers active at once.  We'll alternate what we serve, hopefully based on CTR and such (cf. [Thompson sampling](https://en.wikipedia.org/wiki/Thompson_sampling) if you like statistics).

Shriukan33:
> Remember, the ultimate goal is that there can be more than 4 or 6 launchers active at once. We'll alternate what we serve, hopefully based on CTR and such (cf. [Thompson sampling](https://en.wikipedia.org/wiki/Thompson_sampling) if you like statistics).

Actually understand better now, I didn't imagine we would have more launchers active than the ones on the homepage. 

And OK no need to put this mandatory, this PR is about making it optional.

JeffAbrahamson:
> Actually understand better now, I didn't imagine we would have more launchers active than the ones on the homepage.

We're designing for the future, not just for what we need tomorrow.  We could have actual activity, though we've been failing at that for a bit.

Though it would be good to have what we need for tomorrow, too. ;-)

JeffAbrahamson:
LGTM.

PR #626, opened 2022-04-05
Benja 598 make press sendable

Shriukan33:
This branch makes TopicBlogPress items sendable through email using their own template, in the same way TBEmails currently do. 

Beta test is working fine !


JeffAbrahamson:
LGTM.

PR #629, opened 2022-04-06
Add fade-out effect on new index and center button

Shriukan33:
Draft because failing test

The fade-out suggests that there is more to the story, and I think
encourage the user to read want to read more.
I've no studies to support that assumption, however this is a very
common pattern in news websites.

The centered button feels more natural to me.

This visual improvement has been inspired by the news website I browse, and social medias : You are usually shown just enough to make you want to click, while the text fades out. 

Examples:
image
image
image

So my proposition for this :

image
image


JeffAbrahamson:
LGTM

PR #631, opened 2022-04-06
Add a link to press:new_item in press:list_items

mcmikashi:


Shriukan33:
LGTM

JeffAbrahamson:
LGTM

I can't help thinking that this could have been done for all TB* with something in the TB base template...

Shriukan33:
> LGTM
> 
> I can't help thinking that this could have been done for all TB* with something in the TB base template...

Probably,  the link would have to be a variable then

JeffAbrahamson:
> > LGTM
> > I can't help thinking that this could have been done for all TB* with something in the TB base template...
> 
> Probably, the link would have to be a variable then

We're developers, we can do variables.

PR #646, opened 2022-04-06
Improve pressmention saving

mcmikashi:
Based on [PR](https://github.com/transport-nantes/tn_web/pull/631)
-  Saving a valid  press_mention now redirect to press:detail_item
- Add test for press:detail_item page
- Clean unused import

On the detail view, I left 2 display options

### DIFF
--- a/transport_nantes/press/models.py
+++ b/transport_nantes/press/models.py
@@ -1,4 +1,6 @@
 from django.db import models
+from django.urls import reverse
+from urllib.parse import urlparse
 
 
 class PressMention(models.Model):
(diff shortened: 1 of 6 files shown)


JeffAbrahamson:
LGTM.

Note (and please address today if possible) that if the URL is not fetchable, you get an error.

What do you think is the right way to handle this?
* One way would include scheduling (using celery, so next week) a retry.
* We should surely tell the user that we couldn't do it, because maybe it was a typo.  But maybe the site is just down for a bit.
* We should probably have an active flag that is true unless something like this happens, and then we set it to inactive so we don't try to display.

image

PR #647, opened 2022-04-07
Add suspension marks to Envoyer button

### LINKED ISSUE #636: Improve Envoyer button
In the admin banner to send the TBEmails and TBPress, the "Envoyer" button isn't clear enough that it doesn't intend to send it right away, let's add suspension marks to say "there is more if you click". -> "Envoyer ..."

Shriukan33:
To suggest there is more after the press of this button, we add
suspension marks ...

Closes #636

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/base_email.html
+++ b/transport_nantes/topicblog/templates/topicblog/base_email.html
@@ -57,7 +57,7 @@ Sendable and not topicblog_admin, so viewing published page, so can
 offer to send.
 {% endcomment %}<div class="container-fluid pl-5 py-3">
     <div class="row">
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
LGTM

[Ellipsis](https://en.wikipedia.org/wiki/Ellipsis), in English.
Also know as [Auslassungspunkte](https://de.wikipedia.org/wiki/Auslassungspunkte).  Or, in classic borrowing from abroad style, [リーダー](https://ja.wikipedia.org/wiki/%E3%83%AA%E3%83%BC%E3%83%80%E3%83%BC_(%E8%A8%98%E5%8F%B7)).

Shriukan33:
Thanks for the trivia !

Issue #635, opened 2022-04-06
Improve unsubscription page


Shriukan33:
The text feels loud and robotic.
We would like a simpler message "Désinscrire <mail> de <liste> ?" with a single "oui" button.
under it, an item_teaser that points to our "qui sommes nous" page.

PR #649, opened 2022-04-07
Reword and expand the unsubscription page (benja-635-...)

Shriukan33:
Reword and expand the unsubscription page

- Add the email for clarity and confirmation that this is the email
the user wants to unsub.

- Add the qui-sommes-nous's item_teaser below the confirmation, to
try to convince one to stay subscribed.

Minor adjustement to item_teaser to make it keep the newlines :
The markdown would be either not toggled or would be displayed
unproperly because `truncatewords` would remove the newlines.

Closes https://github.com/transport-nantes/tn_web/issues/635

image


JeffAbrahamson:
LGTM.

Feedback:
* Your commit comments could explain why (rename, update).
* We're still going to need to modify markdown at some point, because, whether it's a character count or a word count, the boundary could fall in the middle of a `[[...` and markdown would emit that verbatim because it's not well formed markdown.

But we can leave that for now, the ints chosen will avoid that happening for the moment.

PR #650, opened 2022-04-07
Improve responsive on Press mentions admin

### LINKED ISSUE #640: Improve responsive on Press mentions admin
On smaller screen the table we use at the moment is too big to fit, and not very good looking anyway.

On those smaller screen we could move out of the table and use a simplified version that shows like this :
* Name of journal in bold
* Title of article (not bold)
* A very small description that is truncated, the rest is displayed in a hover tooltip
* Delete / edit buttons are hypertext

mcmikashi:
Change the display on small and extras small screen

Close #640


mcmikashi:
This is not the same PR .

JeffAbrahamson:
> This is not the same PR .

Good point...

JeffAbrahamson:
In that case, LGTM.

PR #652, opened 2022-04-07
Benja 641 add arrow and plus fa + QoL improvements

### LINKED ISSUE #641: Add FA + and Left arrow to TBList Views
On top of a list of TB objects through their List View, we should have a + that means we can add a new object of the same nature we're watching.
On a ListOne, we add a left arrow to get back to the complete list.

Shriukan33:
This PR brings two things to TopicBlog's List view :
* A refactor
* Quality of life improvements :

## Why a refactor:
All the list views have the same implementation, with as the only
difference the urls it uses in the template. This led to the creation
of multiple templates that were extremely similar and any change had to
be duplicated to every template.

## How did I refactor:
I created a base template that is used by all the list views.
To populate it with the correct urls, I added to the model the
url name that is set inside urls.py, and accessed these properties
inside the base list view.

## About the improvement :
Added a Font-Awesome + sign on every list template as well as an arrow
to get back either to the general list or to the Dashboard.
This allows a smoother navigation into the CMS by providing shortcuts
to the various parts of the dashboard.

In a second commit I propose a 'tab' based way to navigate the TB CMS that short circuits the Dashboard screen :
image

With this structure, you can jump from a type of content (Item, Email, launcher, Press) to another with only one click. Clicking a tab brings you to the list view of the sort of item you clicked and you're presented the list + a button to add another item of the very  same nature.

Closes #641


JeffAbrahamson:
Pretty awesome, well done.

LGTM.

PR #665, opened 2022-04-08
Add Open Graph data of PressMention in table_list template

mcmikashi:


JeffAbrahamson:
This is hard to check because both `/presse/` and `/presse/list/` 500 for me locally because of a missing og image.

It's possible to have records without og images -- they might not be available when we first process the record, the remote site might not even have og data (!), although that's not particularly pro.  But it shouldn't cause us to error.

mcmikashi:
- Allows no open graph image

JeffAbrahamson:
LGTM.

PR #669, opened 2022-04-08
Improve add press_mention

mcmikashi:
If the website of the article_link doesn't have open graph data we're warning the user that the site doesn't have open graph data.


JeffAbrahamson:
I'm not sure this is the right thing to do: if the page doesn't have og data (and it suffices that one of title, description, or image is mising), we log an error and consider the form invalid.  But this is data over which we have no control, we should surely do something intelligent in the absence of having the og data.

JeffAbrahamson:
In the interests of clean-up on @mcmikashi moving on (at least for a while), I'm going to merge this, but we probably should clean this up.

By we, I of course mean @Shriukan33 ;-)

PR #671, opened 2022-04-11
Replace chicken box by a checkbox on email send form

### LINKED ISSUE #638: Replace TB Email/Press Send's chicken box with a checkbox.
Instead of using a pop-up to ask for confirmation, let's use a checkbox that is mandatory that says "Je suis sûr(e)" before sending to a mailing list.

Shriukan33:
This PR contains 2 things : 
* The replacement itself in the form and in Javascript
* The refactor of this

In the same vein of TopicBlog List views, we created two templates
that have as the only difference a url.
Instead of maintaining yet another topicblog template, I passed the
reverse argument into the context.

Closes #638


JeffAbrahamson:
LGTM.

PR #672, opened 2022-04-12
Benja 644 redirect upon subscription ml

Shriukan33:
This PR implements a redirection to a designated article upon subscription to a mailing list.

In the absence of article, the generic thank you page is displayed instead.

How to : 
Select an existing item to link to your mailing list
image
Subscribe to a mailing list : 
image
You are redirected ! 
image
image


JeffAbrahamson:
LGTM, except for the last commit (fa9cd53).
* Why are we shouting?  A good servant blends into the background leaving no desire unfulfilled.  Here, we just need a short phrase indicating that the task is done: "Noté." or "Fait." for example.  In a font size no bigger than the normal `<p>` font, left-justified.  You see it if you want, but you stop seeing it as soon as you lose interest.
* Meanwhile, the article is present -- of course you see some something now.  But we don't infantilise the visitor by suggesting that they need something to read whilst they await their first newsletter.
* Depending on mailinglist type, this may not be a newsletter at all.  It might be a petition, it might be a list for press releases, etc.  But there's no need to talk about that, it is whatever the visitor thinks it is.
* Stick with our chart colours please. ;-)  In the absence of having a colour choice, just display the same as the main text colour.  In any event, these sorts of informational messages should be styled by CSS so that they are all the same.

JeffAbrahamson:
I'm talking about this (only) when I talk about shouting:
image

I know that it's alert green, I saw that in the code.  It's  not in our charte graphique.  We should stick to that.

Greenpeace is a good example.  Their alerts correspond to their visual style guide.

JeffAbrahamson:
BTW, in the commit comment on decb666, you should write out "foreign key".  In informal conversation, one writes FK, but it's a bit too casual for a commit comment where context is missing.  It really obliges the reader to look at the diff to be sure of what you mean.

Shriukan33:
> LGTM.
> 
> Don't buttons signal interaction and not informational alert?

Can change that but it's unclear to me what button should have been used, I figured I'd use the one the user would be the most familiar with but it's no definitive choice

JeffAbrahamson:
LGTM.
Pending tests...

Issue #637, opened 2022-04-06
Create TBMailingListPitch


Shriukan33:
These items will help us create different pages to incite people subscribe to our mailing lists. This way, we will be able to have different arguments to why you should subscribe.
It takes as variable a body text, a CTA button, and a mailing_list_token.

We create those pages through form that is very similar to the one we use for TopicBlogItem, except it has less fields.

Shriukan33:
Merged with https://github.com/transport-nantes/tn_web/pull/681

PR #681, opened 2022-04-13
Benja 637 create topic blog mailing list pitch

Shriukan33:
This PR brings two things : 

1) An implementation of TopicBlogMailingList as we discussed in #637 
2) A refactor of Edit views for TopicBlog.

There are 6 commits but really only 2 can be considered big : The one implementing the views (0642668bdc4dee5170241fc9ea1203426d2648e5), and the refactor (00a3cbcfda6f026b140bef2cabfdaabf6f29f124).

The views that are implemented are very similar to TBItem's : View, ViewOne, List, ListOne, Edit.
In this branch you can : 
* Create a TBMLP object
* Publish it
* See their list and list one in the same way we do for the other TB


JeffAbrahamson:
The commit message in 00a3cbcf would be clearer in the present tense.  The story of how it came to be this way might be interesting, but I don't think this time it is.  What is interesting is what it is now and what we want it to be instead (and why).  You touch on that, but it gets lost in the story telling.

Something like this:

```
Refactor TopicBlog edit views.

The differences between the various edit views in topicblog
are quite small.  This commit removes the multiple almost-
same templates and replaces with a single template (topicblogbase_edit) that encapsulates the common points
and is overridden as needed for each class' edit
view.
```

Modify to be correct, maybe I missed some important details in writing that.

Shriukan33:
Reworded, and you got the gist of it.
As you said, maybe I expanded too much on some details, when your rewording perfectly does the job while being much shorter and clearer.

JeffAbrahamson:
LGTM

Issue #676, opened 2022-04-12
Add is_active flag to TBLauncher 


Shriukan33:
Active by default

JeffAbrahamson:
Note that this can't be part of the normal edit process, which will create a copy of the object.  This is a boolean on the model, but we'll reveal it in the list or something like that.

PR #682, opened 2022-04-13
Benja 675 add publish launchers

### LINKED ISSUE #675: Add publication feature on TBLauncher
Similar to what we do on the other TB

Shriukan33:
Launcher now has a publish function live every other TB Item

That's a step to #676 

Closes #675

### DIFF
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -177,78 +177,6 @@ class TopicBlogObjectBase(models.Model):
 
         return slug_fields
 
-
-class TopicBlogObjectSocialBase(TopicBlogObjectBase):
-
-    """
-    Represent all that is common in various the various
-    TopicBlogObjects that contain social media metadata and are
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
The first line of the commit message on f8fe44c3 is 80 characters, rather longer than the 50 recommended.  This breaks formatting when doing one-line logs.

Perhaps you could say something like
```
Refactor publication functions from social to base

TopicBlogLauncher needs to be able to signal publication.
This refactor just moves those functions from
TopicBlogObjectSocialBase to TopicBlogObjectBase.
    
Part of #675.
```

Note also the period a the end of the last line.  Only the first line should be a phrase devoid of punctuation, a minor bow to one-line log displays.  The rest should be proper English prose, obeying rules of grammar and punctuation.  Admittedly, custom is to say "Part of" instead of "This is part of" in such cases, which is a bit of an exception to the rule...

JeffAbrahamson:
Modulo comments, LGTM.

PR #683, opened 2022-04-19
Enable I18N to display date in French

### LINKED ISSUE #674: Fix localisation in dates
Django generated dates display in English, we want them in French (both email and CP web version)

Shriukan33:
The LANGUAGE_CODE setting has no effect while the I18N is disabled.
See doc :
https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-LANGUAGE_CODE

Closes #674


JeffAbrahamson:
LGTM

PR #684, opened 2022-04-19
Add Django debug toolbar

Shriukan33:
This toolbar allows us to see the SQL queries that Django is running,
and the time it takes to run them.
It also shows us various data, like the context of the request.
See doc : http://django-debug-toolbar.readthedocs.org/en/latest/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,13 +23,14 @@ cryptography==36.0.1
 decorator==4.4.1
 distlib==0.3.0
 distro==1.4.0
-dj-stripe==2.5.1
 Django==3.2.12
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
LGTM

So you really need 0 in internal ip's?

Shriukan33:
No I don't think it's mandatory

Télécharger Outlook pour Android<https://aka.ms/AAb9ysg>

JeffAbrahamson:
> No I don't think it's mandatory 

Also, it means something, and I don't think you mean to mean that.

Shriukan33:
Removed it

PR #685, opened 2022-04-20
Benja 643 create different types mailing lists

### LINKED ISSUE #643: Create different types of Mailing Lists
The idea behind the different types is a different handling of the unsubscription process, so far we see 3 types of mailing lists : 
* Petitions, from which you can't unsubscribe (you don't unsign a petition)
* Normal mailing list (newsletters), to which we have a unsub flow already running
* Press mailing list, specifically for journalists.

The later will have in their emails instead of "se désinscrire", "Préférences presse". 
Upon click, they are going to be shown a list of mailing lists with checkboxes : If you are subscribed the checkbox is checked, and vice versa.
At the bottom, a save preferences button.
 
<hr>
Current problems : 

* The TBEmailSendRecord is the only  model used for now, we might want to add a flag to tell it's a press or the lookup by ID will look for TBESendRecords instead of TBPSendRecords.
For now independently of if the sent object is a TBPress or TBEmail, a TBESendRecord is created.

Shriukan33:
This PR does : 
* Move the unsubscription from newsletters view to mailing list app, because it makes more sense to find code related to mailing list subscriptions inside mailing_list. 
* Create 3 types of mailing list through an enum field in the Mailing list model : Petition, Newsletter and Press
* Add a specific path to manage (aka sub/unsub) press mailing lists. 

closes #643


JeffAbrahamson:
The form tests in `tests_topic_blog_form.py` should surely test the forms directly rather than running the entire view through the chrome web driver.

JeffAbrahamson:
In 70ecff66, might it not be clearer to use full words instead of NL, PT, PR?  When you're not thinking about this bit of code, full words (like newsletter, petition, press-release) would immediately make sense and help avoid errors.

JeffAbrahamson:
In e0ab06a1 I don't think you're actually working on issue #643.

JeffAbrahamson:
This seems really fragile:
```
+        for key, _ in data.items():
+            # Mailing lists are prefixed with "ml"_" in the template
+            if key.startswith("ml_"):
```

JeffAbrahamson:
In 25fca0f1e, why is debug toolbar active during tests?

Shriukan33:
> The form tests in `tests_topic_blog_form.py` should surely test the forms directly rather than running the entire view through the chrome web driver.

Deleted the Selenium test, added one using Django API



> In [70ecff6](https://github.com/transport-nantes/tn_web/commit/70ecff66793ab7e844d2659bf08946a69cc0a360), might it not be clearer to use full words instead of NL, PT, PR? When you're not thinking about this bit of code, full words (like newsletter, petition, press-release) would immediately make sense and help avoid errors.

Changed the migration file and the model to use fully worded strings, in CAPS for the VALUE and lowercase for the key



> In [0eba8d8](https://github.com/transport-nantes/tn_web/commit/0eba8d82230478d62e056d6c7f137df07f193e45), why do you need to pass unsub_link around instead of doing url resolution from symbolic names?

The url given by the resolution (`reverse(..)`) doesn't include the website's FQDN, and thus, needs to be added at some point.
 

> In [25fca0f](https://github.com/transport-nantes/tn_web/commit/25fca0f1e1850d2770ced946fb93346ac1f70c5a), why is debug toolbar active during tests?

This was a mistake, it's not needed in the end.

PR #686, opened 2022-04-21
Add beacons to emails

### LINKED ISSUE #404: Set image beacon in TBEmails to track their perfomances
The receive beacon is an image (probably our logo) that gets an argument attached that encodes, using a cryptographic hash, a specific `TopicBlogEmailSendRecord`.

Ressources : 
* https://en.wikipedia.org/wiki/Web_beacon
* https://www.campaignmonitor.com/dev-resources/guides/coding-html-emails/ (some parts are quite old) 

issue not found :clown_face:

Shriukan33:
To get notified when one opens an email, we add a beacon that sends
a signal to our server when the email is opened, and images loaded.

Closes #404


JeffAbrahamson:
Is `"{{ request.scheme }}://{{ host }}{% url 'name' argument %}"` really best practice for specifying full URLs?  There's not some function somewhere that does this for us?

JeffAbrahamson:
Is there a benefit to saying
```
+    path_to_beacon = (Path(__file__).parent.parent / "asso_tn" / "static" /
+                      "asso_tn" / "beacon.gif")
```
vs
```
+    path_to_beacon = (Path(__file__).parent.parent / "asso_tn/static/asso_tn/beacon.gif")
```
?

Shriukan33:
> Is `"{{ request.scheme }}://{{ host }}{% url 'name' argument %}"` really best practice for specifying full URLs? There's not some function somewhere that does this for us?

Found this [SO answer](https://stackoverflow.com/a/18396895/16279937) but such notation isn't possible in Jinja 2, because ` {{ request.build_absolute_uri  {% url 'name' arg %} }} ` wouldn't parse.
Otherwise I didn't find ways to get the same result in the template :/ It has to be constructed in the context somehow or assembled like this/

JeffAbrahamson:
> > Is there a benefit to saying
> > ```
> > +    path_to_beacon = (Path(__file__).parent.parent / "asso_tn" / "static" /
> > +                      "asso_tn" / "beacon.gif")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > vs
> > ```
> > +    path_to_beacon = (Path(__file__).parent.parent / "asso_tn/static/asso_tn/beacon.gif")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> PathLib handles the separators so it doesn't fail on windows iirc, (sometimes sep is "") It's the most neutral way to write this
> 
> Besides, there is no benefit to the full string notation

Myriad things will fail if off linux anyway.  But I'm ok with the multiple args if you like that better.

JeffAbrahamson:
> > Is `"{{ request.scheme }}://{{ host }}{% url 'name' argument %}"` really best practice for specifying full URLs? There's not some function somewhere that does this for us?
> 
> Found this [SO answer](https://stackoverflow.com/a/18396895/16279937) but such notation isn't possible in Jinja 2, because `{{ request.build_absolute_uri {% url 'name' arg %} }}` wouldn't parse. Otherwise I didn't find ways to get the same result in the template :/ It has to be constructed in the context somehow or assembled like this/

There's also [this clever solution](https://stackoverflow.com/a/52193869/833300), which lets you jump out of template coding and into proper python.

See also [this](https://stackoverflow.com/questions/1451138/how-can-i-get-the-domain-name-of-my-site-within-a-django-template/40414100#40414100) and various other answers and comments on that page.  In particular, note that different methods present different surfaces to spoofing and proxy awareness.  The more this is abstracted so that it exists precisely one in our code, the better we can respond to eventual issues.

PR #702, opened 2022-04-27
Benja resize launchers and fix TBEdit form

Shriukan33:
This PR does 5 things around TBLaunchers : 

* Resize the launchers so they display evenly and with proper responsive, independently of the resolution of its thumbnail
* Force a minimal resolution for thumbnails images, with a validator in the form
* Fix a bug that prevented the errors to show while editing existing items. (For TBLa but not only)
* Improve the visual feedback of above errors.
* Enable the preview for TBLa to display both the launcher and the teaser versions


JeffAbrahamson:
10ef84b4 - Doesn't this make the image non-responsive?  Wouldn't it be better just to insist on a particular resolution?  I think there's also a way to insert into the upload process to change the resolution.

JeffAbrahamson:
528a4d92 - I wonder if you're testing too much.  For example, you are testing that the image is present, but do you really mean to test that it is precisely the "rounded" class and no other?

Wouldn't tests like this be better if they queried the DOM to check that, say, the html contains an IMG tag, that the SRC element in that IMG is what we expect, that the ALT element is present and what we expect?

JeffAbrahamson:
345c535 - LOL, good job!

Shriukan33:
Actually no, the width and height are set to a maximum of 435 px but their actual size is set in the style, I invite you to try it.

The 30vh is meaning this is maximum 30% of the viewports' height.
Width is automatically set to be 100% of its parent div which is responsive (flexbox)

Obtenir Outlook pour Android<https://aka.ms/AAb9ysg>

Shriukan33:
Yeah the test like many others we wrote the last 2 months or so are overly specific, this one included.

I didn't bother here, but certainly there are better and less rigid way to test our template tags.

JeffAbrahamson:
ok, then LGTM.

PR #703, opened 2022-05-02
Improve Launchers responsive

Shriukan33:
The images used to stretch thin, but now keep the same aspect ratio
no matter the size of the screen.


JeffAbrahamson:
Why is this needed
```
width="435" height="435"
```
given that you also say
```
max-width: 100%; max-height:30vh;
```

In particular, we had working code that said
```
<img class="rounded" src="..." alt="..." style="width:100%">
```
What is the interest in specifying max view height?

PR #710, opened 2022-05-03
Benja 709 fix login redirect

### LINKED ISSUE #709: Fix login redirect
When people are asked to login, the timed token they receive must take them where they wanted to go.

We discussed how to do this using django auth's built-in "next" mechanism.  When you construct the timed token, you construct it using the url that the next argument says is where the user should go on success.  Same if the user enters a password.

This is urgent: photo upload won't work without it.

Shriukan33:
This PR allows the user to access the desired page after login automatically.

This is done by passing the ?next= argument to the token and redirecting to it if present, after being logged in.

Closes  #709


JeffAbrahamson:
087639e036 - you really don't want to modify `make_timed_token()`.

```
token = make_timed_token(some_int, some_string)
url = reverse(name, token, {next: next_value})
```

or whatever the syntax is for adding `?arg=val` pairs to URLs is.

JeffAbrahamson:
LGTM.

PR #712, opened 2022-05-04
Allow devs to define locally their email backend

Shriukan33:
While keeping the option to specify an email backend locally, users who don't
have it set will still be able to use django's default.

### DIFF
--- a/transport_nantes/transport_nantes/settings.py
+++ b/transport_nantes/transport_nantes/settings.py
@@ -120,10 +120,13 @@ AUTH_PASSWORD_VALIDATORS = [
 ]
 PASSWORD_RESET_TIMEOUT_DAYS = 1
 
-if ROLE == 'dev':
+if 'EMAIL_BACKEND' in settings_local.__dict__:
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
Why do you import `xml.dom.minidom.Attr`?

Shriukan33:
> Why do you import `xml.dom.minidom.Attr`?

Missed an auto import :) Deleted

JeffAbrahamson:
I like the idea of an override, but I'd think it more natural to do things slightly differently:
* If the user doesn't specify one, then use the logic that turns on `ROLE == 'dev'`
* If the user does specify one, then use it.  But check this with `if 'EMAIL_BACKEND' in settings_local` seems more pythonic and more communicative to me than the try-except clause.

Shriukan33:
Like so ?

JeffAbrahamson:
Yup, great.

LGTM, but could you just change the commit comment before merge: it's not django's default, it's our webapp's default.

PR #713, opened 2022-05-04
Make same_slug_warning adapt on the model we're using

### LINKED ISSUE #700: Fix improper information about the number of slugs used in TB  forms
In the TB forms (except TBItem), when typing the slug of the TB object, the form tells us that a certain number of items share this slug. 
Those items are TBItems. 
To be fair to all TB-Objects and not getting a confusing information, we should display the number of same slug items that share the same class (TBI, TBE, TBLa, TBP ... )

Shriukan33:
Based on PR #712 

The JS used to request only TopicBlogItems slugs.
Now that we're using several models, we're requesting slugs for the
model we're using.

Closes #700

### DIFF
--- a/transport_nantes/topicblog/static/topicblog/same_slug_warning.js
+++ b/transport_nantes/topicblog/static/topicblog/same_slug_warning.js
@@ -1,8 +1,13 @@
 // Calls server to get a list of all slugs and
 // how many of each slug there are.
+const MODEL_NAME = JSON.parse(document.getElementById('model_name').textContent);
+
 async function get_slug_dict() {
     data = await $.ajax({
         url: '/tb/ajax/get-slug-dict/',
+        data: {
+            model_name: MODEL_NAME
+        },
     })
     return data;
 }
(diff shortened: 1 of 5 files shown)


JeffAbrahamson:
LGTM, except here:

```
@@ -296,7 +297,11 @@ class TopicBlogBaseList(LoginRequiredMixin, ListView):
 @StaffRequired
 def get_slug_dict(request):
     """Return a list of all existing slugs"""
-    qs = TopicBlogItem.objects.order_by('slug').values('slug')
+    model_name = request.GET.get('model_name', None)
+    if model_name is None:
+        return JsonResponse({"error": "model_name is required"})
+    model = apps.get_model('topicblog', model_name)
+    qs = model.objects.order_by('slug').values('slug')
     dict_of_slugs = Counter([item['slug'] for item in qs])
     return JsonResponse(dict_of_slugs, safe=False)
```

shouldn't the condition be `if not model_name:`?

Otherwise, LGTM.

PR #717, opened 2022-05-04
Allow admin panel to filter by and search by email instead of username

Shriukan33:
image

The column display can't be user.email afaik, maybe there is a way, but considering this is low priority, I guess this is better than the default


JeffAbrahamson:
LGTM.

PR #718, opened 2022-05-04
Fix TopicBlogLauncher preview

Shriukan33:
The template used to display the published item if any, even when
we requested a single item.

To not have to maintain a separate version of the template tag,
I added a "preview" flag that is set in the template that is used to
display the previews.
The index template doesn't and must not carry this flag, as the "page"
context variable is very different in this context.

Also added a fallback to help visualize if no image is provided.

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/content_launcher.html
+++ b/transport_nantes/topicblog/templates/topicblog/content_launcher.html
@@ -6,8 +6,8 @@
     <p>Cette page n'est pas privée mais n'est pas destinée au public.
 	Elle peut disparaître du jour au lendemain.
     </p>
-    {% launcher page.slug %}
+    {% launcher page.slug is_preview=True%}
     <hr>
-    {% item_teaser page.slug %}
+    {% item_teaser page.slug is_preview=True %}
 </div>
 {% endblock content %}
--- a/transport_nantes/topicblog/templates/topicblog/template_tags/launcher.html
+++ b/transport_nantes/topicblog/templates/topicblog/template_tags/launcher.html
@@ -1,3 +1,5 @@
+{% load static %}
+
 <div class="col-md-3">
 	<div class="thumbnail">
 		{% comment %}
@@ -14,7 +16,11 @@
 	    {% endcomment %} 
 			{% if launcher.article_slug %}<a href="{% url 'topic_blog:view_item_by_slug' launcher.article_slug %}">{% else %} 404 {% endif %}
 			<div class="shadow-md tbla-container">
-				<img class="rounded" src="{{ launcher.launcher_image.url }}" alt="{{ launcher.launcher_image_alt_text }}">
+				{% if launcher.launcher_image %}
+					<img class="rounded" src="{{ launcher.launcher_image.url }}" alt="{{ launcher.launcher_image_alt_text }}">
+				{% else %}
+					<img class="rounded" src="{% static 'asso_tn/beacon.gif' %}" alt="No image provided">
+				{% endif %}
 			</div>
 			<div class="caption pt-2">
 				<h3 class="font-weight-light">{{ launcher.headline }}</h3>
--- a/transport_nantes/topicblog/templatetags/launcher.py
+++ b/transport_nantes/topicblog/templatetags/launcher.py
@@ -7,25 +7,53 @@ register = template.Library()
 logger = logging.getLogger("django")
 
 
-@register.inclusion_tag('topicblog/template_tags/launcher.html')
-def launcher(slug):
-    launcher = TopicBlogLauncher.objects.filter(
-        slug__iexact=slug, publication_date__isnull=False).first()
+@register.inclusion_tag('topicblog/template_tags/launcher.html',
+                        takes_context=True)
+def launcher(context, slug, is_preview=False):
+    if is_preview:
+        # In TBBaseViewOne, "page" holds the object we're trying to visualize.
+        # ViewOne is a view made to display a single object.
+        launcher = context["page"]
+    else:
+        # In the context of a launcher displayed in another view than ViewOne,
+        # we need to get the object from the database, as the "page" variable
+        # is not available. (typically in the index)
+        # because no database rule enforces the rule "one publication date per
+        # slug", we need to get the most recent one just in case
+        launcher = TopicBlogLauncher.objects.filter(
+            slug__iexact=slug, publication_date__isnull=False
+            ).order_by("-publication_date").first()
     if launcher is None:
         logger.error(f"item_teaser failed to find slug: \"{slug}\".")
     return {'launcher': launcher}
 
 
-@register.inclusion_tag('topicblog/template_tags/item_teaser.html')
-def item_teaser(slug):
-    launcher = TopicBlogLauncher.objects.filter(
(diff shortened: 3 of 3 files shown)


JeffAbrahamson:
3d33d0bad - I don't understand the commit message...  :-(

The logic might be clearer if you skipped the "not".  If it's this, do one thing, else do the other.  Rather than "if it's not this, ...".

This very definitely merits code comments to explain why you are doing this.

JeffAbrahamson:
You talk about displaying previews, but I don't know what a preview is.  And I have no idea what your contract is on the meaning of `context['page']`, but it feels pretty draconian to assume that all clients of this templatetag will respect a particular naming scheme for their context dicts.

You are, however, displaying a launcher and a teaser in a context where you want to display even if not published.  I think it would be far clearer to say that.

For example:
```
def launcher(context, slug, require_published=True):
```

More fundamentally, however, I think your logic is wrong, for several reasons.  (Note: there are no tests.)
* You are doing a search and taking the first item, but you haven't ordered by anything.  What you get is undefined.
* Even if you ordered by something, there's no guarantee the user is requesting to preview the most recent.  Maybe I have ten versions of the launcher I've published numbers 5 and then 10, and I'd like to preview number 8 to remind myself what that one looked like.

So I think what you really would want is to pass a pk_id, which defaults to `None`.

```
if pk_id:
    launcher = TopicBlogLauncher.objects.get(pk=pk_id)
else:
    launcher = TopicBlogLauncher.objects.filter(
        slug__iexact=slug,
        publication_date__isnull=False).order_by(
            '-publication_date').first() 
```

You'd need a try/except because of the `get()`, but `filter().first()` on a pk_id feels weird.

And you would need to update your log statement to include the pk_id if appropriate.

JeffAbrahamson:
878419e92 - "the page would" -> "the page previously".

I think it's worth a code comment that you only expect missing images during development of new launchers, that a production-ready and published launcher must have an image, that the `is_publishable()` function guarantees that.

Shriukan33:
> You talk about displaying previews, but I don't know what a preview is. And I have no idea what your contract is on the meaning of `context['page']`, but it feels pretty draconian to assume that all clients of this templatetag will respect a particular naming scheme for their context dicts.
> 

The view displaying this page is TopicBlogLauncherViewOne
https://github.com/transport-nantes/tn_web/blob/7a8b07cd366e9035d6e7271531da59359f3663f1/transport_nantes/topicblog/urls.py#L32
which itself inherits from TopicBlogBaseViewOne
https://github.com/transport-nantes/tn_web/blob/35e8f5ec9ff17eb908cf0c452871ebf3859db4ca/transport_nantes/topicblog/views.py#L190
The get_context_data is defined as this : 
https://github.com/transport-nantes/tn_web/blob/35e8f5ec9ff17eb908cf0c452871ebf3859db4ca/transport_nantes/topicblog/views.py#L203-L216

you can see that in the View One context, the "page" variable is the given tb object (here, Launchers) 

https://github.com/transport-nantes/tn_web/blob/35e8f5ec9ff17eb908cf0c452871ebf3859db4ca/transport_nantes/topicblog/views.py#L212

So yes, unless we change the base class, designating "page" as the object works. 

> You are, however, displaying a launcher and a teaser in a context where you want to display even if not published. I think it would be far clearer to say that.
> 
> For example:
> 
> ```
> def launcher(context, slug, require_published=True):
> ```

I can reword the "preview" flag to requires_published, if that makes it clearer indeed !

> More fundamentally, however, I think your logic is wrong, for several reasons. (Note: there are no tests.)
> 
>     * You are doing a search and taking the first item, but you haven't ordered by anything.  What you get is undefined.

Not quite. 
It's voluntarily that we do not order it because our publication system makes sure that at any times, there is exactly 1 or 0 items that have a `publication_date`.
And we do fetch that item

```python
def launcher(context, slug, preview=False):
    if not preview:
        launcher = TopicBlogLauncher.objects.filter(
            slug__iexact=slug, publication_date__isnull=False).first()
```

The publication does this : 
https://github.com/transport-nantes/tn_web/blob/35e8f5ec9ff17eb908cf0c452871ebf3859db4ca/transport_nantes/topicblog/views.py#L228-L235

As you can see, we fetch every item in a given model, and set `publication_date` to `None` excluding the item we just published.
This is how I know that the Launcher query will return 1 or 0 objects.
This is also why I use first() : First will return either None or the object I want. 

>     * Even if you ordered by something, there's no guarantee the user is requesting to preview the most recent.  Maybe I have ten versions of the launcher I've published numbers 5 and then 10, and I'd like to preview number 8 to remind myself what that one looked like.

With the "preview" flag I added to the template displaying the ViewOne for launchers, you don't fetch the published item, but the one that is saved into the context by the view.
Remember : 
https://github.com/transport-nantes/tn_web/blob/35e8f5ec9ff17eb908cf0c452871ebf3859db4ca/transport_nantes/topicblog/views.py#L207-L212

Without the preview flag it has the same behavior as any other view that displays the published item : It fetches the only one with a publication_date.

```python
def launcher(context, slug, preview=False):
    if not preview:
        launcher = TopicBlogLauncher.objects.filter(
            slug__iexact=slug, publication_date__isnull=False).first()
```

> 
> So I think what you really would want is to pass a pk_id, which defaults to `None`.
> 
> ```
> if pk_id:
>     launcher = TopicBlogLauncher.objects.get(pk=pk_id)
> else:
>     launcher = TopicBlogLauncher.objects.filter(
>         slug__iexact=slug,
>         publication_date__isnull=False).order_by(
>             '-publication_date').first() 
> ```
>
> You'd need a try/except because of the `get()`, but `filter().first()` on a pk_id feels weird.
> 
> And you would need to update your log statement to include the pk_id if appropriate.

The get is already done in the view before rendering the template : 

https://github.com/transport-nantes/tn_web/blob/35e8f5ec9ff17eb908cf0c452871ebf3859db4ca/transport_nantes/topicblog/views.py#L208

Shriukan33:
> 3d33d0bad - I don't understand the commit message... :-(
> 
> The logic might be clearer if you skipped the "not". If it's this, do one thing, else do the other. Rather than "if it's not this, ...".
> 
> This very definitely merits code comments to explain why you are doing this.

Will turn the not to a positive check with rewording

JeffAbrahamson:
Your explanation above makes sense.  I think you just need to comment adequately, because there are a lot of strong assumptions (contracts) in place, and these things aren't necessarily all in the head of a reader of the code.

Shriukan33:
Added comments and reworded commits.
Also reworded the flag and used a positive assertion (`if is_preview` instead of `if not preview`)

JeffAbrahamson:
LGTM.

I'm still a bit confused why it's the right thing to do for the view to provide the launcher in the page in one case but to let the template look it up in the other.  I'd think we'd want the view _always_ to provide the lookup so that the logic is in the view and not distributed over templates.

Shriukan33:
> LGTM.
> 
> I'm still a bit confused why it's the right thing to do for the view to provide the launcher in the page in one case but to let the template look it up in the other. I'd think we'd want the view _always_ to provide the lookup so that the logic is in the view and not distributed over templates.

This works for now, but sure we can discuss changes once we tackle the more dynamic aspect of launchers later down the line. :slightly_smiling_face:

PR #719, opened 2022-05-04
Make users submitting photos automatically subscribed to mailing list

Shriukan33:
They have shown interest by participating, so it seems fair to
automatically subscribe them to the operation pieton mailing list.

### DIFF
--- a/transport_nantes/photo/templates/photo/photoentry_form.html
+++ b/transport_nantes/photo/templates/photo/photoentry_form.html
@@ -32,6 +32,8 @@
                     <li>J'atteste que l'essentiel de la photo soumise est la mien,
                     que toute manipulation éventuelle est technique, pas l'effet
                     principal.</li>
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
On the commi comment: "Subscribe users to mailing list on photo submission" is a bit more natural, less dictatorial. :wink:

Shriukan33:
> I think it might be better to say "J'accepte que ..." rather than "Je suis d'accord que ...". They are very different, and all we need is acceptance, not moral and philosophical agreement. (Aside: I once refused to sign an employment contract because it had language that said I agreed a certain thing was reasonable that I didn't think was reasonable. I might have signed if they'd just said I had to do it, but pushing me to be in agreement was repugnant.)

Reworded both commit message and bullet :)

JeffAbrahamson:
LGTM.

PR #720, opened 2022-05-04
Modify the placeholder line for Operation pieton's form

Shriukan33:
From 
"Exemples : j'y habite, j'y travaille, c'est compliqué, autre..."
to
"Exemples : j'y habite, j'y travaille, c'est compliqué, Aucun mais je vous aime bien :)..."

### DIFF
--- a/transport_nantes/photo/forms.py
+++ b/transport_nantes/photo/forms.py
@@ -20,8 +20,9 @@ class PhotoEntryForm(forms.ModelForm):
             'category': forms.Select,
             'relationship_to_competition': forms.Textarea(
                 attrs={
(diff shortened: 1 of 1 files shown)


Shriukan33:
> I really hate to nitpick on something so simple, but why is "Aucun" the only one that's capitalised?
> 
> BTW, is there a reason to drop "autre"?
> 
> Should we say "je vous adore" instead of "je vous aime bien". Adding the "bien" makes it feel too serious.
> 
> I'd vote for no emoji, wdyt?

Reworded

JeffAbrahamson:
LGTM.

Issue #725, opened 2022-05-10
Write function to clean up TBObjects


JeffAbrahamson:
TBobjects that have never been published and are more than k days old (suppose 15 for now) should be deleted when this script is run.

PR #729, opened 2022-05-12
Add function to clean up TBObjects

mcmikashi:


JeffAbrahamson:
I think `update_tb_object_scheduled_date()` should perhaps filter on the current slug.  Two reasons motivate me.  First, imagine that the system starts to be large.  Filtering to current slug would reduce the amount of work that potentially needs to happen on viewing the list.  Second, in the (highly unlikely) event that we have lots of concurrent users, it reduces the chances of conflict.  Although postgresql should isolate adequately to avoid that, but that isolation comes at some cost.

JeffAbrahamson:
That would be `K_SCHEDULE_FOR_DELETION` rather than `K_SCHEDULE_FOR_DELATION`. ;-)

JeffAbrahamson:
I don't think we need to create a new util.py file.  The constant is naturally a property of the TBObject base class.

The two functions operate on slugs (object classes) rather than individual TBObjects, so let's put them in views.py with names that reflect that.  Also, we might encode a bit of context in the function names.

```
def schedule_moribund_slug_instances_for_deletion(tb_object)
def delete_scheduled_moribund_slug_instances(tb_object)
```

Here, `tb_object` _might_ be an actual object.  (Otherwise, call it a class, because it's a big confusing.)  And then you can do the search by saying (I'm totally inventing this and haven't tested it) something like this:

```
tb_object.model.objects.filter(
    slug=tb_object.slug,
    scheduled_for_deletion_date=None).update(...)
```
etc.  Wdyt?

Both of those functions could benefit from docstrings.  As well as the model changes, too.

JeffAbrahamson:
Don't forget to tag the issue #725.

Shriukan33:
Also, do we want to delete all not published items for a slug ? I thought we would like to keep the previously published versions, so an extra condition would be `first_publication_date == None`.

JeffAbrahamson:
> Also, do we want to delete all not published items for a slug ? I thought we would like to keep the previously published versions, so an extra condition would be `first_publication_date == None`.

Correct, we should keep anything that has ever been published and only delete old things that have never been published.

Shriukan33:
The migration file is number 42, while master is already at 43. 
To avoid conflicts on master, please rebase and re-do your migration file :)

mcmikashi:
- Rebase on master
- Add test
- Move the functions on the view

transport_nantes/topicblog/forms.py
```diff
@@ -14,7 +14,8 @@ class Meta:
         model = TopicBlogItem
         # Admins can still edit those values
         exclude = ('item_sort_key', 'servable', 'user',
-                   'publication_date', 'first_publication_date', 'publisher')
+                   'publication_date', 'first_publication_date', 'publisher',
+                   'scheduled_for_deletation_date')
```
Shriukan33, line 18: deletation -> Deletion

transport_nantes/topicblog/templates/topicblog/topicblogbase_list_one.html
```diff
@@ -42,6 +46,12 @@
             <br>
             Créé {{ object.date_modified|date:"l d-m-Y à G:i" }} par {{ object.user.email }}
             <br>
+            {% if not object.first_publication_date and not object.publication_date and object.scheduled_for_deletation_date  %}
+                <span class="{% if object.get_days_defore_deletation <= 0 %} text-danger {% endif %}">
```
Shriukan33, line 50: deletation -> Deletion

transport_nantes/topicblog/migrations/0047_topicblogemail_scheduled_for_deletation_date_and_more.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 4.0.4 on 2022-05-23 16:19
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('topicblog', '0046_auto_20220518_1241'),
```
Shriukan33, line 9: We added some more migrations, we're now at 48

JeffAbrahamson:
Thanks!

In addition to the other comments in the review, could you do the following before we merge:
* s/deletation/deletion/;
* tag issue
* fix the trailing white space

mcmikashi:
- Fix spelling
- Reword the commit message
- Remove trailing white space

Shriukan33:
Could you please rebase ? :)

mcmikashi:
- Rebased

Shriukan33 (approved):
LGTM

Shriukan33:
There are conflicts in rebase now @mcmikashi :/

mcmikashi:
- Rebased

Shriukan33:
@JeffAbrahamson  ?

JeffAbrahamson:
```
localhost tn_beta_django@tn_beta2_db=> SELECT
            first_publication_date, publication_date, 
            date_modified, scheduled_for_deletion_date
        FROM topicblog_topicblogitem
        WHERE slug = 'devenir-benevole';
    first_publication_date    |       publication_date       |         date_modified         | scheduled_for_deletion_date 
------------------------------+------------------------------+-------------------------------+-----------------------------
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.703823+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.742883+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.842589+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.925381+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.938666+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:41.117461+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:41.132057+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-05-18 13:22:57.935134+00 | (NULL)
 2022-05-18 13:24:55.42542+00 | 2022-05-18 13:24:55.42542+00 | 2022-05-18 13:24:55.447974+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.914582+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.928461+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.931772+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.935062+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.942292+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.945654+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.948656+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:41.464377+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:41.468015+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:41.514054+00 | (NULL)
 (NULL)                       | (NULL)                       | 2022-02-03 17:23:40.90771+00  | (NULL)
(20 rows)

Time: 1.917 ms
localhost tn_beta_django@tn_beta2_db=>
```

Shriukan33:
Did you visit the delete url ?

JeffAbrahamson:
> Did you visit the delete url ? You need to click on the trashcan icon I believe to start the countdown

Hmm, perhaps I should try that. ;-)

JeffAbrahamson:
> > Did you visit the delete url ? You need to click on the trashcan icon I believe to start the countdown
> 
> Hmm, perhaps I should try that. ;-)

No joy.

JeffAbrahamson:
I'll have a closer look next week.

mcmikashi:
Change : 

- Small fix : check if the url have get data if not add a new get data because the list url when we create a new item doesn't have the “from” in the url.

For delete tb_oject we need to check the list_items_by_slug view once we do that if he doesn't have cheduled_for_deletion_date it will set to today's date and after 15 days it will be deletable with the delete url.

JeffAbrahamson:
Hmm, clicking the trash can icon (which I think is meant to clean up even though the icon suggests deletion) adds a `delete=true`.  Each time.  So here I am at

`http://localhost:8000/tb/admin/t/list/une-meilleure-mobilit-pour-le-climat/?from=articles&delete=true&delete=true`

JeffAbrahamson:
I've fixed the conflicts and merged this.

PR #739, opened 2022-05-18
Benja sendview refactor

Shriukan33:
Refactor SendView

TopicBlogPress and TopicBlogEmails are the two classes that can be
send through email.
Their code shared lots of similarities with some variations.

The refactor consists in creating a new View base class that is then
implemented by the two classes.


JeffAbrahamson:
LGTM.

Suggested commit comment amendment:

Instead of
```
Refactor SendView
    
TopicBlogPress and TopicBlogEmails are the two classes that can be
send through email.
Their code shared lots of similarities with some variations.
    
The refactor consist in creating a new View base class that is then
implemented by the two classes.
```
write
```
Refactor SendView
    
TopicBlogPress and TopicBlogEmails send mail and are almost
identical.  Refactor to avoid repeating common code.
```

Shriukan33:
Reworded !

PR #742, opened 2022-05-18
Benja 736 send record status

### LINKED ISSUE #736: Add status field to SendRecord
Values are PENDING, FAILED, SENT.

Shriukan33:
This PR adds 
* A status field that has one of these 3 values : PENDING, FAILED or SENT
* Disable the auto-now and the set of a send_time in view. 

Closes #736


JeffAbrahamson:
If I'm not getting confused, I think the commit message is off on the first commit.  It's not that we need a _signal_ from AWs.  (What signal, precisely, are you expecting if so?  State it.)  It's that we should provide the send_time only once we have passed it off to SES.  And this is different, because (1) we have to create the send record before we can hand off to SES, and something could go wrong, and (2) we'd normally expect to hand off to celery and ask celery to do it.

In other words, we're recording that we sent the message, not that we intended to send the message and we just blithely assume that everything went fine from intention to fact.

JeffAbrahamson:
The status enum needs documentation.  What, precisely, do these words mean to us?

Shriukan33:
Added comments on the Enum
Also added a comment above Send time
Reworded commits
Dropped the status_note field for now, because we don't have a clear purpose for it for now, YAGNI rule

JeffAbrahamson:
We add the send_time when SES accepts the mail for sending.  We don't have visibility on when it gets sent, and SNS might later (days later) tell us it wasn't.  So it's important to be clear on the meaning here.

I'm talking about the commit message,
```
commit b3e9c686b564f59ab099a9dda51110f36ce1ae3c
Author: Benjamin Mourgues <benjamin@mobilitains.fr>
Date:   Wed May 18 14:24:09 2022 +0200

    Make send_time on send records not auto-added
    
    Emails are handled by Amazon SES, we want to set a send_time only if
    Amazon SES actually sends the email.
    
    Part of #736

 transport_nantes/topicblog/migrations/0045_auto_20220518_1224.py | 23 +++++++++++++++++++++++
 transport_nantes/topicblog/models.py                             |  6 ++++--
 transport_nantes/topicblog/views.py                              |  1 -
 3 files changed, 27 insertions(+), 3 deletions(-)

```

transport_nantes/topicblog/models.py
```diff
@@ -776,9 +776,21 @@ class TopicBlogEmailSendRecord(models.Model):
 
     """Represent the fact that we sent an email.
     """
+
+    class StatusChoices(models.TextChoices):
+        # This is the default state
+        PENDING = 'PENDING', "Pending"
```
JeffAbrahamson, line 782: It's not so much the default state as the initial state.

transport_nantes/topicblog/models.py
```diff
@@ -776,9 +776,21 @@ class TopicBlogEmailSendRecord(models.Model):
...
+        # Toggled to this state when Amazon SES sends the email
+        SENT = 'SENT', "Sent"
```
JeffAbrahamson, line 784: Present tense makes more sense.
More importantly, it's when SES _accepts_ the mail for sending.

transport_nantes/topicblog/models.py
```diff
@@ -776,9 +776,21 @@ class TopicBlogEmailSendRecord(models.Model):
...
+        # Toggled to this state if Amazon SES refuses to send the email
```
JeffAbrahamson, line 785: It's not that SES refuses, it's that SES accepted the mail for sending and later reports that it was unable to do so for reasons quite out of its control.  This is quite different from refusing.

transport_nantes/topicblog/models.py
```diff
@@ -960,9 +972,21 @@ class TopicBlogPressSendRecord(models.Model):
...
+        FAILED = 'FAILED', "Failed"
+
     slug = models.SlugField(max_length=90, allow_unicode=True, blank=True)
     mailinglist = models.ForeignKey(MailingList, on_delete=models.PROTECT)
     recipient = models.ForeignKey(User, on_delete=models.PROTECT)
+    status = models.CharField(
+        max_length=10, choices=StatusChoices.choices,
```
JeffAbrahamson, line 988: PostgreSQL doesn't care about `max_length`, but django does.  It's generally wise to give yourself extra space to avoid having to fiddle with the schema definition because you add an enum element.

JeffAbrahamson:
Otherwise, LGTM.

transport_nantes/topicblog/models.py
Shriukan33: changed to
# This is the initial state

transport_nantes/topicblog/models.py
Shriukan33: Changed to 
`# Toggle to this state when Amazon SES accepts to send the email`

transport_nantes/topicblog/models.py
Shriukan33: Changed to 
`        # Toggle to this state if Amazon SES accepted the sending, but is unable to deliver the email`

JeffAbrahamson:
LGTM.

PR #748, opened 2022-05-19
Refactor SendRecord models

### LINKED ISSUE #740: Make abstract classes for SendRecord
> > In `topicblog/views.py`, you reveal that you have copied and pasted code to two different places. Might this be the time to clean that up? ;-)
> 
> Did a refactor of the Send Email code so both Press and Email use the same. On this topic, Press still uses TBESendRecords (while there is a PressSR model).
> 
> I prepared the transition, but a next step would be to rework the beacon so it's able to tell if the ID it is receiving is a TBESendRecord or a TBPSendRecord.
> 
> As it's non blocking for now, I didn't do that yet.
> 
> Which brings me to another question : Do we want to split SendRecords in two or not ?

I'd think we'd want a single abstract `SendRecordBase` from which `TBEmail` and `TBPress` (in brief, used by `SendableObjectMixin`, which was a nice refactor, btw).

More specifically, I think the inheritance would be

```
SendRecordBase                    (abstract)
  \- SendRecordTransactional
  \- SendRecordMarketingBase      (abstract)
       \- SendRecordEmail
       \- SendRecordPress
```

Only marketing mails are sent to mailinglist objects.  Transactional mails are only sent to a single user.  (I still don't remember why we want the mailinglist on `SendRecordMarketingBase`, but we can always remove it later or remove it now and re-add it later.)

Transactional emails don't permit unsubscribe, because it's precisely in response to an action you took.  Marketing emails do permit unsubscribe, so the field makes sense there.

Also, marketing emails you should only ever get one of, even if we send to multiple lists, so we'd want a unique index on `(slug, recipient)`.  Transactional emails, well, every time you make a gift or login, you'll get a mail about it, and we shouldn't not send you a thank you note because we already sent you one with this thank-you template (slug).

Remember, the goal is that sending mail should be something like

```
def send_mail(send_record_id):
    ...
```
and that would queue it for celery to handle.  So celery is going to look up the send record and do whatever it takes to send the mail when it gets to it.

_Originally posted by @JeffAbrahamson in https://github.com/transport-nantes/tn_web/issues/735#issuecomment-1129193914_

Shriukan33:
Refactor with abstract classes according to #740

Will extend the PR (or make another) with application of these in Authentication views and TopicBlog views later :slightly_smiling_face: 

Closes #740


JeffAbrahamson (changes requested):
Two migrations for a single commit?  That smells of not cleaning up the branch.

I think it would be easier to see that this is correct if it were two commits:
* The first commit would only refactor to the abstract model bases, and it would have a migration that sorted table names but clearly didn't change anything else.
* The second commit would add the new fields and make the concrete classes actually be different.

Wdyt?

Shriukan33:
Applied the modifications, I had 2 migrations because I forgot to add the null=True first time so I did another migration. 

I did split in 2 this time, adding the field in the 2nd commit. 

It really only is a refactor as you can see in migrations, the only 2 diffs are the creation of Transactional Send Record and the addition of AWS message ID to the fields.

JeffAbrahamson:
LGTM.

PR #749, opened 2022-05-23
Benja handle ses events

Shriukan33:
Based on #748 

This PR adds the "Sent" even handling.
It also add error handling when sending an email. If the sending is rejected, the send record associated with the email gets a 'FAILED' status, and the rest of emails is processed.

A successfully sent email is properly updated with a "SENT" status, an updated "sent_time" and AWS message id.

### DIFF
--- a/transport_nantes/authentication/tests_auth.py
+++ b/transport_nantes/authentication/tests_auth.py
@@ -1,5 +1,5 @@
 from django.contrib.auth.models import User
-from django.test import TestCase
+from django.test import TestCase, TransactionTestCase
 import datetime
 
 from django.urls.base import reverse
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
In 96c03c8cf8f, you say "I suspect that ...".  So maybe you couldn't figure it out, but if it was hard to understand, can you help the next reader to understand what you did understand and how by fully documenting (in code or in git, as seems appropriate) what you did and what you concluded and why?

Shriukan33:
Reworded

JeffAbrahamson:
LGTM.

(Actually, I'm not re-reviewing, but the reword based on before is enough for me.)

PR #755, opened 2022-05-25
Benja celery

Shriukan33:
Draft while tests aren't sorted, it breaks some because outbox is empty, now that mail are handed to Celery 

Based on #749 

This branch merge our celery settings and expand it to allow TopicBlogSendView to hand to celery the sending part.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 amqp==5.1.1
-anyjson==0.3.3
 appdirs==1.4.3
 asgiref==3.4.1
 astroid==2.5.6
@@ -36,6 +35,7 @@ distlib==0.3.0
 distro==1.4.0
 dj-stripe==2.5.1
 django-amazon-ses==4.0.1
+django-celery-results==2.3.1
 django-countries==7.2.1
 django-crispy-forms==1.14.0
 django-debug-toolbar==3.2.4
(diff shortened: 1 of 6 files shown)


JeffAbrahamson:
Why do you want a results backend?

Shriukan33:
> Why do you want a results backend?

Simply allows you to see the task list status and history on django admin 

image

thought it would be a nice addition, although it's optional

JeffAbrahamson:
In `celery.py` you've added comments that make sense in a tutorial, but seem a bit repetitive in production code.

JeffAbrahamson:
You should generally never serialise objects to pass to celery.  You should have clear, well-defined interfaces that, mostly, are just a pk_id from some table and all the necessary information lives in that one table.

In addition, celery tasks need documentation, because the caller is not only not nearby, it is in another process that is not even guaranteed to be running on the same host or with the same code base, although in our setting both of those things are currently true.

JeffAbrahamson:
Have you read the celery docs?  [In particular](https://docs.celeryq.dev/en/latest/getting-started/next-steps.html#calling-tasks),

> Also note that result backends aren’t used for monitoring tasks and workers: for that Celery uses dedicated event messages (see [Monitoring and Management Guide](https://docs.celeryq.dev/en/latest/userguide/monitoring.html#guide-monitoring)).

PR #757, opened 2022-05-25
Benja new mailinglist donators

Shriukan33:
This PR creates a new type of mailing list named "DONORS". 
It also automatically add donors to a current year's mailing list (donors-2022, donors-2023 and so on)


transport_nantes/stripe_app/views.py
```diff
@@ -451,6 +453,12 @@ def make_donation_from_webhook(event: dict) -> Donation:
         "originating_view": metadata["originating_view"],
         "originating_parameters": metadata["originating_parameters"],
     }
+    try:
+        logger.info(f"Adding {kwargs['user']} to donators mailing list...")
+        add_user_to_donator_mailing_list(user=kwargs['user'])
+    except Exception as e:
```
JeffAbrahamson, line 459: Catch generic exceptions after catching specific ones.  Here, anything could happen and you'll report the same error.  Think through what could go wrong (key error in `kwargs`, can't access database, what else can go wrong?) and catch and report individually, which also allows us to think through how to respond to the user and maybe how to recover.  Catching the base case is sort of saying "I didn't think this could happen, but at least I caught it", which is useful, but it should also signal that there's a case we didn't think through.

JeffAbrahamson (changes requested):

transport_nantes/stripe_app/views.py
Shriukan33: I did edit the commit to check the kwargs, but couldn't think of other errors and / or better handling of said errors. There are two pieces : The user and the mailing list. 
The mailing list is get or create, so the absence of it is handled. The user is now handled, with the .get().

JeffAbrahamson: What specific exception would you expect there?

Shriukan33: None I can think of

JeffAbrahamson: I don't believe you.  But if you really think that this can never happen, perhaps you are being a bit shy on your logging.  Something very serious has happened that leads us to 500 here, but you're just saying "an error happened".  How will we debug this?  We don't know anything about the type of error or (in the case of the missing user) where it happened -- we'll just see a 500 served but no other logged info.

I think it also deserves comment why you can catch an unknown exception, log that an error occurred (info, not error) and then continue.  Why are you sure you can create the donation object?  Maybe you can, but this is all quite light, assuming that things will go well.

The same thing happens a bit further down: you log some info and then promptly 500.  Either it's an error that we expect never to be able to happen under normal circumstances (log.error and throw) or else it's ordinary (log.info and don't 500).

JeffAbrahamson:
LGTM.

JeffAbrahamson (approved):

Issue #764, opened 2022-05-30
TBPress base template is missing contact info


JeffAbrahamson:
A contact email and phone number should be available in the model.  If absent, the email should default to press@ whilst the phone number should be absent (i.e., no default value).

Shriukan33:
Hey I don't feel this is especially high, given that the contact info doesn't change very often or at all, the default might be enough. Want me to add a "phone_number" and a "mail_contact" field to TBPress ?

JeffAbrahamson:
Didn't we already implement this with the contact field?

Shriukan33:
Haha, right ! Guess we can close this now

PR #767, opened 2022-05-30
Quick fix to add default email to press contacts

JeffAbrahamson:
In real life, this should check the database for a configured email
address and even phone number.  But this provides the default
behaviour whilst waiting for that.

This is a small bit of #764.


Shriukan33 (commented):
image

Is this how you want it to look ?
Also if you want it to be clickable, it should hold a `<a href="mailto:press@mobilitains.fr">` tag, but maybe you don't want that ?

JeffAbrahamson:
I'm ok with that look.  And I'm ok with it not being clickable.

Shriukan33 (approved):
LGTM then !

Issue #779, opened 2022-05-31
Use django's 400/500 custom page mechanism


JeffAbrahamson:
As part of this issue, all instances of raising `Http404`, `HttpResponseNotFound`, etc. should be remediated to present a propertly styled error page.

JeffAbrahamson:
@Shriukan33 Is this done?

PR #780, opened 2022-05-31
Add 404 page template

Shriukan33:
This provides a friendlier experience for 404 errors

Error message is provided : 
image

Fallback
image

Part of #779

### DIFF
--- a/transport_nantes/mailing_list/views.py
+++ b/transport_nantes/mailing_list/views.py
@@ -337,18 +337,14 @@ class NewsletterUnsubscriptionView(TemplateView):
 
         if email is None or send_record_id is None:
             logger.info(f"Token {token} is invalid")
-            raise Http404("Désolé, il semblerait qu'il y ait eu une erreur, "
-                          "veuillez vérifier que le lien est correct "
-                          "s'il vous plaît")
+            raise Http404()
 
         try:
             send_record = TopicBlogEmailSendRecord.objects.get(
                 pk=send_record_id)
         except ObjectDoesNotExist:
             logger.info(f"Send record ID : {send_record_id} not found")
-            raise Http404("Désolé, nous n'avons pas retrouvé le contenu "
-                          "demandé, vérifiez que votre lien est correct "
-                          "s'il vous plaît.")
+            raise Http404()
         context["send_record_mailing_list"] = \
             send_record.mailinglist.mailing_list_name
         context["user_email"] = email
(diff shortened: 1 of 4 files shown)


JeffAbrahamson:
The language is much too technical.  The link?  Magic powers?

First, most users absolutely don't care what went wrong, they just want their furry kittens or whatever.  So the first thing you have to say, and you have to say it clearly and succinctly starting with an `<h2>`, is that something didn't work out.

Also, note that this is a 4xx error, not a 5xx, so clearly the problem _will_ persist.  This is careless language on our part.  We can certainly invite users to contact us, but it's disingenuous to suggest that trying the same link later will change the results.

Finally, there's not much point in a button to return to the home page, because we don't know here where the user was or where thy meant to go.  The full menus are above, and the error message should be followed by project squares or a teaser and project squares or something like that that just invites the user to browse more.  No user (aside from a few geeks) arrive at this page and want to think about it.  Either we incite them visually to continue on the site or we lose them.  You have less than a second to make your case.  The time simply to _decide_ what the home page might be like and weather or not I want to go there is enough to send most users back to youtube or tiktok or something.

Shriukan33:
> The language is much too technical. The link? Magic powers?
> 
> First, most users absolutely don't care what went wrong, they just want their furry kittens or whatever. So the first thing you have to say, and you have to say it clearly and succinctly starting with an `<h2>`, is that something didn't work out. You get a couple words in the h2, and a tiny handful that stand alone as a first paragraph and that's it. If I can't read and understand it at a glance on my phone, you've lost me (statistically).
> 
> Also, note that this is a 4xx error, not a 5xx, so clearly the problem _will_ persist. This is careless language on our part. We can certainly invite users to contact us, but it's disingenuous to suggest that trying the same link later will change the results.
> 
> Finally, there's not much point in a button to return to the home page, because we don't know here where the user was or where thy meant to go. The full menus are above, and the error message should be followed by project squares or a teaser and project squares or something like that that just invites the user to browse more. No user (aside from a few geeks) arrive at this page and want to think about it. Either we incite them visually to continue on the site or we lose them. You have less than a second to make your case. The time simply to _decide_ what the home page might be like and weather or not I want to go there is enough to send most users back to youtube or tiktok or something.

Simplified the wording ! 

But I do disagree with the home page button, it's *really* common practice to do so : 
image
image
image
image
image
image
image
image
image
image
image
image


However, I do agree that the projects would fit nicely there ! 

I have pushed the normal version that is showing on the homepage, but I don't hate that the smaller version, see : 
image
Mobile : 
image


normal : 
image

The smaller one allow the desktop user to see without scrolling the titles. 

Wdyt ?

JeffAbrahamson:
OK, I'm convinced that such buttons have a place.

I like your text.  Grammar question: on est à plein d'autres (choses ?) intéressantes ?

Should we strike the word "vraiment", it makes it sound like we're discouraging contact.

What does "normal" mean?  Do you mean desktop, which is the minority viewport on the web compared to mobile (over 50% market share)?  ;-)

Shriukan33:
> I think this one looks best / is most readable: https://user-images.githubusercontent.com/70256364/171355444-6ce272a5-3447-4d16-aba7-8c14c9067272.png

This is the mobile version of the "narrow" version, with the blue section inside the message's div

Shriukan33:
> I just deployed origin/master to beta.
> 
> image

We have a case of "But it works on my local machine !"
image

PR #782, opened 2022-05-31
Add more logging to email sending

Shriukan33:
We got invalid token in GET request on press mailing list
In an attempt to reproduce the error, we log more details about
it.
Adding the token + email and ID to logs will allow us to better
reproduce the conditions that lead to incorrect token.

### DIFF
--- a/transport_nantes/asso_tn/utils.py
+++ b/transport_nantes/asso_tn/utils.py
@@ -96,8 +96,28 @@ def token_valid(encoded_timed_token, test_value_now=None):
     """
     try:
         timed_token = urlsafe_base64_decode(encoded_timed_token).decode()
-    except (TypeError, ValueError):
-        logger.info(f"token_valid: invalid token ({encoded_timed_token})")
+    except (TypeError, ValueError) as e:
+        # On May 31th 2022 :
+        #
+        # While sending a press release, it appears that at least 10 token out
+        # of 24 were invalid. We know this because we got 10 times a
+        # "token_valid: invalid token" error on GET requests to the
+        # 'mailing_list:press_subscription_management' url, seconds after the
+        # send.
+        # These errors all occured on the same try except block in
+        # asso_tn.urls.token_valid(), the one including :
+        #
+        # "timed_token = urlsafe_base64_decode(encoded_timed_token).decode()"
+        #
+        # Using the same tokens raises UnicodeError and once a ValueError.
+        # Attemps to create and decode tokens from the very same mails and
+        # send_record ids didn't raise any errors.
+        #
+        # In an attempt to reproduce and understand the error, we log more
+        # details about it for the next time it appears
+        # See issue https://github.com/transport-nantes/tn_web/issues/777
+        logger.info(f"token_valid: invalid token ({encoded_timed_token})\n"
+                    f"More info: {e}")
         return (None, 0)
     (the_rand_value, the_hash_rand, the_string_key,
      the_soon, the_int_key, the_hmac) = \
--- a/transport_nantes/topicblog/views.py
+++ b/transport_nantes/topicblog/views.py
@@ -301,6 +301,7 @@ class TopicBlogBaseList(LoginRequiredMixin, ListView):
 # to renderers.
 k_render_as_email = "render_as_email"
 
+
 class SendableObjectMixin:
     """ Define the sending by email behaviour for TopicBlog Objects """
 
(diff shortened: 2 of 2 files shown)


transport_nantes/asso_tn/utils.py
```diff
@@ -96,8 +96,12 @@ def token_valid(encoded_timed_token, test_value_now=None):
     """
     try:
         timed_token = urlsafe_base64_decode(encoded_timed_token).decode()
-    except (TypeError, ValueError):
-        logger.info(f"token_valid: invalid token ({encoded_timed_token})")
+    except (TypeError, ValueError) as e:
+        # We got invalid token in GET request on press mailing list
```
JeffAbrahamson, line 100: Please cite the issue and provide a date so that future readers have context.

transport_nantes/topicblog/views.py
```diff
@@ -459,6 +460,9 @@ def _set_email_context(
         context["token"] = \
             self.get_unsubscribe_token(recipient[0], send_record_id)
 
+        logger.info(f"Token {context['token']} is associated "
```
JeffAbrahamson, line 463: Please cite the issue and provide a date so that future readers have context.

JeffAbrahamson (changes requested):
This isn't a fix, it's the beginning of a strategy to understand more about a certain bug for the next time it happens.  As such, you not only need to explain why you do what you do, but you need to make sure that future readers of the code have a quick link to the issue, both from the code (missing) and from the commit message (strangely, also missing, but this is the citation that will give you a back-link from the issue to the commits, short of doing a `git log --grep`, which also needs the issue tagged).

There's _way_ too much state in your head that isn't in the code comments and isn't in the commit.

transport_nantes/asso_tn/utils.py
```diff
@@ -96,8 +96,18 @@ def token_valid(encoded_timed_token, test_value_now=None):
...
+    except (TypeError, ValueError) as e:
+        # On the 31/05/2022 :
+        # While sending a press release, it appears that at least 10 token out
+        # of 24 were invalid. We know this because we got 10 times a
+        # "token_valid: invalid token" error on GET requests to the
+        # 'mailing_list:press_subscription_management' url, seconds after the
+        # send.
+        # In an attempt to reproduce and understand the error, we log more
+        # details about it for the next time it appears
+        # See issue https://github.com/transport-nantes/tn_web/issues/777
```
JeffAbrahamson, line 108: This is good, but note that that issue doesn't have any useful information in it.

transport_nantes/asso_tn/utils.py
```diff
@@ -96,8 +96,18 @@ def token_valid(encoded_timed_token, test_value_now=None):
...
         timed_token = urlsafe_base64_decode(encoded_timed_token).decode()
-    except (TypeError, ValueError):
-        logger.info(f"token_valid: invalid token ({encoded_timed_token})")
+    except (TypeError, ValueError) as e:
+        # On the 31/05/2022 :
+        # While sending a press release, it appears that at least 10 token out
+        # of 24 were invalid. We know this because we got 10 times a
+        # "token_valid: invalid token" error on GET requests to the
+        # 'mailing_list:press_subscription_management' url, seconds after the
+        # send.
```
JeffAbrahamson, line 105: Good.  You might note all the things you learned yesterday whilst debugging this.  For example, that the error occurred in `token_valid()`, whatever else you learned.

You spent (if I understood correctly) over an hour on this.  Surely you learned more than you can summarise in two sentences!

Some of that summary can go in the issue, of course.  The code comment needs to get future devs up to speed on what the issue is, the issue can provide all the gory detail.

JeffAbrahamson (changes requested):

transport_nantes/asso_tn/utils.py
Shriukan33: Added my debug document to issue

transport_nantes/asso_tn/utils.py
```diff
@@ -96,8 +96,18 @@ def token_valid(encoded_timed_token, test_value_now=None):
     """
     try:
         timed_token = urlsafe_base64_decode(encoded_timed_token).decode()
-    except (TypeError, ValueError):
-        logger.info(f"token_valid: invalid token ({encoded_timed_token})")
+    except (TypeError, ValueError) as e:
+        # On the 31/05/2022 :
```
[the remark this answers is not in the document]
Shriukan33: Reworded

transport_nantes/asso_tn/utils.py
Shriukan33: Reworded + added document to issue

JeffAbrahamson:
Minor issues:
* The exceptions don't have spaces in the names: `UnicodeError` rather than `Unicode Error`, etc.
* Blank lines and indents make for more readable comments. ;-)

Otherwise, LGTM.

Shriukan33:
> Minor issues:
> 
>     * The exceptions don't have spaces in the names: `UnicodeError` rather than `Unicode Error`, etc.
> 
>     * Blank lines and indents make for more readable comments. ;-)
> 
> 
> Otherwise, LGTM.

Added some air in the comments, and removed space in error names :)

Shriukan33:
CodeQL check fail is not very helpful, it seems that the result analysis failed to do a PUT request to upload its results.
It might be all on GitHub's end. 

Given the nature of the PR, I doubt they would flag anything anyway, can I merge ?

JeffAbrahamson (approved):
LGTM.

PR #783, opened 2022-06-01
Capture and log click and open signals from SES

### LINKED ISSUE #778: Catch and log open and click SNS signals from AWS
This is still enabled on beta.

Once we log that we receive them, we should tell amazon no longer to send them.

Shriukan33:
We catch the click and open signals from SES and log them, so that
if we someday change our mind about using these features, the code is
already there.
It also tells future reader that we didn't forget to handle these,
simply we have reasons to not use them at this moment in time.

Closes #778

### DIFF
--- a/transport_nantes/topicblog/views.py
+++ b/transport_nantes/topicblog/views.py
@@ -27,7 +27,7 @@ from django.views.generic.list import ListView
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.urls import reverse, reverse_lazy
 from django.utils.html import strip_tags
-from django_ses.signals import (bounce_received, delivery_received,
+from django_ses.signals import (open_received, click_received,
                                 send_received)
 
 from asso_tn.utils import StaffRequired, make_timed_token, token_valid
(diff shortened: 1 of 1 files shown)


transport_nantes/topicblog/views.py
```diff
@@ -1084,3 +1084,44 @@ def send_received_handler(sender, mail_obj, send_obj, *args, **kwargs):
...
+def open_received_handler(sender, mail_obj, send_obj, *args, **kwargs):
+    """Handle AWS SES open_received notifications
+
+    AWS Receiver
+    This function will run when a open_received is received from
+    Amazon SES.
+    The signal is sent from django_ses' view.
+
+    We voluntarily ignore open_received notifications because
+    we already have the beacon view handling that for us.
```
JeffAbrahamson, line 1099: Good.  Maybe just add that (here and for click_recieved) that this signal introduces AWS tracking that we don't want because it triggers warnings from ublock origin and obfuscates links, and so whilst we catch and log it, we in principle have turned it off on the AWS console.

Shriukan33: https://github.com/transport-nantes/tn_web/blob/99795780de25a22910b1bf64edffb466f7f209ca/transport_nantes/topicblog/views.py#L1118

I already do that :) Only the click triggers the uBlock warning

Shriukan33: Added the quoted sentence to both docstrings

PR #786, opened 2022-06-01
Fix the 404.html path in settings so it doesn't require settings_local

Shriukan33:


JeffAbrahamson:
The gh review mechanism isn't making my life easy with a commit from another PR.  So,

* There's a trailing print.
* I'm still unclear why this template is so special that it can't otherwise be found, unlike every other template.

Shriukan33:
> The gh review mechanism isn't making my life easy with a commit from another PR. So,
> 
>     * There's a trailing print.
> 
>     * I'm still unclear why this template is so special that it can't otherwise be found, unlike every other template.

Removed the trailing print and the previous commit (will let the rebase fix the test) 

It's not special, it's more about how the template is "called" and searched, if I may use such terms.

When we want to use a template in our code, we give a string like "app/template.html". This is enabled because `APP_DIRS` is `True`

In the instance of `raise Http404()`, it calls a default view that simply looks for "`404.html`". Not "`transport_nantes/404.html`". 
When no app is given, the only folders the template engine will look at are the one mentioned in the `DIRS` list in `settings.py` TEMPLATES const, which for us is 
```python
TEMPLATES = [
    {
        'BACKEND': 'django.template.backends.django.DjangoTemplates',
        'DIRS': [
            os.path.abspath(__file__ + "/../templates/transport_nantes/"),
            ],
        'APP_DIRS': True,
        'OPTIONS': {
            'context_processors': [
                'django.template.context_processors.debug',
                'django.template.context_processors.request',
                'django.contrib.auth.context_processors.auth',
                'django.contrib.messages.context_processors.messages',
                'asso_tn.context_processors.role',
                'django.template.context_processors.media'
            ],
        },
    },
]
```
It tells django where to look when you don't mention any app and you are looking for a template. 
I think it's done so you can have projects-wide templates, `base_mobilitain.html` could have been in such directory

PR #787, opened 2022-06-01
Fix test that relied on live data to pass

Shriukan33:
Mock is guaranteed to be stable in time.

### DIFF
--- a/transport_nantes/press/tests.py
+++ b/transport_nantes/press/tests.py
@@ -1,3 +1,4 @@
+from unittest.mock import patch
 from django.test import TestCase, Client
 from .models import PressMention
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


transport_nantes/press/tests.py
```diff
@@ -175,23 +176,33 @@ def test_delete_view_press(self):
...
+            <!DOCTYPE html>
+            <html>
+                <head>
+                <meta property="og:title" content="{og_title}" />
+                <meta property="og:description" content="{og_description}" />
+                <meta property="og:image" content="{og_image}" />
+                </head>
+            </html>
+        """.encode("utf-8")
+        complete_url = "http://a-test-url.com/"
```
JeffAbrahamson, line 195: RFC 2606

JeffAbrahamson (changes requested):
I've some concerns about the tests (discussed in slack), but that's all going to have to change once this moves to celery.  So let's just change the RFC 2606 stuff and be done with it for now.

JeffAbrahamson:
(or maybe there's no issue)

JeffAbrahamson (approved):
LGTM.

PR #788, opened 2022-06-01
Enable matomo only in production

### LINKED ISSUE #759: Are we using the same matomo token on beta and in prod?

Shriukan33:
Closes #759


JeffAbrahamson:
LGTM (this one commit only).

PR #790, opened 2022-06-02
Add Unique constraint to prevent multiple sends to same recipient

### LINKED ISSUE #752: Marketing emails should unique on (slug, user)
Let's set a db constraint so we never send someone the same marketing mail.

Shriukan33:
If one is subbed to multiple newsletters and we want to send the same
article to the those newsletters, we need to make sure that we don't
send it to the same person more than once.
In the event we try to do so, we log the error and move on to next
recipient.

Closes #752


JeffAbrahamson (changes requested):
I think it just needs a test to confirm that for each of TB{Press,Email} sending once succeeds and trying again fails.

JeffAbrahamson (commented):
I think this rather critically needs to come after, and be rebased onto, the #800 changes, since the index only makes sense for marketing emails.

Shriukan33:
Rebased

transport_nantes/topicblog/tests_topic_blog.py
```diff
@@ -1048,3 +1048,82 @@ def test_send_email_form(self):
...
+        # Check if the emails have been sent (one for each subscribed user
+        self.assertEqual(len(mail.outbox), 2)
+
+        # Empty outbox
+        mail.outbox = []
+
+        response = self.admin_client.post(url, form_data)
+        # Redirect to success url
+        self.assertEqual(response.status_code, 302)
+        # Check if the emails were not sent twice
```
JeffAbrahamson, line 1072: Check if the emails were sent once and not twice.

Lest one read that you check twice. ;-)

transport_nantes/topicblog/tests_topic_blog.py
```diff
@@ -1048,3 +1048,82 @@ def test_send_email_form(self):
...
+                      args=[self.press_release.slug])
+
+        form_data = {
+            "mailing_list": 'the_mailing_list_token',
+            "confirmation_box": 'on'
+        }
+        response = self.admin_client.post(url, form_data)
+        # Redirect to success url
+        self.assertEqual(response.status_code, 302)
+        # Check if the emails have been sent (one for each subscribed user
```
JeffAbrahamson, line 1119: )

JeffAbrahamson (approved):
LGTM modulo requested comment changes.

PR #795, opened 2022-06-02
Add a button to send to ourselves any sendable TopicBlog object

### LINKED ISSUE #768: email and press need a "test me" button
Whether published or not, we should provide those who can view it to send themselves a test copy (i.e., send to user.email).  (Not on the public ViewLatest version, of course, just on the ViewOne version.)

Shriukan33:
The email is sent to logged-in user's email address.

Closes #768


JeffAbrahamson (commented):
Shouldn't this be using the transactional/marketing mechanism rather than introducing a whole new thing?

That is, I think it just says this:

> If I'm a TBEmail, then compose a TBTransactionalEmail to self.user.
> If I'm a TBPress, then compose a TBTransactionalPress to self.user.

Wdyt?

Shriukan33:
> Shouldn't this be using the transactional/marketing mechanism rather than introducing a whole new thing?
> 
> That is, I think it just says this:
> 
> > If I'm a TBEmail, then compose a TBTransactionalEmail to self.user.
> > If I'm a TBPress, then compose a TBTransactionalPress to self.user.
> 
> Wdyt?

It does use Transactional ! :) 

But the mechanism is a bit different compared to its usual use-case : We don't send the latest published item, we send precisely one, and this takes place with a form. 
Usually we pick a mailing list from a dropdown list, this isn't the case. 

In the case of logging, the flow is simpler because we simply render to one template without any complexity, just a text with a link. 
here we want the unsub links, and all the usual to work but without involving any mailing list (because it's transactional), but keeping the whole email / press rendering.

In the end, this isn't "a whole new thing", I really only : 
* Add a button ViewOne admin panel (it hides a form that gives us all the context we need to send the email)
* Add a view that processes the form and send the transactional email.

Shriukan33:
Notes from call : 
Create subclasses to SendRecordTransactional for Press, Email and Adhoc. 
AdHoc is for single use (send to oneself), it doesn't save information about the content, simply about the email itself. 
Email is for when you send a Transactional that renders an email, it saves the slug so we can find what content has been sent. 
Same for press.

JeffAbrahamson:
Is this awaiting action by @JeffAbrahamson or by @Shriukan33 ?

transport_nantes/topicblog/views.py
```diff
@@ -619,6 +625,105 @@ def form_valid(self, form):
                 send_record.save()
 
         return super().form_valid(form)
+
+
+class TopicBlogSelfSendView(PermissionRequiredMixin, LoginRequiredMixin,
+                            SendableObjectMixin, BaseFormView):
+    """Allow authorised users to send a not published sendable TopicBlog
```
JeffAbrahamson, line 632: PEP-8 etc: one-line, blank line, then body, all respecting line length limits.

JeffAbrahamson: and: that would be an "unpublished".  But, more generally, the point is that you can send by id whether it's published or not.

transport_nantes/topicblog/views.py
```diff
@@ -619,6 +625,105 @@ def form_valid(self, form):
...
+    object to themselves."""
+    # Maybe a more specific permission would be better ?
+    permission_required = 'topicblog.tbe.may_send'
```
JeffAbrahamson, line 635: Why `tbe` if this could be a `TBPress`?

transport_nantes/topicblog/views.py
```diff
@@ -619,6 +625,105 @@ def form_valid(self, form):
...
+        elif sent_object_class == "topicblogpress":
+            self.send_record_class = SendRecordTransactionalPress
+        else:
+            self.send_record_class = SendRecordTransactionalAdHoc
+
+        try:
+            send_record = self.send_record_class(
+                recipient=User.objects.get(email=email))
+            # Press and Email do require the slug of the object to send.
+            if hasattr(self.send_record_class, "slug"):
```
JeffAbrahamson, line 708: This is kludgy and fragile.  Perhaps use polymorphism.  You call some function that probably exists in the base class and that calls its `super()` to do the common stuff.

transport_nantes/topicblog/views.py
```diff
@@ -619,6 +625,105 @@ def form_valid(self, form):
...
+                # Add a fallback if no slug is saved on the object
+                if getattr(object_to_send, "slug",
+                           f"{sent_object_class}-{object_to_send.id}"):
+                    send_record.slug = object_to_send.slug
+
+            send_record.save()
+        except Exception as e:
+            logger.error(f"Failed to create send record : {e}")
+            # I don't think we have a debug level handler ?
```
JeffAbrahamson, line 717: In the code, you can't know what kind of log handlers will be provided.  You just log at the level that is appropriate.  It's an operational decision how to output log messages.

transport_nantes/topicblog/views.py
```diff
@@ -619,6 +625,105 @@ def form_valid(self, form):
...
+            logger.debug(f"Local variables : \n{locals()}")
+            return None
+
+        logger.info(f"Successfully created send record : Recipient: {email}, "
```
JeffAbrahamson, line 721: The word "successfully" doesn't add meaning but takes space.  "Created ..." is fine.

transport_nantes/topicblog/views.py
```diff
@@ -619,6 +625,105 @@ def form_valid(self, form):
...
+        Returns:
+        The send record for the given class and email address.
+        Or if an error prevents the creation of a send record, None.
+        """
+        logger.info(f"Creating send record for {sent_object_class} ")
+        sent_object_class = sent_object_class.lower()
+
+        # Maybe we should put the SendRecord in the model ?
+        # i.e. topicblogemail.transactional_send_record = SendRecordTransactionalEmail # noqa
+        if sent_object_class == "topicblogemail":
```
JeffAbrahamson, line 697: Use polymorphism, see below.  Or make `sent_object_class` be the actual name of the thing so that you can just call it.  Classes should never need to know a list of their possible derived classes.

JeffAbrahamson (changes requested):
Did you search for uses of `SendRecordTransactional`?
This breaks login, no (can't instantiate an abstract class)?

JeffAbrahamson:
In the mail that gets sent, you set
```<meta property="og:url" content="http://localhost:8000/tb/admin/self-send/" />
```

But that URL 500's (no `render_to_response`).  You also send an email that is the actual web page rather than the email that would have been sent.

transport_nantes/topicblog/views.py
Shriukan33: I don't understand your suggestion, how would I get the appropriate send record class if I don't mention it anywhere in the model ? :/ I see this is fragile, but couldn't come up with something else on the spot :/

Shriukan33:
> In the mail that gets sent, you set
> 
> But that URL 500's (no `render_to_response`). You also send an email that is the actual web page rather than the email that would have been sent.

Nothing in the block  ?

JeffAbrahamson:
> > In the mail that gets sent, you set
> > But that URL 500's (no `render_to_response`). You also send an email that is the actual web page rather than the email that would have been sent.
> 
> Nothing in the block ?

Yes, that was strange!

You send the web page rather than the sendable as mail.  And the web page has some buttons, and the buttons don't work when rendered in mail because cross-site.

transport_nantes/topicblog/models.py
```diff
@@ -310,6 +315,26 @@ def get_participating_field_names(self) -> set:
...
+        """Return appropriate class for sending transactional emails.
+
+        Return the class of the transactional SendRecord object associated
+        with this topicBlog Object.
+        """
+        if self.transactional_send_record_class is None:
+            raise NotImplementedError
+        try:
+            send_record_class = \
+                apps.get_model('topicblog',
```
JeffAbrahamson, line 329: Maybe it's worth a comment why we chose to use strings and then look them up in the global dictionary rather than letting the derived class just do its thing?

transport_nantes/topicblog/models.py
```diff
@@ -445,6 +484,34 @@ class SendRecordTransactional(SendRecordBase):
...
 
+class SendRecordTransactionalEmail(SendRecordTransactional):
+    """Represent a transactional email that sends a TopicBlogEmail."""
+
+    slug = models.SlugField(max_length=200)
+
+    def __str__(self):
+        return super().__str__() + self.slug
+
+    def fill(self, **kwargs):
```
JeffAbrahamson, line 495: Please tell me the relatively undescriptive name `fill()` is not ours. ;-)

JeffAbrahamson (changes requested):
You're still sending the underlying web page and not a mail, it seems.
image

and this is still in the headers:
image

In addition, the test button only appears in one-view and not on published items.  But it's just as important to be able to send a test just before sending to a live audience.

transport_nantes/topicblog/models.py
```diff
@@ -310,6 +315,26 @@ def get_participating_field_names(self) -> set:
...
 
+    def get_transactional_send_record_class(self) \
+            -> Type['SendRecordTransactional']:
+        """Return appropriate class for sending transactional emails.
+
+        Return the class of the transactional SendRecord object associated
+        with this topicBlog Object.
+        """
+        if self.transactional_send_record_class is None:
+            raise NotImplementedError
```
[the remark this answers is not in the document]
Shriukan33: https://github.com/transport-nantes/tn_web/blob/4ab8d0ddf93d06ae145db5b0c4786749f5292474/transport_nantes/topicblog/views.py#L696
Error is caught and logged by caller, it made more sense to place it there for me.

transport_nantes/topicblog/models.py
Shriukan33: We do the operation in that sense : We're given a TB object (be it Email or press) and from here, need to figure out its corresponding Transactional send record. 
Given that, I can't use the derived class because this is precisely what I'm looking for.

transport_nantes/topicblog/models.py
Shriukan33: I'm out of inspiration I'm affraid haha, fill_send_record seemed redundant send_record.fill_send_record(**kwargs) ...

Shriukan33: I'm out of inspiration I'm affraid haha, fill_send_record seemed redundant send_record.fill_send_record(**kwargs) ...

transport_nantes/topicblog/models.py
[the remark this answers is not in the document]
JeffAbrahamson: In this function, you do not have absolute control of the caller.  So it's still worth logging an error.

As you've probably seen, it's usually better to err by logging to much and then scaling back than having an operational problem you can't debug, have to hope it maybe happens again some day without embarrassing you too much. ;-)

JeffAbrahamson:
> > > > In the mail that gets sent, you set
> > > > But that URL 500's (no `render_to_response`). You also send an email that is the actual web page rather than the email that would have been sent.
> > > 
> > > 
> > > Nothing in the block ?
> > 
> > 
> > Yes, that was strange!
> > You send the web page rather than the sendable as mail. And the web page has some buttons, and the buttons don't work when rendered in mail because cross-site.
> 
> I still don't understand, how to you get this result ? the email I receive after pressing the button look fine to me

I'm seeing it in the console log.

JeffAbrahamson:
When I change the email backend to use SES, I actually get a different email than what gets sent to the console log!

transport_nantes/topicblog/models.py
JeffAbrahamson: This comment in the PR is notably not a comment in the code.

That said, this is a function called `get_transactional_send_record_class()` that lives in `TopicBlogObjectBase`.  So this presents two questions to me:

1.  Why does `TopicBlogObjectBase` even know what a send record is?  That knowledge is injected by sendable mixins in derived classes.  I think you're taking too free advantage of the fact that you have total control over the code and so putting stuff where it's most convenient at the moment rather than where it is most maintainable later.

2.  If this were the way to do it, derived classes could have methods that return whatever is right for them (possibly calling `super()`) and the base class' method would perhaps error because no such class is right for it.

3.  Isn't it the view that makes this association, not the model?

Shriukan33: I totally missed the fact that I can insert it inside the view. 
Welp, time to re-write, it would have been easier if I thought of that earlier ...

JeffAbrahamson:
LGTM.

Not blocking, but still need to be able to test-send published objects.

JeffAbrahamson:
This really, really needs to be in production ASAP.  What remains to do?

Shriukan33:
Tests failing, need to see why. Error codes seem to point to two things : One html element that is not where Selenium expects it to be ("Publier" button) and A sendrecord that is set to None.

Shriukan33:
Fixed, it was just that transactional send_record doesn't exist for tb item but I didn't check for existence in the get_context_data.
Simply added `if self.transactional_send_record_class:` and it's all good.

PR #802, opened 2022-06-09
Jeff updates pending benja return

JeffAbrahamson:
This is a collection of mostly comments (non-functional stuff).


transport_nantes/topicblog/models.py
```diff
@@ -397,12 +397,18 @@ class Meta:
...
     class StatusChoices(models.TextChoices):
-        # This is the initial state
+        # Newly created objects derived from SendRecordBase start life
+        # as PENDING.
         PENDING = 'PENDING', "Pending"
-        # Toggle to this state when Amazon SES accepts to send the email
+        # If SES indicates a non-permanent failure, the object enters
+        # state RETRY.  From this state, it should eventually
+        # transition to SENT or FAILED.
+        RETRYING = 'RETRY', 'Retry'
```
Shriukan33, line 408: It's not clear to me what events would trigger this state ?

transport_nantes/topicblog/views.py
```diff
@@ -305,26 +305,38 @@ def get_template_names(self):
...
             # Not implemented yet, but a published TBObject should be
-            # the only one with a publication date. For now it picks the
-            # most recent.
+            # the only one with a publication date. For now it picks
+            # the most recent.  **BUG**: assumes precisely one object
+            # returned, no order_by.
             publication_date__isnull=False
         )
         if len(tb_objects) > 1:
+            logger.error("Found multiple objects in class \"f{tb_slub}\" "
```
Shriukan33, line 333: tb_slub ?

transport_nantes/topicblog/views.py
```diff
@@ -305,26 +305,38 @@ def get_template_names(self):
...
+                         "because we didn't try to do otherwise.")
             raise ValueError(
                 f"There is more than one {self.base_model.__name__} with slug"
                 f" {tb_slug} and a not-null publication date")
         elif len(tb_objects) == 0:
+            logger.error(f"Failed to find requested email object in class \"f{tb_slub}\".")
```
Shriukan33, line 339: tb_slub

transport_nantes/topicblog/views.py
```diff
@@ -305,26 +305,38 @@ def get_template_names(self):
...
             slug=tb_slug,
             # Not implemented yet, but a published TBObject should be
-            # the only one with a publication date. For now it picks the
-            # most recent.
+            # the only one with a publication date. For now it picks
+            # the most recent.  **BUG**: assumes precisely one object
+            # returned, no order_by.
             publication_date__isnull=False
         )
         if len(tb_objects) > 1:
```
Shriukan33, line 332: Given the order_by and the fact that now multiple items might have a publication date (at least we expect that behavior to happen), checking the length for anything other than 0 seems useless

transport_nantes/topicblog/views.py
```diff
@@ -372,12 +384,18 @@ def _create_email_object(self, tb_object, context: dict,
...
-        for example to put recipients in BCC field or add attachments.
-        It also improves the performances by reusing the same connexion to SMTP
-        server.
+        """Create an EmailMultiAlternatives object.
+
+        To send emails to multiple persons, django send_mail function
+        isn't enough, we need to create a EmailMultiAlternatives
+        object to be able, for example to put recipients in BCC field
+        or add attachments.  It also improves the performances by
+        reusing the same connexion to SMTP server.  **BUG 1** We don't
```
Shriukan33, line 393: BUG1 : The docstringn is outdated in the sense we don't use SMTP anyway, so better delete it to not maintain confusion.

transport_nantes/topicblog/views.py
```diff
@@ -372,12 +384,18 @@ def _create_email_object(self, tb_object, context: dict,
...
+        actually use an SMTP server dirctly, we are using the SES API.
+        **BUG 2** We probably don't want ever to send to a list, since
```
Shriukan33, line 395: The email objects are created and sent individually, we don't add everyone to a single mail so this isn't a concern I think ?

transport_nantes/topicblog/views.py
```diff
@@ -504,6 +524,7 @@ def create_send_record(self,  slug: str, mailing_list: MailingList,
             recipient=recipient_user_object,
         )
         send_record.save()
+        logger.info(f"Created ")
```
Shriukan33, line 527: Bare "Created", out of context, is hard to figure what you created. Many things are, so you should probably mention that you created the send record. ( with its ID perhaps ?)

Shriukan33 (changes requested):

transport_nantes/topicblog/models.py
JeffAbrahamson: An example might be an informational message that the email can't be delivered but will keep trying.

transport_nantes/topicblog/views.py
JeffAbrahamson: Right, but currently we signal an error if it's great than one.  So we'll change the comment again (and remove the "**BUG**") when we fix this.

transport_nantes/topicblog/views.py
JeffAbrahamson: The variable is nonetheless `recipient_list`.  So we might not use it as a list but it's still something to fix so that it isn't a list.

transport_nantes/topicblog/views.py
JeffAbrahamson: Clearly something distracted me at just that moment...

transport_nantes/topicblog/models.py
Shriukan33: The informational message would come from AWS ? 
https://docs.aws.amazon.com/ses/latest/dg/event-publishing-retrieving-sns-contents.html#event-publishing-retrieving-sns-contents-top-level-json-object

transport_nantes/topicblog/views.py
Shriukan33: We need to loop through an iterable at some point, how do you define to who you're sending it to otherwise ?

transport_nantes/topicblog/models.py
JeffAbrahamson: Yup, e.g. `DeliveryDelay`.

transport_nantes/topicblog/views.py
JeffAbrahamson: The thing that calls create mail already loops over users.  The create mail function should simply create a send record, no?  Or maybe it's further down and it should create the mail object from the send record.  But it should surely match one-to-one with a send record (and so a single recipient).

transport_nantes/topicblog/models.py
Shriukan33: The DeliveryDelay event isn't handled by django-ses
image

JeffAbrahamson: So we should at some point soon.  That list should surely include all the events AWS documents it might send us.  Wdyt?

JeffAbrahamson: Oh, I see, it's not us, it's django-ses...

JeffAbrahamson: It's not urgent, but at some point our unknown notification handler should check for things django-ses didn't.  And maybe we should file an issue / offer a patch to django-ses?

JeffAbrahamson: SGTM.  In the meantime, maybe we add this enum entry whilst we are thinking of it and understand we can't currently transition into that state?

Shriukan33: Yes, adding it wont hurt anyway.

transport_nantes/topicblog/views.py
Shriukan33: Ha, I see the confusion : The mail object takes a list for the recipient argument, but that list is only 1 sized : 
image

PR #804, opened 2022-06-15
Sanitise send record classes' names

### LINKED ISSUE #800: Sanitise SendRecord class names
Some start with `SendRecord...`, others end with it.  This creates confusion on reading the code.

I propose the following hierarchy:
```
SendRecordBase
  SendRecordTransactional
    SendRecordTransactionalEmail
    SendRecordTransactionalPress
  SendRecordMarketing
    SendRecordMarketingEmail
    SendRecordMarketingPress
```

I'm not immediately clear how `SendRecordBase` or `SendRecordMarketing` could be used without a derived class, but `SendRecordTransactional` certainly could be (e.g., login).

Only `...Email` and `...Press` have slugs.

Otherwise, the main difference between transactional and marketing is that marketing send records should have unique indices on `(slug, recipient)` whilst transactional send records do not.  And otherwise the interfaces should assuredly be nearly identical: are we sending to a person or to a mailing list (just loops over members and sends to a person).

Sorting this is urgent, because we currently write send records to a common table, which prevents us from knowing where to look up the slug and so how to interpret it, how to preview it, etc.  And that, in turn, is prevening me from sending out press releases and reminders about the photo competition.

Note that the `TopicBlogEmailClicks` class needs to follow the above hierarchy, although the only difference may be the foreign key pointing to a different table.

Shriukan33:
Given that the branch changes existing models, I'd like to be sure there is a backup in case whatever goes wrong while migrating before deploying to beta. 

Closes #800


transport_nantes/mailing_list/views.py
```diff
@@ -15,7 +15,7 @@
 from django.urls import reverse_lazy
 from django.utils.crypto import get_random_string
 from django.views.generic import FormView, ListView, TemplateView
-from topicblog.models import TopicBlogEmailSendRecord
+from topicblog.models import SendRecordMarketingEmail
```
JeffAbrahamson, line 18: People can subscribe, unsubscribe, resubscribe, unsubscribe.  So this is a transactional mail, not a marketing mail.  Sending it to you once doesn't end the cycle.

transport_nantes/mailing_list/views.py
```diff
@@ -340,7 +340,7 @@ def get_context_data(self, **kwargs) -> dict:
             raise Http404()
 
         try:
-            send_record = TopicBlogEmailSendRecord.objects.get(
+            send_record = SendRecordMarketingEmail.objects.get(
```
JeffAbrahamson, line 343: Transactional

JeffAbrahamson (changes requested):
LGTM except for the one bit about transactional vs marketing.

I don't know why this distinction is hard, but I found it so at first, too, and now I don't understand how it can be confusing...

transport_nantes/mailing_list/views.py
Shriukan33: you unsubscribe from a mailing list. Sending mails to mailing list is saved as a SendRecordMarketing because it is Marketing right ?
So when one clicks on unsub, the page loads a page "Do you want to unsub from X mailing list ?" where X is extracted from the original send record. Which is Marketing. 
How is it transactional ?

PR #809, opened 2022-06-18
Jeff fixes

JeffAbrahamson:
I need to think of a way to test that the admin log handler is active.  It needs a 500.  You'd think that would be easy enough...


transport_nantes/transport_nantes/settings.py
```diff
@@ -169,44 +170,50 @@
...
             # 'handlers': ['console', 'mail_admins'],
-            'handlers': ['django.server'],
+            'handlers': ['django.server', 'mail_admins'],
             'level': 'INFO',
             'propagate': True,
         },
         'django': {
             # 'handlers': ['console', 'mail_admins'],
-            'handlers': ['console', 'file'],
+            'handlers': ['console', 'file', 'mail_admins'],
```
Shriukan33, line 206: Django's doc : 
>class AdminEmailHandler(include_html=False, email_backend=None, reporter_class=None)[[source]](https://docs.djangoproject.com/fr/4.0/_modules/django/utils/log/#AdminEmailHandler)[¶](https://docs.djangoproject.com/fr/4.0/ref/logging/#django.utils.log.AdminEmailHandler)
>
>   Ce gestionnaire envoie un courriel aux administrateurs du site ([ADMINS](https://docs.djangoproject.com/fr/4.0/ref/settings/#std-setting-ADMINS)) pour chaque message de journal qu’il reçoit.

Are you sure you want to receive an email for each log ? That seems like a lot.

transport_nantes/topicblog/views.py
```diff
@@ -1154,30 +1168,26 @@ def _extract_data_from_ses_signal(mail_obj: dict) -> Tuple[str, str, str]:
...
-    """Handle AWS SES send_received notifications
+    """Handle AWS SES send_received notifications.
 
     AWS Receiver
     This function will run when a send_received is received from
     Amazon SES.
     The signal is sent from django_ses' view.
     """
-    logger.info("send_received received !")
+    logger.info("Received send_receive.")
```
Shriukan33, line 1178: The signal's name is send_received

transport_nantes/topicblog/views.py
```diff
@@ -1154,30 +1168,26 @@ def _extract_data_from_ses_signal(mail_obj: dict) -> Tuple[str, str, str]:
...
+    logger.info(f"  {send_record_class} ID : {send_record_id}")
     if send_record_class and send_record_id:
         try:
             send_record = send_record_class.objects.get(pk=send_record_id)
             send_record.status = "SENT"
             send_record.send_time = datetime.now(timezone.utc)
             send_record.aws_message_id = aws_message_id
             send_record.save()
-            logger.info(
-                f"\nSES confirmed email sending\n"
```
Shriukan33, line 1178: Why would you remove details from the logs ? I think that SES confirmed the sending is easy to read and understand, while being important information.

transport_nantes/transport_nantes/settings.py
JeffAbrahamson: If I didn't make a mistake, it is configured only for ERROR.

Shriukan33: Ok, I wasn't sure because some of these loggers are INFO level.

transport_nantes/topicblog/views.py
JeffAbrahamson: Good catch, fixed.

transport_nantes/topicblog/views.py
JeffAbrahamson: From the diff (`d3a3076..211ee23`):
```
@@ -1154,19 +1168,18 @@ def _extract_data_from_ses_signal(mail_obj: dict) -> Tuple[str, str, str]:
 
 @receiver(send_received)
 def send_received_handler(sender, mail_obj, send_obj, *args, **kwargs):
-    """Handle AWS SES send_received notifications
+    """Handle AWS SES send_received notifications.
 
     AWS Receiver
     This function will run when a send_received is received from
     Amazon SES.
     The signal is sent from django_ses' view.
     """
-    logger.info("send_received received !")
+    logger.info("Received send_receive.")
     aws_message_id, send_record_class, send_record_id = \
         _extract_data_from_ses_signal(mail_obj)
-    logger.info(
-        f"\nsend_received signal received for message {aws_message_id}\n"
-        f"SendRecord class : {send_record_class} ID : {send_record_id}")
+    logger.info(f"Received send_received for {aws_message_id}")
+    logger.info(f"  {send_record_class} ID : {send_record_id}")
     if send_record_class and send_record_id:
         try:
             send_record = send_record_class.objects.get(pk=send_record_id)
@@ -1174,10 +1187,7 @@ def send_received_handler(sender, mail_obj, send_obj, *args, **kwargs):
             send_record.send_time = datetime.now(timezone.utc)
             send_record.aws_message_id = aws_message_id
             send_record.save()
-            logger.info(
-                f"\nSES confirmed email sending\n"
-                f"SR.class : {send_record_class.__name__}\n"
-                f"SR.id :{send_record.pk})")
+            logger.info(f"send_received : {send_record_class.__name__} id={send_record.pk}")
         except Exception as e:
             logger.error(
                 f"Error while updating send_record : {e}")

```
I'm not clear what data you see not being logged anymore.

Shriukan33: Goes from : 
```
logger.info(
-                f"\nSES confirmed email sending\n"
-                f"SR.class : {send_record_class.__name__}\n"
-                f"SR.id :{send_record.pk})")
```

to 

`logger.info(f"send_received : {send_record_class.__name__} id={send_record.pk}")`

I was referring to first sentence

JeffAbrahamson: Ah, right.  But we just logged a moment ago "Received send_received.".

JeffAbrahamson: And, critically, this makes the log message single-line (so more automation-compatible).

Shriukan33: Fair enough

Shriukan33 (approved):

JeffAbrahamson:
Perhaps we should remove the comment about not creating users until they respond to mail?  ;-)

Otherwise, LGTM.

PR #811, opened 2022-06-20
Benja-django-dbbackup

Shriukan33:
Draft while it's not been tested at least on beta

This PR adds django-dbbackup, that allows us to dump and restore using management commands. 

Doc : 
https://django-dbbackup.readthedocs.io/en/master/index.html

It can be combined to django-cron but as it's the only command we'd run for now, it's useless to set it up.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,10 +20,10 @@ certifi==2019.11.28
 cffi==1.15.0
 chardet==3.0.4
 charset-normalizer==2.0.6
+click==8.1.3
 click-didyoumean==0.3.0
 click-plugins==1.1.1
 click-repl==0.2.0
-click==8.1.3
 colorama==0.4.3
 configparser==5.2.0
 contextlib2==0.6.0
@@ -34,7 +34,7 @@ decorator==4.4.1
 Deprecated==1.2.13
 distlib==0.3.0
 distro==1.4.0
-dj-stripe==2.5.1
+Django==4.0.4
 django-amazon-ses==4.0.1
 django-countries==7.2.1
 django-crispy-forms==1.14.0
@@ -42,7 +42,7 @@ django-debug-toolbar==3.2.4
 django-ranged-response==0.2.0
 django-ses==3.0.1
 django-simple-captcha==0.5.13
-Django==4.0.4
+dj-stripe==2.5.1
 docutils==0.15.2
 folium==0.12.1
 future==0.18.2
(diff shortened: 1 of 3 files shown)


Shriukan33:
> This is a clever package, but I'm wondering how we'd want to integrate it in a production context. It feels a bit amateur to me. Consider, for example, the phrase "backups can lie around and be forgotten, so wise to encrypt". Let's think through how this operating philosophy works with the GDPR and other legal obligations.
> 
> Some questions I haven't resolved yet:
> 
>     * How do we manage key management securely?  (Is salt good enough for our needs here?)
> 
>     * How can we make it structurally impossible to drop the production database by accident?
> 
>     * How do we manage logging so that we know every action that is taken with this module?
> 
>     * How do we version / clean old backups to avoid having hundreds or thousands of backups sitting around eating space?
> 
>     * How do we make sure that we can control which developers have access to the backups (GDPR) -- salt, maybe.

I'm going to quote what I can find on the doc !

>     * How do we manage key management securely?  (Is salt good enough for our needs here?)

Key would probably be stored on the infra. Creation is done with PGP https://django-dbbackup.readthedocs.io/en/master/configuration.html#pgp

>     * How can we make it structurally impossible to drop the production database by accident?

I don't think this is specific to this lib ? This as an every-moment concern. 

>     * How do we manage logging so that we know every action that is taken with this module?

Reading dbbackup source code, the lib logs with this logger : 
`venv/lib/python3.8/site-packages/dbbackup/management/commands/_base.py`
```python
logger = logging.getLogger('dbbackup.command')
```
So much like django-ses, we'd have to add a logger and probably it would be best to give it a separate file. (tn-web_dbbackup.log ?)

>     * How do we version / clean old backups to avoid having hundreds or thousands of backups sitting around eating space?

I don't think the lib handles this sort of operation (deleting backups), I think this would be doable with a bash script and a cronjob though. This is more your area of expertise than mine :sweat_smile: 

 >     * How do we make sure that we can control which developers have access to the backups (GDPR) -- salt, maybe.

Backups are stored on the server, and encrypted with GPG. 
Note that the lib offers to save on s3/Google Cloud/Dropbox (https://django-dbbackup.readthedocs.io/en/master/storage.html), so I think the access control is granular enough to satisfy the GDPR ?

I mean, I'm not especially defending the lib, I've just seen this as  a solution to a problem you've been wanting to solve for a while :slightly_smiling_face:

JeffAbrahamson:
I'm going to close this, because, while it's a possible direction to take, I don't believe it's what we should do (based on my own experience).  It leaves too many open ends to my taste.

PR #812, opened 2022-06-20
Add rollbar

### LINKED ISSUE #339: integrate rollbar on free tier

Shriukan33:
This sets up rollbar monitoring. 

Closes #339

### DIFF
--- a/transport_nantes/transport_nantes/settings.py
+++ b/transport_nantes/transport_nantes/settings.py
@@ -13,6 +13,7 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 from django.urls import reverse_lazy
 from . import settings_local
 import sys
+import os
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/3.0/howto/deployment/checklist/
@@ -75,6 +76,11 @@ MIDDLEWARE = [
     'asso_tn.middleware.sessionCookie.SessionCookieMiddleWare',
     'utm.middleware.utm.UtmMiddleware',
 ]
+
+if ROLE in ("beta", "production"):
+    MIDDLEWARE += [
+        'rollbar.contrib.django.middleware.RollbarNotifierMiddleware',
+    ]
 
 ROOT_URLCONF = 'transport_nantes.urls'
 
@@ -275,3 +281,12 @@ CELERY_TIMEZONE = "UTC"
 CELERY_TASK_TRACK_STARTED = True
 CELERY_TASK_TIME_LIMIT = 30 * 60
 CELERY_BROKER_URL = "amqp://localhost"
+
+if ROLE in ("beta", "production"):
+    ROLLBAR = {
+        'access_token': settings_local.ROLLBAR_ACCESS_TOKEN,
+        'environment': 'dev' if DEBUG else 'production',
+        'root': os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+    }
+    import rollbar
+    rollbar.init(**ROLLBAR)
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,10 +20,10 @@ certifi==2019.11.28
 cffi==1.15.0
 chardet==3.0.4
 charset-normalizer==2.0.6
+click==8.1.3
 click-didyoumean==0.3.0
 click-plugins==1.1.1
 click-repl==0.2.0
-click==8.1.3
 colorama==0.4.3
 configparser==5.2.0
 contextlib2==0.6.0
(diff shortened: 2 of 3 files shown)


transport_nantes/transport_nantes/settings.py
```diff
@@ -13,6 +13,8 @@
 from django.urls import reverse_lazy
 from . import settings_local
 import sys
+import os
```
JeffAbrahamson, line 16: Is this needed?  See below.

transport_nantes/transport_nantes/settings.py
```diff
@@ -272,3 +275,10 @@
 CELERY_TASK_TRACK_STARTED = True
 CELERY_TASK_TIME_LIMIT = 30 * 60
 CELERY_BROKER_URL = "amqp://localhost"
+
+ROLLBAR = {
+    'access_token': settings_local.ROLLBAR_ACCESS_TOKEN,
+    'environment': 'dev' if DEBUG else 'production',
+    'root': os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
```
JeffAbrahamson, line 282: Why not `settings_local.BASE_DIR`?

transport_nantes/transport_nantes/settings.py
```diff
@@ -73,6 +75,7 @@
     'asso_tn.middleware.default_context.DefaultContextMiddleware',
     'asso_tn.middleware.sessionCookie.SessionCookieMiddleWare',
     'utm.middleware.utm.UtmMiddleware',
+    'rollbar.contrib.django.middleware.RollbarNotifierMiddleware',
```
JeffAbrahamson, line 78: You don't comment that this must be last.  You also add always even though we don't want to run rollbar in dev (uses credits).

Should this perhaps instead be
```
if ROLE in ("beta", "production"):
    # Rollbar wants to be last.  And we don't want it in dev.
    MIDDLEWARE.append('rollbar.contrib.django.middleware.RollbarNotifierMiddleware')
```

transport_nantes/transport_nantes/settings.py
```diff
@@ -13,6 +13,8 @@
 from django.urls import reverse_lazy
 from . import settings_local
 import sys
+import os
+import rollbar
```
JeffAbrahamson, line 17: Perhaps import at bottom just before calling init?

transport_nantes/transport_nantes/settings.py
```diff
@@ -272,3 +275,10 @@
 CELERY_TASK_TRACK_STARTED = True
 CELERY_TASK_TIME_LIMIT = 30 * 60
 CELERY_BROKER_URL = "amqp://localhost"
+
+ROLLBAR = {
```
JeffAbrahamson, line 279: Make this conditional?
```
if ROLE in ("beta", "production"):
    ROLLBAR = {
        'access_token': settings_local.ROLLBAR_ACCESS_TOKEN,
        'environment': 'development' if DEBUG else 'production',
        'root': settings_local.BASE_DIR,
    }
    import rollbar
    rollbar.init(**ROLLBAR)
```

JeffAbrahamson (changes requested):

transport_nantes/transport_nantes/settings.py
Shriukan33: Because settings_local.py states :

> # NOT AVAILABLE IN PRODUCTION
> # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
> BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))

base dir is absent in infra afaik

transport_nantes/transport_nantes/settings.py
Shriukan33: Good catch

transport_nantes/transport_nantes/settings.py
Shriukan33: Yes as base_dir isn't in prod (see below)

transport_nantes/transport_nantes/settings.py
Shriukan33: Imports traditionnaly are on top of file, why do you want to move it to bottom ?

transport_nantes/transport_nantes/settings.py
Shriukan33: Done

transport_nantes/transport_nantes/settings.py
Shriukan33: Moved it to bottom anyway, saves an import in dev, yay

JeffAbrahamson: Yup, that was the only reason. ;-)

JeffAbrahamson:
Modulo the conversations, LGTM.

Shriukan33:
> Modulo the conversations, LGTM.

What conversation ? Do I still have something to change ?

JeffAbrahamson:
> > Modulo the conversations, LGTM.
> 
> What conversation ? Do I still have something to change ?

Nope, I meant the conversations we just had and closed.

PR #814, opened 2022-06-21
Benja 679 fill default template

### LINKED ISSUE #679: Fill in default template for all TBObject's
There should be no "select a template", just provide the default one if none is selected.

### LINKED ISSUE #743: TBObjectEdit repeats template field
At this page

    https://www.mobilitains.fr/tb/admin/t/new/

there is a field called `template_name` that duplicates the field called `template`.

Shriukan33:
closes #679
closes #743 

Removes the extra "template name" in tbitems (#743) and pre-select first template (which is usually the default one) if none other is provided.


transport_nantes/topicblog/forms.py
```diff
@@ -5,7 +5,45 @@
...
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.set_initial_template()
+
+    def get_template_list(self) -> list:
+
+        templates = self.instance.template_config
+        template_list = \
+            [(None, "Selectionnez un template ...")]
```
JeffAbrahamson, line 27: The first element in the list is now being selected, which is good.
But this is a non choice: what could it possibly mean for someone to choose it?

transport_nantes/topicblog/forms.py
```diff
@@ -5,7 +5,45 @@
...
+            template_list.append((template, value["user_template_name"]))
+
+        return template_list
+
+    def set_initial_template(self):
+        template_list = self.get_template_list()
+        if self.instance.template_name == "" and len(template_list) >= 2:
+            self.fields['template_name'] = forms.ChoiceField(
+                choices=template_list,
+                initial=template_list[1][0],
```
JeffAbrahamson, line 38: Do I correctly understand that we are arbitrarily picking the first element of the list as the default?
* Shouldn't you filter by the `active` flag?
* Might it not be more appropriate to add a `default_choice: True` and use that (or the first instance of that) if it exists?
* Is it possible that thie template config has become complex enough that it should be a class (a second issue, second PR, no doubt)?

JeffAbrahamson (changes requested):

transport_nantes/topicblog/forms.py
Shriukan33: All good suggestions ! The third one (making it class) is what we originally went for before deciding a dict was better. 
We can open an issue but because this isn't really blocking for now, I suggest we stick with that and work on more productive issues.

I'll rework the template_config so it uses active and default_choice

transport_nantes/topicblog/forms.py
JeffAbrahamson: When I do pull down the dropdown, I see "Selectionnez un template ..." in the dropdown.  It's not selected, but why is it even there?  It's never the right thing to select.

transport_nantes/topicblog/forms.py
JeffAbrahamson: IIRC, it was originally a model.  I'm proposing it be a class so that it can control its interface and help us avoid errors, have methods like `is_active()` instead of all the code having to know that we stored the data at just this one place in this increasingly complex dict.

JeffAbrahamson: Agreed, it's not blocking for now.

JeffAbrahamson:
What remains to do here?

Shriukan33:
> What remains to do here?

You need to review it, I applied changes you requested in your previous review

JeffAbrahamson:
The commit comment is

> Update handoff time on sendrecords

but I think you mean something like

> Set handoff_time on send records when handing off to SES.

JeffAbrahamson:
Do you also need to update `send_activation()` in `authentication/views.py`?

Shriukan33:
> The commit comment is
> 
> > Update handoff time on sendrecords
> 
> but I think you mean something like
> 
> > Set handoff_time on send records when handing off to SES.

Wrong PR , but noted will change

JeffAbrahamson:
> Wrong PR , but noted will change

How did I do that?  <facepalm>

JeffAbrahamson:
I'm confused.  The commit comment on the head commit is
```
    Pre-select default choice for new items and refactor
    
    We previously had to select a template even if only one was available.
    
    Closes #679
```

But you create a new ModelForm that doesn't set `self.model`.  And that still contains
```
            template_list = \
                [(None, "Selectionnez un template ...")]
```
and we want for that non-item never to be present.

JeffAbrahamson (approved):
LGTM.

Shriukan33:
Rebased on master, Mickael's PR changed a test and it did conflict, easy merge. Will wait til tests go green and go

PR #815, opened 2022-06-21
Remove references to "servable"

### LINKED ISSUE #462: TBItem quick issues

Shriukan33:
We don't use this attribute anymore. Its presence in templates showed
buttons that had no functionality.

Closes #462

### DIFF
--- a/transport_nantes/topicblog/forms.py
+++ b/transport_nantes/topicblog/forms.py
@@ -13,7 +13,7 @@ class TopicBlogItemForm(ModelForm):
     class Meta:
         model = TopicBlogItem
         # Admins can still edit those values
-        exclude = ('item_sort_key', 'servable', 'user',
+        exclude = ('item_sort_key', 'user',
                    'publication_date', 'first_publication_date', 'publisher')
 
     def __init__(self, *args, **kwargs):
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -151,8 +151,12 @@ class TopicBlogObjectBase(models.Model):
         else:
             return f'{str(self.title)} - ID : {str(self.id)} (NO SLUG)'
 
-    def get_servable_status(self):
-        """Return True if page is user visible, False otherwise."""
+    def get_servable_status(self) -> bool:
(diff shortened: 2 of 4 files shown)


JeffAbrahamson (changes requested):
Is this function still correct?
```
    def get_servable_status(self):
        """Return True if page is user visible, False otherwise."""
        if self.publication_date is None or \
                datetime.now(timezone.utc) < self.publication_date:
            return False
        return True
```

Shriukan33:
> Is this function still correct?
> 
> ```
>     def get_servable_status(self):
>         """Return True if page is user visible, False otherwise."""
>         if self.publication_date is None or \
>                 datetime.now(timezone.utc) < self.publication_date:
>             return False
>         return True
> ```

Yes it is :)
We use that to check that the item we're selecting is indeed servable (has a publication date that is inferior to now)
It comes after we filter and order. It's a very edge case, as we don't usually set a publication date to future (can only be done in admin).

JeffAbrahamson:
> > Is this function still correct?
> > ```
> >     def get_servable_status(self):
> >         """Return True if page is user visible, False otherwise."""
> >         if self.publication_date is None or \
> >                 datetime.now(timezone.utc) < self.publication_date:
> >             return False
> >         return True
> > ```
> 
> Yes it is :) We use that to check that the item we're selecting is indeed servable (has a publication date that is inferior to now) It comes after we filter and order. It's a very edge case, as we don't usually set a publication date to future (can only be done in admin).

Maybe just deserves a comment to indicate that extra dependency, that this isn't enough to determine servability, that there are other constraints?

Shriukan33:
Added a comment in the docstring

JeffAbrahamson (approved):
LGTM.

PR #816, opened 2022-06-21
Add a delete button for never published tb objects

Shriukan33:
this CRUD operation has been lacking due to our maintain content policy, however I think we can make an exception for drafts and errors, which can pollute the object list pretty bad. 

image

Objects without a first_publication date might be errors, tries or
drafts.
This button allows users with permissions to delete their own objects,
and with elevated permissions to delete all objects they didn't author, if they never been published.

Objects with a first_publication_date aren't offered this button. This condition is checked both at page load and in POST.


JeffAbrahamson:
There might be a need for this in case one adds something confidential, but presumably one wouldn't have published it yet, so the risk is small.

I'd rather this be automated.  @mcmikashi started work on something (PR #729) but it never got finished.

Shriukan33:
> There might be a need for this in case one adds something confidential, but presumably one wouldn't have published it yet, so the risk is small.
> 
> I'd rather this be automated. @mcmikashi started work on something (PR #729) but it never got finished.

It's only to delete stuff we don't want to keep and that has been never published, so same idea as @mcmikashi 's PR, but it's done manually, case per case. 

I'd say both can coexist, @mcmikashi 's system for auto-cleanup, and this one if you want to tidy things on the spot and not wait for the auto delete.

JeffAbrahamson:
I disagree.  Every button demands attention.  Constraints are important in navigating the world.  As developers, we create guides for navigating.

In addition, if we create a button for deleting objects, people will sometimes do so even when they don't need to.  They will also inevitably ask to have manual control and so to be able to override the automatic control.  How far are you willing to go in developing and maintaining this nascent feature set?  How much money and/or time is it worth?  Can you think of a budget?

Are you solving a problem we really have, if you imagine that unpublished TBObjects will be deleted after a few weeks automatically?

PR #819, opened 2022-06-22
Log what we do in photo upload

JeffAbrahamson:
Successful photo uploads were silent, we had no trace in the logs.
Failed photo uploads logged an info message and not an error.

This commit fixes those two issues.


Shriukan33 (approved):
LGTM

Shriukan33:
Duplicate commit in #820

PR #823, opened 2022-06-23
Benja 822 photo contest improvements

### LINKED ISSUE #822: Improve photo contest form
- [x] Add more logs to photo contest
- [x] Display the uploaded photo on thank you page
- [x] Improve layout so submitting looks more obvious

Shriukan33:
closes #822 

* Added some logs to GET (form and confirmation)
* Display the photo after submission
* Make the submit button more obvious

### DIFF
--- a/transport_nantes/photo/templates/photo/photoentry_form.html
+++ b/transport_nantes/photo/templates/photo/photoentry_form.html
@@ -30,7 +30,7 @@
                     <li>J'accepte que Les Mobilitains me contactent à propos de
                     ma participation</li>
                     </ul>
-                <button class="btn donation-button" type="submit">Valider</button>
+                <button class="btn donation-button p-3" type="submit">Soumettre ma photo !</button>
             </form>
         </div>
     </div>
--- a/transport_nantes/photo/views.py
+++ b/transport_nantes/photo/views.py
@@ -21,6 +21,10 @@ class UploadEntry(LoginRequiredMixin, CreateView):
 
     success_url = reverse_lazy('photo:confirmation')
 
+    def get(self, request, *args, **kwargs):
+        logger.info(f"UploadEntry.get() from {request.user}")
+        return super().get(request, *args, **kwargs)
+
     def post(self, request, *args, **kwargs):
         """
         Override the post method to add the user to the form
@@ -49,3 +53,7 @@ class Confirmation(TemplateView):
     Confirmation page after successful submit of a PhotoEntry
     """
     template_name = 'photo/confirmation.html'
+
+    def get(self, request, *args, **kwargs):
+        logger.info(f"Confirmation.get() from {request.user}")
+        return super().get(request, *args, **kwargs)
--- a/transport_nantes/photo/views.py
+++ b/transport_nantes/photo/views.py
@@ -1,9 +1,12 @@
 """Application to manage a photo competition."""
 import logging
+from django.http import HttpResponseRedirect
 from django.urls import reverse_lazy
 from django.views.generic import (TemplateView, CreateView)
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.core.exceptions import ObjectDoesNotExist
+
+from asso_tn.utils import make_timed_token, token_valid
 from .forms import PhotoEntryForm
 from .models import PhotoEntry
 from mailing_list.models import MailingList
@@ -47,6 +50,16 @@ class UploadEntry(LoginRequiredMixin, CreateView):
         else:
             return self.form_invalid(form)
 
+    def form_valid(self, form):
+        """
+        Override the form_valid method to add the user's submission to session
+        """
+        self.object = form.save()
+        encoded_object_id = make_timed_token(
+            string_key="", int_key=self.object.id, minutes=60*24*30)
+        self.success_url += f"?submission={encoded_object_id}"
+        return HttpResponseRedirect(self.get_success_url())
+
 
 class Confirmation(TemplateView):
     """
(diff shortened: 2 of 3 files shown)


transport_nantes/photo/templates/photo/photoentry_form.html
```diff
@@ -30,7 +30,7 @@ <h1>Formulaire de participation à l'Opération piéton</h1>
                     <li>J'accepte que Les Mobilitains me contactent à propos de
                     ma participation</li>
                     </ul>
-                <button class="btn donation-button" type="submit">Valider</button>
+                <button class="btn donation-button p-3" type="submit">Soumettre ma photo !</button>
```
JeffAbrahamson, line 33: good idea!

transport_nantes/photo/views.py
```diff
@@ -21,6 +21,10 @@ class UploadEntry(LoginRequiredMixin, CreateView):
 
     success_url = reverse_lazy('photo:confirmation')
 
+    def get(self, request, *args, **kwargs):
+        logger.info(f"{request.user} landed on photo contest page")
```
JeffAbrahamson, line 25: Careful, this is vague.  When reading thousands of lines of logs, will you really know what this means?  Will another dev who hasn't worked on this?  What I mean is that your language in the log message is about the user experience and not about the code.  So it's not enough to know the code, you need to know, for example, that we use this code for a photo competition, and, morever, at some future date when we use it for something totally different (as always happens with the code that is not consigned to the dust bin), that we _used_ to use it for a photo competition.

The log message includes the module and line of code that emits it now, and this is really key to understanding logs and tracing back to code.

Something more functional makes more sense to me:
```f"UploadEntry.get() from {request.user}"```

File: transport_nantes/photo/views.py [hunk 3]
```diff
@@ -49,3 +53,14 @@ class Confirmation(TemplateView):
     Confirmation page after successful submit of a PhotoEntry
     """
     template_name = 'photo/confirmation.html'
+
+    def get(self, request, *args, **kwargs):
+        logger.info(f"{request.user.email} landed on confirmation page.")
```
JeffAbrahamson, line 58: Same comment about comment.

transport_nantes/photo/views.py
```diff
@@ -49,3 +53,14 @@ class Confirmation(TemplateView):
...
+        return super().get(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        last_submitted_photo = \
+            PhotoEntry.objects.filter(user=self.request.user).last()
```
JeffAbrahamson, line 64: Doesn't this create a race condition if the user is submitting in two windows simultaneously?

Can we put the data in the context when we process the upload POST and so have it available here in the context?

JeffAbrahamson (changes requested):

transport_nantes/photo/views.py
```diff
@@ -43,9 +50,38 @@ def post(self, request, *args, **kwargs):
...
+        logger.info(f"Confirmation.get() from {request.user}")
+        return super().get(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        last_submitted_photo = None
+        submission_token = self.request.GET.get("submission")
+        photo_id, _ = token_valid(submission_token)
+
+        if photo_id.isnumeric():
```
JeffAbrahamson, line 80: Why `is_numeric()`?  If it's not valid, you'll get `None`.

This is a weird abuse of the timed token, though.  If you need to recover an int and not a string, then just pass that.  If you don't need the string, pass an empty string.  Casting back and forth presents too many opportunities for errors or things going wrong that lead to a string that can be converted to an int and you get what you asked for but not what you wnated.

JeffAbrahamson (changes requested):
One comment below, one here.

On submitting, I see this:
image

Clearly it's the image I uploaded, but maybe we could say it in a fun and friendly way instead of just kind of throwing it there and heading back into our cave? ;-)

In addition, this kinda makes it look like "par ici" goes with the photo and not with what came before.  White space is really important to leading eyes and brains.

transport_nantes/photo/views.py
Shriukan33: Fixed, passed through the int_key instead

JeffAbrahamson:
> I changed the confirmation template :
> 
> image
> 
> Thoughts ?

Looks generally good.

A few thoughts:
* For the user, unless this is my French playing games with me, this isn't really "my most recent contribution", this is "the photo I just submitted".  A subtlety, but it just seems less abstract.  Thoughts?
* Is "envoyer" the correct word given that we're talking about "submission"?
* Perhaps increase padding below the button or above the next text, say `pt-5` and see if that looks better, more separated?
* Let's make the image smaller, say 50% screen width in a div that provides left and right margins and with `img-fluid img-thumbnail` to give it a border and make it say "I'm not part of the site, I'm your submission".  (Disclaimer: I don't remember what `img-fluid` does, maybe we don't need that.)

Wdyt?

transport_nantes/photo/views.py
JeffAbrahamson: You also pass the string as `None` instead of empty ("").  But the validation function returns `None` on invalid.  So even though you don't look at the string on return (not looking at error return?), you can't distinguish between a valid token with `int_key = 0` and an invalid token.

Shriukan33:
> Looks generally good.
> 
> A few thoughts:
> 
>     * For the user, unless this is my French playing games with me, this isn't really "my most recent contribution", this is "the photo I just submitted".  A subtlety, but it just seems less abstract.  Thoughts?

I thought the page would say "soumission" a bit too much, so I looked for synonyms, but you're right that it would be more precise. 

> 
>     * Is "envoyer" the correct word given that we're talking about "submission"?

Same thought about "Soumission", can replace with "Soumettre d'autres photos"

>     * Perhaps increase padding below the button or above the next text, say `pt-5` and see if that looks better, more separated?
>     * Let's make the image smaller, say 50% screen width in a div that provides left and right margins and with `img-fluid img-thumbnail` to give it a border and make it say "I'm not part of the site, I'm your submission".  (Disclaimer: I don't remember what `img-fluid` does, maybe we don't need that.)

Images in Bootstrap are made responsive with .img-fluid. This applies max-width: 100%; and height: auto; to the image so that it scales with the parent element.
But sure I'll make it a bit smaller, you're right it takes a lot of space on the screen, maybe it shouldn't take more than say 40% of the viewport's height

JeffAbrahamson (changes requested):
Looks almost good, it's just the token issue.

PR #826, opened 2022-06-24
Fix Deprecation warning of selenium

mcmikashi:
* Fix the DeprecationWarning:
executable_path has been deprecated, please pass in a Service object

* Fix the DeprecationWarning:
 find_element_by_* commands are deprecated. Please use find_element() instead


Shriukan33 (approved):
LGTM as well, thanks @mcmikashi  !

PR #843, opened 2022-07-05
PoC Mobilito

Shriukan33:
This PR is a discussion around Mobilito's PoC

This branch is currently (05/07/2022 ddmmyyyy) deployed to beta where you can go to /mobilito to access the landing page. 

Landing will depend on 2 things : If you are auth or not, and if you already visited the page.

If you're not auth and never visited the page, you're arriving on the tutorial page. 

image

If you already visited the page, you're still seeing the tutorial but a button on top offers to skip it (redirection to auth with a next to mobilito).

image

Clicking the buttons will redirect to login. 

Once logged in you see the actual landing, that still offers to see the tutorial or to start recording. 

image

Start recording will bring you to a very simple form that asks you the address you're monitoring, city, and zip code. I'm considering adding more details like "Adresse à laquelle vous enregistrez le traffic" or something more explicit, although this will be all explained in the tutorial, but I guess we shouldn't count on that. I'm open to your suggestions here. 

Once completed you finally land on the core app that actually registers the modal share : 

image

It is dumb easy to use, 4 big buttons with icons and number to provide feedback to user, and a button to end the recording.

mobilito_poc

Ending the recording shows a thank you page that recaps what the user did input.

We can access these results in django admin : 

image


Shriukan33:
v0.2:

## Text: 
* Bienvenue sur l'application Mobilito -> Bienvenue sur Mobilito
* Si vous avez besoin d'un rappel sur comment utiliser Mobilito --> Si vous avez besoin d'explications
* Added a header to address form so it specifies we're talking about the place they are monitoring.
## Models 
* Dropped Recording model
* Created Session and Event models : Each time a user click a button it POSTs to the server to register the event and link it with the current session. Doing so, a session can be interrupted (= No end timestamp) and the event kept nonetheless.
I added in the admin the possibility to see all related events on a session: 
image
# Color scheme
* Add Mobilitain's color scheme to buttons. They have have arbitrarily placed, I think we can come with a better one, as our colors have no specific intended meaning, they are just here to help differentiate the different buttons. 
image

Shriukan33:
History edition : 
Initially deployed to beta migration's has been restored (and immediately deleted in a 2nd migration). 
this is meant to smooth the deploy process on beta.

JeffAbrahamson:
I'm looking at this branch now with a goal of extracting something stable we can merge to master.

I note you use local storage to track whether or not people have visited.  This presents a number of perils:
* On a different device, they will have a different status.  That's likely unexpected.
* Local storage may not be particularly persistent.  Check out [this article](https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/) for example, the part entitled "7-Day Cap on All Script-Writeable Storage".

> Now ITP has aligned the remaining script-writable storage forms with the existing client-side cookie restriction, deleting all of a website’s script-writable storage after seven days of Safari use without user interaction on the site. These are the script-writable storage forms affected (excluding some legacy website data types):
>
>     Indexed DB
>     LocalStorage
>     Media keys
>     SessionStorage
>     Service Worker registrations
Impacts localStorage for sure, but impacts everything else (Service Workers, IndexedDB, etc.) so we have to live with it, there is no workaround. 

In general, when you think you can solve a problem with javascript, you should think again.  Very often, the result of solving a problem with javascript is that you then have two problems.

Here, the simple solution is simply to maintain a table of users who have used the app, since no one has used the app who hasn't authenticated.

JeffAbrahamson:
```
+++ b/transport_nantes/mobilito/static/mobilito/css/index.css
@@ -0,0 +1,4 @@
+#start-recording {
+    color: inherit;
+    text-decoration: none;
+}
```

This is a touch odd.  It's a funny name out of context (how could anyone know this is mobilito-related without searching the entire code base?).  It also doesn't do much: inherits colour (so I think that's a no-op) and specifies no text decoration.

A kinda natural question to the reader might be why this deserved a class at all.  Code comments are great for addressing this sort of thing.

JeffAbrahamson:
Postal codes in France currently have 5 digits.  In other countries, they are often longer.  (We have a rather crude postal code system here.)

Assuming no one outside your country will ever use your software is a time-tested recipe. ;-)

JeffAbrahamson:
Unless I'm horribly mistaken, when you use `EventTypes` and `CharField(choices=...)`, you get validation automatically.

See the [django docs](https://docs.djangoproject.com/en/4.0/ref/models/fields/#enumeration-types).  This simplifies things quite a bit.

So does noting that we don't support translation (anywhere), so it's not worth going out of our way to do it here.  That means we can allow automatic user-readable name generation.

It's also worth noting that the events table is likely to grow quite a lot, so using verbose names _in the table_ doesn't provide much value besides tripling or quadrupling table size, with some knock-on effect on performance.  Ideally postgresql would use an enum, but I'm not sure that's spec'd to be guaranteed.

Shriukan33:
> I'm looking at this branch now with a goal of extracting something stable we can merge to master.
> 
> I note you use local storage to track whether or not people have visited. This presents a number of perils:
> 
>     * On a different device, they will have a different status.  That's likely unexpected.
> 
>     * Local storage may not be particularly persistent.  Check out [this article](https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/) for example, the part entitled "7-Day Cap on All Script-Writeable Storage".
> 
> 
> > Now ITP has aligned the remaining script-writable storage forms with the existing client-side cookie restriction, deleting all of a website’s script-writable storage after seven days of Safari use without user interaction on the site. These are the script-writable storage forms affected (excluding some legacy website data types):
> > ```
> > Indexed DB
> > LocalStorage
> > Media keys
> > SessionStorage
> > Service Worker registrations
> > ```
> 
> Impacts localStorage for sure, but impacts everything else (Service Workers, IndexedDB, etc.) so we have to live with it, there is no workaround.
> 
> In general, when you think you can solve a problem with javascript, you should think again. Very often, the result of solving a problem with javascript is that you then have two problems.
> 
> Here, the simple solution is simply to maintain a table of users who have used the app, since no one has used the app who hasn't authenticated.

Hmm indeed, didn't notice that safari would delete it after 7 days of inactivity.
Will need to think of something different. Maybe just drop the skip tutorial. 

The tutorial is accessible for anon users, so the user table idea isn't filling its role here, unless we're forcing login to even get there.

Shriukan33:
> ```
> +++ b/transport_nantes/mobilito/static/mobilito/css/index.css
> @@ -0,0 +1,4 @@
> +#start-recording {
> +    color: inherit;
> +    text-decoration: none;
> +}
> ```
> 
> This is a touch odd. It's a funny name out of context (how could anyone know this is mobilito-related without searching the entire code base?). 

You can tell because it's located in /transport_nantes/mobilito/static/mobilito/css/index.css :) I made a css file per page for stuff specific to mobilito, so mobilito's classes aren't applying site-wide like mobilitains.css does.

Also there are several things on this piece of code : 

- It's not a class, it's an ID (you can tell with leading  #), IDs being unique on a page, it only applies to a single element. It has higher specificty than a class (https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#selector_weight_categories) it allows me to override class properties that would apply for other file s... Like mobilitains.css. This leads me to another point.
- Color: inherit isn't a no-op. Indeed in mobilitains.css we have lots of classes that changes the `<a>` tag's color, that's a behavior I didn't want here. inherit, with the specificity of an ID, will override these, and catch the color of its next closest parent. 
- Not text decoration is also not a noop because .. it is enable by default, it adds the small line under links in `<a>` tags.

Sure a comment could have cleared this up, but you usually don't really care about CSS at all.

Shriukan33:
> Postal codes in France currently have 5 digits. In other countries, they are often longer. (We have a rather crude postal code system here.)
> 
> Assuming no one outside your country will ever use your software is a time-tested recipe. ;-)

So should be using a longer string ? Like 100 ?

Shriukan33:
> Unless I'm horribly mistaken, when you use `EventTypes` and `CharField(choices=...)`, you get validation automatically.
> 
> See the [django docs](https://docs.djangoproject.com/en/4.0/ref/models/fields/#enumeration-types). This simplifies things quite a bit.
> 
> So does noting that we don't support translation (anywhere), so it's not worth going out of our way to do it here. That means we can allow automatic user-readable name generation.
> 
> It's also worth noting that the events table is likely to grow quite a lot, so using verbose names _in the table_ doesn't provide much value besides tripling or quadrupling table size, with some knock-on effect on performance. Ideally postgresql would use an enum, but I'm not sure that's spec'd to be guaranteed.

I hate being the bearer of bad news, but sadly no, the validity isn't enforced, I know because I tried it. I changed the html in my browser and changed "pedestrian" name in "albert", and proceeded to submit. Sure enough, it created an event type Albert in database ! 
The label is displayed in the admin panel. Sure the generated one would be enough in this case, but I don't know the performance/size delta here between the two nor how you measure it :/

Shriukan33:
Closing this PR because this branch isn't the main one anymore, created another one for convenience.
We can keep interacting about your review here, it's just to make clear it isn't this branch that is meant to be worked on.

PR #849, opened 2022-07-15
Improve press app allow user to refetch og data

### LINKED ISSUE #662: Permit OG refetch
One #659 and #661 are done, let's add a third button to the right of the pressmention admin lines: an og refetch button.  Perhaps `fa-refresh` or `fa-repeat`.

Whilst doing this, make sure `fa-trash` is right-most and separated by a bit of extra space so it's hard to hit by accident.

mcmikashi:
Closes #662


JeffAbrahamson:
Hey, cool, good idea @mcmikashi !

I tried entering a new article (because my previous tests locally had been with example.com links).  I got this:

image

The link I'd used was this:

https://www.ouest-france.fr/faits-divers/incendie/temoignages-il-y-a-deux-ans-la-cathedrale-de-nantes-etait-en-feu-ca-m-a-bouleverse-0c708cae-02b1-11ed-8fbe-98a3866fe2c5

It's got an adblock detector, so it thinks you're a bot (well, you are...) and so what it renders doesn't have OG info.  That shouldn't stop us from being able to enter the data, though, but now it does.

Wdyt?

JeffAbrahamson:
The commit message is a bit light. ;-)

mcmikashi:
Allow the user to save the data and set default data if og data is missing like the title and the description for the og title and og description and add a default image ?

mcmikashi:
* Add a new function for fetching open graph data of press_mention
* Allow user to send article that doesn't have open graph data
* Allow user to update open graph data on the table list view
* When user update the a press_mention the open graph data of this object is updated too
* Reword the commit message

transport_nantes/press/views.py
```diff
@@ -165,3 +141,66 @@ class PressMentionDetailView(PermissionRequiredMixin, DetailView):
...
+        press_mention = PressMention.objects.get(
+            pk=self.request.GET.get("refresh"))
+        url = press_mention.article_link
+    try:
+        page = requests.get(url)
+    except Exception as e:
+        logger.error(f"Can't acess to the website {e}.")
+        return messages.add_message(
+            self.request, messages.WARNING,
+            'Le liens est incorrect ou le site web est indisponible.')
```
Shriukan33, line 167: liens -> lien

transport_nantes/press/views.py
```diff
@@ -165,3 +141,66 @@ class PressMentionDetailView(PermissionRequiredMixin, DetailView):
...
+    tree = html.fromstring(page.content.decode("utf-8"))
+    og_title = tree.xpath('//meta[@property="og:title"]/@content')
+    og_description = tree.xpath(
+        '//meta[@property="og:description"]/@content')
+    og_image = tree.xpath('//meta[@property="og:image"]/@content')
+    if not og_title or not og_description:
+        logger.error("This website doesn't have all open graph data.")
+        return messages.add_message(
+            self.request, messages.WARNING,
+            'Le site ne contient pas de meta données Open Graph.')
```
Shriukan33, line 177: I think it's a bit misleading, let's assume some article does have a title and an image, we would still claim that there is none available.
Is it mandatory to have a description ? Maybe some only define a title ?

transport_nantes/press/views.py
```diff
@@ -165,3 +141,66 @@ class PressMentionDetailView(PermissionRequiredMixin, DetailView):
...
+        else:
+            press_mention.og_title = og_title[0]
+            press_mention.og_description = og_description[0]
+    resp = requests.get(og_image[0])
+    if resp.status_code != requests.codes.ok:
+        logger.error("The link of the open graph "
+                     "image may be dead, or doesn't exist.")
+        return messages.add_message(
+            self.request, messages.WARNING,
+            'Le site ne contient pas de meta données Open Graph.')
```
Shriukan33, line 191: Same here, you should probably just state that the image URL is broken, rather than saying there is no data.

transport_nantes/press/views.py
```diff
@@ -165,3 +141,66 @@ class PressMentionDetailView(PermissionRequiredMixin, DetailView):
...
+    else:
+        fp = BytesIO()
+        fp.write(resp.content)
+        file_name = og_image[0].split('/')[-1]
+        if form:
+            form.instance.og_image.save(file_name, files.File(fp))
+            return messages.add_message(
+                self.request, messages.SUCCESS,
+                "L'opération a été  réalisée avec succès.")
```
Shriukan33, line 200: This might be a too generic message, What operation ? What has it done ? "Les données open graph on été enregistrées avec succès" would be clearer, I think.

transport_nantes/press/views.py
```diff
@@ -165,3 +141,66 @@ class PressMentionDetailView(PermissionRequiredMixin, DetailView):
...
+            self.request, messages.WARNING,
+            'Le site ne contient pas de meta données Open Graph.')
+    else:
+        if form:
+            form.instance.og_title = og_title[0]
+            form.instance.og_description = og_description[0]
+        else:
+            press_mention.og_title = og_title[0]
+            press_mention.og_description = og_description[0]
+    resp = requests.get(og_image[0])
```
Shriukan33, line 185: This line raises an IndexError if no og_image is available. You assume that there is an og:image meta tag here.

Shriukan33 (changes requested):
Good job on this, got it to work for a lemonde.fr article :) 

However there are few caveats : 

image

The filter doesn't seem to apply with both start and end date. Just start date works. End date alone doesn't.

transport_nantes/press/views.py
JeffAbrahamson: It's not an error that someone chooses to configure their site without opengraph data or twitter cards.  We just have to roll with that and make the best of things.  Indeed, this is why we ask for a short user summary: so we have something to display if we don't fetch og data.  (I increasingly think our asking for a description is redundant, that we should just offer to provide one in cases where the og data doesn't represent well.)

transport_nantes/press/views.py
JeffAbrahamson: We really don't need or want a user message about opengraph data.  Most users will not quite know what to make of that.  Logging it is great, and in that case more tersely and factually and in English.

transport_nantes/press/views.py
JeffAbrahamson: And, again, the user response to that message should be a mitterrandien "et alors ?".

JeffAbrahamson (changes requested):

mcmikashi:
- remove django message to the user
- a new try catch to the fetching_open_graph_data
- change the log type
- add comment
For duplicate name by default django add an underscore and a random 7  character alphanumeric string (before the file extension, if one exists) to the filename.[link](https://github.com/django/django/blob/main/django/core/files/storage.py)

transport_nantes/press/views.py
```diff
@@ -54,6 +57,10 @@ def get_context_data(self, **kwargs):
                 self.request.GET.get("article_date_start")
             context["search_form"].fields["article_date_end"].initial = \
                 self.request.GET.get("article_date_end")
+        if self.request.GET.get("press_mention_refresh"):
+                fetching_open_graph_data(
```
Shriukan33, line 61: Over indented :)

transport_nantes/press/views.py
```diff
@@ -165,3 +144,74 @@ class PressMentionDetailView(PermissionRequiredMixin, DetailView):
     model = PressMention
     template_name = "press/press_detail.html"
     permission_required = 'press.press-editor'
+
+
+def fetching_open_graph_data(form=False, presmention_id=None):
+    """
+    This function allow fetch open graph data of an url from
+    a form or from the get data of the url and return
+    a django message.
```
Shriukan33, line 153: No message is returned anymore

transport_nantes/press/views.py
```diff
@@ -165,3 +144,74 @@ class PressMentionDetailView(PermissionRequiredMixin, DetailView):
...
+        if form:
+            form.instance.og_title = og_title[0]
+            form.instance.og_description = og_description[0]
+        else:
+            press_mention.og_title = og_title[0]
+            press_mention.og_description = og_description[0]
+    try:
+        resp = requests.get(og_image[0])
+    except IndexError:
+        logger.warning(f"This website doesn't have open graph image.")
```
Shriukan33, line 187: f-string is missing placeholders flake8(F541)

transport_nantes/press/templates/press/press_detail.html
```diff
@@ -1,5 +1,19 @@
 {% extends 'asso_tn/base_mobilitain.html' %}
 {% block content %}
+{% if messages %}
+<div class="row">
+{% for message in messages %}
```
Shriukan33, line 5: There are no messages anymore, I believe ?

transport_nantes/press/models.py
```diff
@@ -1,3 +1,4 @@
+from email.policy import default
```
Shriukan33, line 1: This import has probably been made automaticaly by your IDE

Shriukan33 (commented):
Other than details, LGTM !

transport_nantes/press/models.py
mcmikashi: Fixed

transport_nantes/press/views.py
mcmikashi: Fixed

transport_nantes/press/views.py
mcmikashi: Fixed

transport_nantes/press/views.py
```diff
@@ -11,6 +11,9 @@
 import requests
```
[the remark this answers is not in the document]
mcmikashi: Fixed

JeffAbrahamson:
Thanks a million @mcmikashi and sorry for the slow code reviews.

I've merged this.

PR #850, opened 2022-07-17
Add clean up code for TBObjects

JeffAbrahamson:
Hey, @Shriukan33 , this is far from perfect but wdyt?

My test env is broken, so whilst I fix it, let's see what gh test says.


transport_nantes/topicblog/views.py
```diff
@@ -1375,3 +1383,34 @@ def click_received_handler(sender, mail_obj, click_obj, *args, **kwargs):
...
+    slug_queryset.filter(publication_date=None)
+    for tb_object in slug_queryset:
+        if tb_object.is_deletable():
+            logger.info(f"Deleting object {tb_object.id}.")
+            tb_object.delete()
+        elif tb_object.is_moribund() and \
+             tb_object.scheduled_for_deletion_date is None:
+            logger.info(f"Marking as moribund object {tb_object.id}.")
+            tb_object.scheduled_for_deletion_date = datetime.now(timezone.utc)
+            tb_object.save()
```
Shriukan33, line 1416: When saving, tb_object.date_modified will be set to today because of auto_now (https://docs.djangoproject.com/fr/4.0/ref/models/fields/#django.db.models.DateField.auto_now ). Doing so this line in models, `is_moribund` : 

```python
        if self.first_publication_date is None and \
           ((datetime.now(timezone.utc) - self.date_modified).days >=
                self.K_MORIBUND_DELAY_DAYS):
```
Will evaluate to today - today >= 15 -> 0 >=15 and return False, while it should be saying yes, it's moribund.
The consequence of this is that you don't get to see the countdown in the listone.

This doesn't show up in your tests because you manually set the date_modified attribute.

Shriukan33 (changes requested):

Shriukan33:
Also I noticed that the is_moribund function is called twice when loading a listviewone

transport_nantes/topicblog/models.py
```diff
@@ -314,6 +323,62 @@ def get_participating_field_names(self) -> set:
...
+            # old it is, so it's moribund.
+            return True
+        if self.first_publication_date is None and \
+           ((datetime.now(timezone.utc) - self.date_modified).days >=
+            self.K_MORIBUND_DELAY_DAYS):
+            return True
+        return False
+
+    def is_deletable(self) -> bool:
+        """Return True iff the object may be deleted.
```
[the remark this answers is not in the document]
JeffAbrahamson: if and only if
https://en.wikipedia.org/wiki/If_and_only_if

transport_nantes/topicblog/views.py
JeffAbrahamson: Wait, does it reset on every save??  😢

JeffAbrahamson: Right.  Done.

JeffAbrahamson:
I think this is ready for you again @Shriukan33

Issue #840, opened 2022-06-30
Verity that all tnmd operations render in email


JeffAbrahamson:
Because the CTA didn't.  ;-)

Shriukan33:
Should CTA buttons render MD ?  Like the label ?

JeffAbrahamson:
The markdown engine identifies requests for html and, for TN-specific items, calls functions that render it.  Those functions consult the context to determine if they are in email or on the web, then emit the appropriate html for that context.

The CTA does not render nicely in email.  It should.

PR #854, opened 2022-07-21
Make inline tn_markdown context aware

Shriukan33:
This allows us to know if we're rendering for web or email.
Email don't accept css classes, and require hardcoded styles.

Part of #840


JeffAbrahamson:
LGTM.

PR #856, opened 2022-07-22
Add MailingListPitchView to url paths

Shriukan33:
We kinda forgot to add the MailingListPitchView to the url paths.

Deployed to beta 22/07


JeffAbrahamson (approved):
LGTM

PR #857, opened 2022-07-25
Benja add missing elements for publication list

Shriukan33:
This PR adds a message stating what's missing to publish the item in the admin banner.

I took this opportunity to  also :
* Add some padding to buttons so they're not as glued together
* Refactor the templates

We went with 5 different templates (`base_(topicblog | mlp | email | press | launcher).html`) that didn't bring anything, the only differences were :
* In urls, so I added a url property to models when relevant much like the Sendable Views were refactored.
* Press and Email had a button to send to self, however this can be easily controlled by setting the "sendable" context variable to a truthy value. So I merged that bit in the base template. 

Doing so we reduce the number of templates from 5 to 1.

While cleaning refs to no-longer-used templates, I noticed that the "context_appropriate_base_template" referred to these no longer used templates, so I cleaned that bit as well, defaulting to base_topicblog. 

This made me realize that when we pick a template in create form in Email and CP, it does only matter for the email version of those objects. 

## How does the view pick which template to use to display a given object ?

The base classes, `TopicBlogBaseView` and `TopicBlogBaseViewOne` both set the `self.template_name` attribute in the `get_context_data` method. 

https://github.com/transport-nantes/tn_web/blob/6236a89480ea19ea66311ec9d67226c7b4cfbad6/transport_nantes/topicblog/views.py#L217-L225

This behavior *is not* overrode in the following child classes : 
* `TopicBlogItemView` and `TopicBlogItemViewOne`
* `TopicBlogLauncherView` and `TopicBlogLauncherViewOne`
* `TopicBlogMailingListPitchView` and `TopicBlogMailingListPitchViewOne`

On the other hand, it is overridden in the following child classes : 
* `TopicBlogEmailView` and `TopicBlogEmailView`
* `TopicBlogPressView` and `TopicBlogPressView`

In these last classes, we instead hard code a template that is either
 
https://github.com/transport-nantes/tn_web/blob/6236a89480ea19ea66311ec9d67226c7b4cfbad6/transport_nantes/topicblog/views.py#L1021-L1027

or 

https://github.com/transport-nantes/tn_web/blob/6236a89480ea19ea66311ec9d67226c7b4cfbad6/transport_nantes/topicblog/views.py#L907-L912

(Same for ...ViewOne variations)

## What can I learn from this ?

In web version, no matter what template you pick in creation form, it's always one of those two that will be used in the web View and ViewOne version. 

## And how does it pick the correct template for email clients then ? 

This snippet shows that the template is fetched at the object's level, so the template we actually picked in the creation form 

https://github.com/transport-nantes/tn_web/blob/6236a89480ea19ea66311ec9d67226c7b4cfbad6/transport_nantes/topicblog/views.py#L355-L378

This is the same code being used to send to oneself or to to send to a mailing list. 

# What's to take from all this ?

I think this is a weird move to have an hardcoded template only for press and emails.
To be consistent, the template picked at creation should be the one use to display on the website, just like the other objects.

Maybe we should have a matching email template for each template we create. 
Doing so, picking a template in creation form sets both web and email templates.

We would go from 

https://github.com/transport-nantes/tn_web/blob/edd18f378943514c456059725287e98e830a3656/transport_nantes/topicblog/models.py#L834-L852

To something like : 

```python
    template_config = {
        'topicblog/content_email.html': {
            'email_template': 'topicblog/content_email_client.html',
            'default_choice': True,
            'user_template_name': 'Classique',
            'active': True,
            "fields": {
                'slug': True,
                'title': True,
                'subject': True,
                'body_text_1_md': True,
            },
            "optional_fields_for_publication":
                template_config_default['optional_fields_for_publication'],
            "one_of_fields_for_publication":
                template_config_default['one_of_fields_for_publication'],
            "dependent_field_names":
                template_config_default['dependent_field_names'],
        },
    }
```

It still has the flaw to pair one template to another, but I think it does make sense to have pairs (because it is supposed to adapt one to another)

WDYT ?


JeffAbrahamson:
We shouldn't have two templates for each.  You should simply inherit from the email base or the web base template.
* `asso_tn/templates/asso_tn/base_mobilitain.html`
* `asso_tn/templates/asso_tn/base_email_mobilitain.html`

That used to happen via a `... | default` in the `{% extends %}`.

JeffAbrahamson:
The missing elements should be available to anyone who is allowed to edit or view the unpublished object, not just to staff.  Perhaps the block is simply misnamed?

transport_nantes/topicblog/templates/topicblog/base_topicblog.html
```diff
@@ -10,42 +10,98 @@
...
 
-	    {% if page.publication_date %}
-	    <a type="button" class="btn btn-outline-info btn-sm"
-	       href="{% url 'topic_blog:view_item_by_slug' page.slug|default:'--sans-slug--' %}">Visualiser (usager)</a>
-	    {% endif %}
-	</p>
-    </div>
-</div>
-{% endif %}{{ block.super }}
+			{% if transactional_send_record_class %}
```
JeffAbrahamson, line 55: I think this should be done with a second template that inherits from this one and simply modifies this bit, possibly by calling `block.super` and then adding something.  Imagine that we have another thing that's sort of like sendable but different, and then we add that block here.  And then another.  This template shouldn't know about all the ways it might be used.

JeffAbrahamson: but it's a good idea to generalise it!

transport_nantes/topicblog/templates/topicblog/base_topicblog.html
```diff
@@ -10,42 +10,98 @@
...
+		{% if not page.is_publishable %}
+			<span class="row"> Éléments manquants pour la publication :
+			{% for missing_item in page.get_missing_publication_field_names %}
+					{{ missing_item }}, 
+			{% endfor %}
+			</span>
+		{% endif %}
+	</div>
+	{% endif %}
+	{% if user.is_staff %}
```
JeffAbrahamson, line 87: Why is this only for staff?

transport_nantes/topicblog/templates/topicblog/base_topicblog.html
```diff
@@ -10,42 +10,98 @@
...
+		{% if sendable %}
```
JeffAbrahamson, line 88: Same comment about non-sendables shouldn't know about sendable.

transport_nantes/topicblog/templates/topicblog/content_press.html
```diff
@@ -1,4 +1,4 @@
-{% extends context_appropriate_base_template %}
+{% extends 'topicblog/base_topicblog.html' %}
```
JeffAbrahamson, line 1: ah, yes, here is where the email vs web switch happens!

Shriukan33:
I edited this PR so base_email and base_press extend the topicblog_base template and add the Send to self and send buttons. 
I also removed the is_staff condition, as the view already checks the permissions to toggle the context var "sendable"

I didn't create pairs of template just yet.

JeffAbrahamson:
Perhaps the way to sort the two template problem is simply to require the config to specify two templates (one for mail, one for web), since clearly we can't have One Template to Rule Them All (tm).

Wdyt?

JeffAbrahamson (approved):
LGTM after our conversation yesterday updating my view on the world.

Shriukan33:
> Perhaps the way to sort the two template problem is simply to require the config to specify two templates (one for mail, one for web), since clearly we can't have One Template to Rule Them All (tm).
> 
> Wdyt?

Yup we can do that, so TBE and TBP would have an extra Charfield that would be self.email_template, drawn from an enum of all our email templates ? My only concern would be that it would be confusing to have an email version that doesn't match the web version. (hence the pair idea)

JeffAbrahamson:
> > Perhaps the way to sort the two template problem is simply to require the config to specify two templates (one for mail, one for web), since clearly we can't have One Template to Rule Them All (tm).
> > Wdyt?
> 
> Yup we can do that, so TBE and TBP would have an extra Charfield that would be self.email_template, drawn from an enum of all our email templates ? My only concern would be that it would be confusing to have an email version that doesn't match the web version. (hence the pair idea)

I'm pretty open.  Maybe I misunderstood your pair idea.  I was just thinking that if we're going to have two templates (a pair of templates), we should specify them in the config rather than playing magical games with paths.

Shriukan33:
> > > Perhaps the way to sort the two template problem is simply to require the config to specify two templates (one for mail, one for web), since clearly we can't have One Template to Rule Them All (tm).
> > > Wdyt?
> > 
> > 
> > Yup we can do that, so TBE and TBP would have an extra Charfield that would be self.email_template, drawn from an enum of all our email templates ? My only concern would be that it would be confusing to have an email version that doesn't match the web version. (hence the pair idea)
> 
> I'm pretty open. Maybe I misunderstood your pair idea. I was just thinking that if we're going to have two templates (a pair of templates), we should specify them in the config rather than playing magical games with paths.

yes my idea was to say in config 'content_email.html' has 'content_email_mail_client.html' as its mail client counterpart, and we'd create another template say 'awesome_tbe.html' we'd have to precise in template_config that is has "awesome_tbe_mail_client.html' as its mail client counterpart etc

The send views would fetch the `object.template_name` string (which is used to display on web), look in `self.template_config[object.template_name] ["email_template"]` and tada ?

PR #858, opened 2022-07-26
Add header title and description to email templates

Shriukan33:
I fiddled with the absolute positioning to get a website-like result, turns out it's incompatible with Outlook and Gmail (both web and mobile), the only solution I found has been to put it below the image.


JeffAbrahamson (approved):
LGTM.

PR #860, opened 2022-07-26
Make optional mailing list for Mailing List Pitches

Shriukan33:
That way, one can save a draft in case the
mailing list isn't existing yet.


transport_nantes/topicblog/templates/topicblog/content_mlp.html
```diff
@@ -79,8 +79,11 @@ <h1 class="hero-title">{{ page.header_title }}</h1>
                 </a>
             </div>
           {% endif %}
-          {% show_mailing_list  mailinglist=page.mailing_list.mailing_list_token title=page.subscription_form_title %}
-
+          {% if page.mailing_list %}
+               {% show_mailing_list  mailinglist=page.mailing_list.mailing_list_token title=page.subscription_form_title %}
+          {% else %}
+               <p>Aucune liste de diffusion n'est disponible pour l'instant.</p>
```
JeffAbrahamson, line 85: I find this visually a bit odd, since the user has to _remember_ that what is missing is a button.  Also, the "for the moment" doesn't quite make sense: what is the source of the temporality?  You're assuming the user remembers what all these things are supposed to do.  But it's your job to guide the user so that s/he remembers as it happens.

Instead, perhaps make an inactive button with "no list" as the button text.

JeffAbrahamson (changes requested):
Careful of non-empty blank lines (`^ +$`).

Shriukan33:
Removed non empty blank lines

Trying to click the button while no token was set did 500, so I saved the kittens by displaying an error instead.

image

JeffAbrahamson:
> Removed non empty blank lines
> 
> Trying to click the button while no token was set did 500, so I saved the kittens by displaying an error instead.
> 
> image

This is good for devs to see, but an actual user would be pretty annoyed to be asked to click a button that only leads to an unintelligible error.

Can we just disable the button if it's not usable?  We can know this before the user clicks, no?  And in that case, perhaps simply display a message like "Something went wrong.  It's us, not you.  If this persists, we'd be happy if you tell us."

transport_nantes/mailing_list/forms.py
```diff
@@ -68,19 +68,35 @@ class SubscribeUpdateForm(Form):
 # First step of quick signup to the newsletter
 
 
+def validate_mailinglist_name(mailinglist_token):
+    """Validate the mailing list name.
+
+    """
+    if not MailingList.objects.filter(
+            mailing_list_token=mailinglist_token).exists():
+        raise forms.ValidationError(
```
JeffAbrahamson, line 77: Does the validation handler catch this for you and turn it into "invalid"?

JeffAbrahamson (commented):
LGTM, modulo question about catching validation throw.

In the commit comment, "inexisting" -> non-existent".  And "token" is a pretty generic word, use the technical term from the code, `mailing_list_token` (with spaces or underscores as seems context appropriate).

Shriukan33:
> > Removed non empty blank lines
> > Trying to click the button while no token was set did 500, so I saved the kittens by displaying an error instead.
> > image
> 
> This is good for devs to see, but an actual user would be pretty annoyed to be asked to click a button that only leads to an unintelligible error.
> 
> Can we just disable the button if it's not usable? We can know this before the user clicks, no? And in that case, perhaps simply display a message like "Something went wrong. It's us, not you. If this persists, we'd be happy if you tell us." Most people who land on the page won't click the button anyway, so we don't want to distract, just explain if you do get to the button.

Hopefully we don't publish a page that doesn't have a mailing list :sweat_smile:
I originally only left a message and no button, you asked that there is a button because it's less confusing but if it doesn't work I think it's even more confusing isn't it ?
I totally understand the presence of the button for layout pre-visualisation, maybe I could update the message ?

JeffAbrahamson:
> > > Removed non empty blank lines
> > > Trying to click the button while no token was set did 500, so I saved the kittens by displaying an error instead.
> > > image
> > 
> > 
> > This is good for devs to see, but an actual user would be pretty annoyed to be asked to click a button that only leads to an unintelligible error.
> > Can we just disable the button if it's not usable? We can know this before the user clicks, no? And in that case, perhaps simply display a message like "Something went wrong. It's us, not you. If this persists, we'd be happy if you tell us." Most people who land on the page won't click the button anyway, so we don't want to distract, just explain if you do get to the button.
> 
> Hopefully we don't publish a page that doesn't have a mailing list sweat_smile I originally only left a message and no button, you asked that there is a button because it's less confusing but if it doesn't work I think it's even more confusing isn't it ? I totally understand the presence of the button for layout pre-visualisation, maybe I could update the message ?

True, though most imaginable mistakes eventually happen (and plenty that one doesn't image). ;-)

Let's not hold up the PR on account of this, but in general it's best not to say "click this" followed by "ha ha".

transport_nantes/mailing_list/forms.py
Shriukan33: If we raise form.ValidationError, the form is considered Invalid and form_invalid() is called, if this is your question ?

Shriukan33:
I reworded the commit according to your suggestion, didn't change the message though. Any preference ? Or is it unimportant enough so we can merge this ?

transport_nantes/mailing_list/forms.py
JeffAbrahamson: Yup, just didn't remember precisely and wanted to be sure it was caught!

JeffAbrahamson:
> I reworded the commit according to your suggestion, didn't change the message though. Any preference ? Or is it unimportant enough so we can merge this ?

LGTM.

PR #861, opened 2022-07-28
Add validation on Header image

Shriukan33:
If one fills any of the title or description field for the header,
then they must also fill the image.


JeffAbrahamson (approved):
LGTM.

Issue #40, opened 2020-05-23
Redesign observatoire de la grande mobilité as per Simon suggestions


JeffAbrahamson:
So here's how I see it. Ok, the drawing attached is crap due to my cellphone camera but you'll get the idea.

3 screens. Number 1 on the left:

* Logo, "observatoire", hamburger menu, all on the same stripe.
* then, underneath, a map.
* and a button that says 'localisez moi" and another one that says "commencer l'enregistrement" or something like that.

When you click on that last button, you'll get to screen 2:

* same menu on top.
* a button that says "arreter l'enregistrement" with a "stop" or "pause" icon
* 4 buttons at the bottom of the screen using the whole bottom half, with 4 icons that you can find on noun project or somewhere else (a font maybe) showing a bike, a pedestrian, a car and a train. Of 4 different colors. Such as a windows logo

When you press "arreter l'enregistrement" you'll get to the 3rd screen

*    same menu
*    the same is a recap with a map of where you are
*    the a line that says "le 19/05/20 de: 13h05 à: 13h24"
*    then the numbers for each transport mode
*    then a buttons that says send

A forth screen, that I haven't drawn, which says thank you and allows you to share the thing on FB, Twitter, etc.

There is also a screen zero, where you explain shortly what it is and what's going to happen.

Why all this? The user needs a lot of feedback. A user needs to know what's going on - is it started or not, since when, etc - and needs a reward at the end: a result screen (so that he/she can see the work that's done) and and a way to share it.

I don't think you need to be able to correct the numbers if you press the buttons wrong. This is going to be very minor occurance, I think. Maybe you can make the screen 3 editable though.

About screen 2: the user needs to be able to press the buttons without looking, so they have to occupy the whole bottom screen and there needs to be nothing else there. If you can add a little sound with each press that's even better. also if you can show the current duration of the recording on top of the "stop" button that's top.

I would have done that for you in no time had you keep the Divi/WP combo. But it's bootstrap so you'll have to do it yourself.

Anyway think of your users as mentally challenged people who understand very little of the job they have to accomplish. We're all like this when we're not really in charge, not paying attention, and not paid in the end. Make everything super simple and clear. As little words as possible.

Also imho not sure you can ignore someone in a small street staying there for 15 minutes and seing one car. If you don't record any start/end time, you'll lose that session and I'd say it's relevant - as in "no bike seen in this street despite being the shortest route between 2 major points". 

[...]

PR #865, opened 2022-08-03
Mobilito tutorial

Shriukan33:
This PR adds an embryo of tutorial.

It directs users until they've seen all of the pages the tutorial has to show, checking MobilitoUser model first then Session cookie.

Part of #40


JeffAbrahamson:
This is really strange: a dropdown that displays text that is not in the dropdown.  Where have you seen such a thing?

It should show what it is displaying now.

image

JeffAbrahamson:
The first time through I'm not asked where I am.  The second time I am.

Shriukan33:
Hmm actually I didn't realize that at first, oops ! 
Also, this doesn't implement the redirect to tutorial before recording yet.

Shriukan33:
> This is really strange: a dropdown that displays text that is not in the dropdown. Where have you seen such a thing?
> 

It's bootstrap's dropdown pills https://getbootstrap.com/docs/5.0/components/navs-tabs/#pills-with-dropdowns

JeffAbrahamson:
In dev, sending mail tries to use SES if that's configured.  This is nice, but I wonder if this is the right thing to do.

First, regardless of what I've configured, we should log mails sent to console in dev mode, because we need to see what happens.

Second, if we actually sent mail to someone other than us, it would look really weird with links to localhost.  So if we are gong to do that, we really had better make sure the mail flags that it's being sent from a development instance.

JeffAbrahamson:
In a cookieless session, I started (from `/mobilito/`) by clicking "record" and I was directed through auth, then dropped into the recorder.  This should have instead sent me through the tutorial.

Shriukan33:
> In a cookieless session, I started (from `/mobilito/`) by clicking "record" and I was directed through auth, then dropped into the recorder. This should have instead sent me through the tutorial.

-> https://github.com/transport-nantes/tn_web/pull/865#issuecomment-1203939150

transport_nantes/mobilito/views.py
```diff
@@ -38,6 +38,132 @@ class TutorialView(TemplateView):
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        tutorial_page = kwargs.get('tutorial_page', 'presentation')
+        if tutorial_page not in self.all_tutorial_pages:
+            # Trying to access a tutorial page that doesn't exist.
+            raise Http404()
+        context['active_tab'] = tutorial_page
+        self.template_name = f'mobilito/tutorial_{tutorial_page}.html'
+        # We're to use this pattern to make str <-> datetime conversions
+        self.strftime_pattern = "%Y-%m-%d %H:%M:%S"
```
JeffAbrahamson, line 74: When you want an RFC 3399 or an ISO 8601 date, you should use the specific functions that give you that rather than specifying formats.

At least use "%F %T" if no other function is available.

Shriukan33:
> After playing with this a bit, I got to `/mobilito/` and click "record". I notice that I do not have a mobilito cookie and I do not have a mobilito-user record. Why does it decide I should be able to record?

You only need auth to access record, as I said, it does NOT redirect you to tutorial for now

transport_nantes/mobilito/views.py
```diff
@@ -38,6 +38,132 @@ class TutorialView(TemplateView):
...
+                    self.request.session[tutorial]["last_seen"],
+                    self.strftime_pattern).replace(tzinfo=timezone.utc)
+                # The page has been updated since the last time we saw it.
+                if self.all_tutorial_pages[tutorial]["last_modified"] > last_seen:
+                    # The tutorial page exists, and has been updated since the
+                    # last time the user saw it, according to session cookie
+                    to_visit_pages.append(tutorial)
+                # The page has not been updated since the last time we saw it.
+                # But it's been more than 60 days since they saw it.
+                elif (self.now - last_seen).days > 60:
```
JeffAbrahamson, line 139: Please drop this entire block.

1. We have no data to support its necessity.
2. I rather suspect marketing emails will be a better way to do it anyway.
3. In so much as we haven't yet thought through the user experience, and that will require extra work, let's not put it on the critical path to getting this code delivered, since if it's here, we have to test it and provide reasonable user experience for it.
4. Why should "60" be hard-coded as a magic number?

JeffAbrahamson:
The commit comment on 7131f14 is quite terse and rather underrepresents what is in the nearly 400 line commit.

transport_nantes/mobilito/models.py
```diff
@@ -14,6 +14,7 @@ class MobilitoUser(models.Model):
     user = models.ForeignKey(User, on_delete=models.PROTECT, related_name='+')
     # Set first_time to False once the user has read the tutorial.
     first_time = models.BooleanField(default=True)
+    completed_tutorial_timestamp = models.DateTimeField(null=True, blank=True)
```
JeffAbrahamson, line 17: This needs a comment.

Shriukan33:
> The commit comment on [7131f14](https://github.com/transport-nantes/tn_web/commit/7131f145f1bc4d2568665db6c0023da012eaeaa7) is quite terse and rather underrepresents what is in the nearly 400 line commit.

Agreed, I had the same thought but really it's all it really does, displays a tutorial to users.
The how is in the code, so I thought I shouldn't talk about the how in the commit message, there is enough comments in the code.

JeffAbrahamson:
The commit comment on 25a89df doesn't make sense to me.  Of course if the user deletes their cookie we don't know them anymore.  That's the point, really.

JeffAbrahamson (changes requested):

JeffAbrahamson:
> > The commit comment on [7131f14](https://github.com/transport-nantes/tn_web/commit/7131f145f1bc4d2568665db6c0023da012eaeaa7) is quite terse and rather underrepresents what is in the nearly 400 line commit.
> 
> Agreed, I had the same thought but really it's all it really does, displays a tutorial to users. The how is in the code, so I thought I shouldn't talk about the how in the commit message, there is enough comments in the code.

You do more than "Add small description and images in mobilito's tutorial".  You also manage the logic of showing the user all the pages.

transport_nantes/mobilito/views.py
Shriukan33: This pattern works for its purpose, I needed one that can make a string go back to datetime object and vice versa, it's not down to the ms sure, but it doesn't need to be, afaik.

What the change to make here ?

transport_nantes/mobilito/views.py
Shriukan33: Dropped in next push

transport_nantes/mobilito/models.py
Shriukan33: Added

Shriukan33:
Reworded commit

JeffAbrahamson:
Your tutorial view is unnecessarily complicated.  Here's a simpler version (with some pseudocode where I didn't remember the right grammar).  Also adds some DRY pixie dust.

```
class TutorialView(TemplateView):
    """Present the tutorial.
    """

    template_name = 'mobilito/tutorial.html'
    tutorial_cookie_name = 'mobilito_tutorial'

    # Map each tutorial page name to the datetime before which the
    # user is assumed not to have see it (i.e., the page's last
    # modification or its creation or just we want to show it again).
    last_mod_date = datetime(
                year=2022, month=8, day=3, tzinfo=timezone.utc)
    all_tutorial_pages = {
        'presentation': {"last_modified": last_mod_date},
        'pietons': {"last_modified": last_mod_date},
        'voitures': {"last_modified": last_mod_date},
        'velos': {"last_modified": last_mod_date},
        'transports-collectifs': {"last_modified": last_mod_date},
    }

    def get_context_data(self, **kwargs):
        this_page = kwargs.get('tutorial_page', 'presentation')
        if this_page not in self.all_tutorial_pages:
            # Trying to access a tutorial page that doesn't exist.
            raise Http404()
        context['active_tab'] = this_page # should really call this 'active_page', but whatever.
        self.template_name = f'mobilito/tutorial_{this_page}.html'

        pages_seen = get_cookie(self.tutorial_cookie_name, default=[])
        pages_seen.append(this_page)
        set_cookie(self.tutorial_cookie_name, page_seen, expire=2 hours)

        next_page = get_next_tutorial_page()
        if next_page:
            set_arrow_target(next_page)
        else:
            set_arrow_target(not_this_page(this_page))
        return context

    def get_pages_to_visit(self) -> list:
        """Return the list of pages to visit.

        This is used by the bottom arrow to redirect to the next page.
        """
        // This returns a list of pages visited.
        pages_seen = get_cookie('mobilito_tutorial')
        if user.is_auth:
            last_tutorial = mobilito.MobilitoUser.get().completed_tutorial_timestamp
        else or catch:
            last_tutorial = datetime(0)
        pages_to_visit = [page for page, timestamp in all_tutorial_pages
                          if page not in pages_seen and last_tutorial < timestamp]
        return pages_to_visit

    def get_next_tutorial_page(self) -> str:
        """Return the name of the next tutorial page to visit.

        This is used for the forward arrow.  Return None if the user
        has seen all tutorial pages.
        """
        pages_to_visit = get_pages_to_visit()
        if pages:
            return pages[0]
        return None

    def not_this_page(self, this_page) -> str:
        """Return the next tutorial page that is not this one.
        """
        return list(set(self.all_tutorial_pages.keys) - set(this_page))[0]
```

transport_nantes/mobilito/views.py
JeffAbrahamson: The change (although, cf my code suggestion that avoids all this complexity) is just to say "%F %T" instead of "%Y-%m-%d %H:%M:%S".

JeffAbrahamson:
commit message 3 - `s/hand logic/handle logic/;`

JeffAbrahamson (approved):
LGTM.

Issue #867, opened 2022-08-03
PressMention updates


JeffAbrahamson:
We should smooth out the final rough edges on PressMention.

* [x] We shouldn't upload a default image if we don't have one.  If there's no image, then there's no image.
* [x] The current views are all admin, so call them that.
* [ ] Make a user view at `/press/`, which we have but it doesn't look good.  Model it on [GP's view](https://www.greenpeace.fr/?s=&type=press_release), with an image on the left (if there is one) and text on the right.  There's is for press releases, but it's the same thing.
* [ ] We should fetch the image alt text if it's available and use it as alt text.  If we don't have alt text, we shouldn't make something up (currently, the site name), that doesn't help the visually impaired in the least.
* [ ] When displaying, we should display OG title and and description if we have it, falling back on what the user provided only if OG data is unavailable.  This is individually, of course, not both or nothing.
* [ ] In the `/press/admin/new/` view, we should drop the list of existing items.  It's confusing.  Instead, we should just have a [fa cross](https://www.iconfinder.com/icons/335691/x_cross_close_icon) on the left for cancel.
* [ ] The `presse/admin/detail/<id>` page shows two views but doesn't explain why.  It should explain why.  (I think it's just showing two different layouts, but this should be explained in terse but clear French why the person should care.)  I'm not immediately convinced we need both views, but I don't remember why we did this.
* [ ] On the `presse/admin/detail/<id>` page, show the user-entered data (labeled as such) and the opengraph data (labeled as such).
* [ ]

Shriukan33:
I couldn't find an example of website that has an alt in its image, at best I've seen LeMonde.fr using the description of the article as an alt tag, but this required to inspect the image, not the meta tag. 

I don't see an easy solution to alt tags, sadly. We already capture the title, website, and description, I believe that should be enough for visually impaired users as for context. If they click, their screen reader will describe the image if the website did.

Shriukan33:
>  The presse/admin/detail/<id> page shows two views but doesn't explain why. It should explain why. (I think it's just showing two different layouts, but this should be explained in terse but clear French why the person should care.) I'm not immediately convinced we need both views, but I don't remember why we did this.

I looked for it as well but couldn't find an instance where we use the bigger one. @mcmikashi , do you remember ? :)

mcmikashi:
The second views is the mobile version.

PR #868, opened 2022-08-03
Update PressMention

JeffAbrahamson:
* Label admin views as such.
* Drop uploading a fresh copy of a default view with every article
  missing opengraph data.  (It's wasteful for the server, confusing
  for the user.)
* If we've no image to show, show no image.  Show a non-breaking space
  instead just to avoid some browser shortcutting.
* Update a few comments.

Part of #867.


Shriukan33 (approved):
LGTM

Shriukan33 (commented):
Missing migration noticed after merge

PR #873, opened 2022-08-05
Benja mobilito enforce tutorial reading

Shriukan33:
This PR :

* tweaks the display o the  "Passer à l'enregistrement >>"  button to only show once one has read all tutorial's pages
* Force redirection to tutorial when trying to access Address form or Recording views

### DIFF
--- a/transport_nantes/mobilito/views.py
+++ b/transport_nantes/mobilito/views.py
@@ -67,14 +67,14 @@ class TutorialView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        self.this_page = kwargs.get('tutorial_page', 'presentation')
-        if self.this_page not in self.all_tutorial_pages:
+        this_page = kwargs.get('tutorial_page', 'presentation')
+        if this_page not in self.all_tutorial_pages:
             # Trying to access a tutorial page that doesn't exist.
             raise Http404()
-        context['active_page'] = self.this_page
-        self.template_name = f'mobilito/tutorial_{self.this_page}.html'
+        context['active_page'] = this_page
+        self.template_name = f'mobilito/tutorial_{this_page}.html'
 
-        pages_seen = self.get_seen_pages_from_cookie()
+        pages_seen = self.get_seen_pages_from_cookie(this_page)
         pickled_pages_seen = pickle.dumps(pages_seen)
         pages_seen = make_timed_token(
             string_key=b64encode(pickled_pages_seen).decode(),
@@ -94,11 +94,11 @@ class TutorialView(TemplateView):
                 user.first_time = False
                 user.save()
             context["seen_all_pages"] = True
-            context["the_next_page"] = self.not_this_page(self.this_page)
+            context["the_next_page"] = self.not_this_page(this_page)
 
         return context
 
-    def get_seen_pages_from_cookie(self) -> list:
+    def get_seen_pages_from_cookie(self, this_page=None) -> list:
         """Return the list of pages seen from the cookie."""
         pages_seen = self.request.COOKIES.get(
             self.tutorial_cookie_name, '')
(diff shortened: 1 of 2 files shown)


transport_nantes/mobilito/views.py
```diff
@@ -67,16 +67,14 @@ def get(self, request, *args, **kwargs):
...
         context = super().get_context_data(**kwargs)
-        this_page = kwargs.get('tutorial_page', 'presentation')
-        if this_page not in self.all_tutorial_pages:
+        self.this_page = kwargs.get('tutorial_page', 'presentation')
+        if self.this_page not in self.all_tutorial_pages:
             # Trying to access a tutorial page that doesn't exist.
             raise Http404()
-        context['active_page'] = this_page
-        self.template_name = f'mobilito/tutorial_{this_page}.html'
+        context['active_page'] = self.this_page
```
JeffAbrahamson, line 74: Why are you putting `this_page` on the object (on `self`) instead of passing it as an argument to `get_seen_pages_from_cookie()`?

In general, passing arguments makes dependencies clear, whereas assigning to global state (or to larger than necessary state) makes code harder to read, because it's hard to know where things come from, what came from what.

transport_nantes/mobilito/views.py
```diff
@@ -108,11 +107,17 @@ def get_seen_pages_from_cookie(self) -> list:
             if pickled_pages_seen:
                 pages_seen_list: list = pickle.loads(
                     b64decode(pickled_pages_seen))
+                if hasattr(self, 'this_page') and self.this_page not in pages_seen_list:
```
JeffAbrahamson, line 110: If you passed `this_page` as an argument, you could just test if it's `None` rather than doing the somewhat mystical `has_attribute`.

transport_nantes/mobilito/views.py
```diff
@@ -122,17 +127,20 @@ def get_pages_to_visit(self) -> list:
...
+        completed_tutorial_timestamp = None
         if self.request.user.is_authenticated:
             user = get_MobilitoUser(self)
             completed_tutorial_timestamp = user.completed_tutorial_timestamp
-        else:
+
+        if not completed_tutorial_timestamp:
             completed_tutorial_timestamp = datetime(
-                year=2022, month=8, day=3, tzinfo=timezone.utc)
+                year=2022, month=8, day=2, tzinfo=timezone.utc)
```
JeffAbrahamson, line 137: Where do these magic numbers (some day in August 2022) come from?

transport_nantes/mobilito/views.py
```diff
@@ -166,13 +174,21 @@ class AddressFormView(LoginRequiredMixin, FormView):
...
+        context = super(FormView, self).get_context_data(**kwargs)
         # Clear session data if user had filled the form before.
         self.request.session['address'] = None
         self.request.session['city'] = None
         self.request.session['postcode'] = None
         return context
 
+    def get(self, request, *args, **kwargs):
+        context = self.get_context_data(**kwargs)
+        if self.get_pages_to_visit():
```
JeffAbrahamson, line 186: This deserves a comment.  I _think_ what you mean is that if the user arrives without having specified a page, you'll 302 (don't 301!) to the correct URL.  But it is not obvious from reading the code why you would do that, even if I know because we've been talking about it.

transport_nantes/mobilito/views.py
```diff
@@ -166,13 +174,21 @@ class AddressFormView(LoginRequiredMixin, FormView):
...
+            return HttpResponseRedirect(
```
JeffAbrahamson, line 187: Unless you mean that the client's browser may, the next time, assume the same redirect, you should make sure you 302.

Indeed, specifying what you want instead of relying on the default value makes your code more robust to future changes in underlying libraries (including accidental ones).

transport_nantes/mobilito/views.py
```diff
@@ -217,6 +233,14 @@ def get_context_data(self, **kwargs) -> dict:
             f'id={session_object.id}')
         return context
 
+    def get(self, request, *args, **kwargs):
+        context = self.get_context_data(**kwargs)
+        if self.get_pages_to_visit():
```
JeffAbrahamson, line 238: comment, see above

transport_nantes/mobilito/views.py
```diff
@@ -217,6 +233,14 @@ def get_context_data(self, **kwargs) -> dict:
...
+            return HttpResponseRedirect(
+                reverse('mobilito:tutorial',
```
JeffAbrahamson, line 240: 302, see above

JeffAbrahamson (changes requested):

transport_nantes/mobilito/views.py
Shriukan33: HttpResponseRedirect is a 302.

transport_nantes/mobilito/views.py
Shriukan33: Added a comment, and HttpResponseRedirect has a status code of 302, to get a 301 code it would be a HttpResponsePermanentRedirect, which isn't appropriate here I agree.

transport_nantes/mobilito/views.py
Shriukan33: needed a date to start from and compare, not set in stone.

transport_nantes/mobilito/views.py
Shriukan33: This is done

transport_nantes/mobilito/views.py
Shriukan33: Moved to args, removed from self

JeffAbrahamson (approved):
LGTM.

PR #877, opened 2022-08-08
Improve press mention list view

mcmikashi:
* Add pagination
* Change the display of the press mention card

Part of #867


Shriukan33 (approved):
It looks a bit better, and nice that you used paginate !

JeffAbrahamson (approved):
LGTM.

The `paginate_by` causes the SQL to do a `LIMIT` so that, if we had millions of press mentions, we wouldn't attempt to fetch them all, right?

mcmikashi:
> LGTM.
> 
> The `paginate_by` causes the SQL to do a `LIMIT` so that, if we had millions of press mentions, we wouldn't attempt to fetch them all, right?

Yes

PR #878, opened 2022-08-09
Benja clean up publication model 869

### LINKED ISSUE #869: Clean up publication model for TB
This supercedes #543, #722, and #821.

We should not be modifying TBObjects except when we publish them or prepare to delete them.

We have four dates associated with each TBObject:
* `date_created`, which is automatically set on creation (only).
* `first_publication_date`, which we set the _first_ time we publish the object.
* `publication_date`, which we set _every time_ we publish the object.
* `scheduled_for_deletion_date`, which we set when the cleaner-upper decides the object is moribund.

Publishing should null out the `scheduled_for_deletion_date`.

The function that gets the currently visible version should take the most recently published version, if there is one, else conclude that none are published.

The list view should make clear each of these values.  It's currently rather confusing and hard to read.  So it should say something like
```
Créé le %F %T.
```
Then one of
```
Jamais publié.
{maybe}Signalé pour suppression le %F %T ({days} jours restants).
(ou)
<b>Publié</b> le %F %T.
Publié depuis le %F %T.  (but only if pub and first pub are different)
```

Shriukan33:
It's now easier to spot which instance is served using boldness and
color hint.
Also the "Publier" button only shows when this action is possible.

I also suggest a change in the list view layout, by centering it. It's responsive. 

Fix: Only the last item created would show, instead of the
last published one
Fix: Creation form would show scheduled_for_deletion_date field.

Closes #869

### DIFF
--- a/transport_nantes/topicblog/forms.py
+++ b/transport_nantes/topicblog/forms.py
@@ -51,7 +51,8 @@ class TopicBlogItemForm(ModelFormTemplateList):
         model = TopicBlogItem
         # Admins can still edit those values
         exclude = ('item_sort_key', 'user',
-                   'publication_date', 'first_publication_date', 'publisher')
+                   'publication_date', 'first_publication_date', 'publisher',
+                   'scheduled_for_deletion_date')
 
 
 class TopicBlogEmailSendForm(forms.Form):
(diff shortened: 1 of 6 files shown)


JeffAbrahamson (approved):
LGTM.

I rather suspect the model documentation needs updating...

PR #883, opened 2022-08-09
Add missing migration file after #868

Shriukan33:
Missing migration file from #868


JeffAbrahamson (changes requested):
Sorry, there was confusion here.  We definitely want the email contact in the press releases on the web.

What we'd also like is to be able to provide a contact phone number in the press releases we mail out.  But the phone number should only display in the mails, it should not show up on the web.

That will require a new field in the model, I think.  I think we discussed making this a single free-form field, so it can be "06-12-34-56-78" or "Jean-Yves Du Pont à 06-12-34-56-78" or whatever seems appropriate to the user.  But when it displays on the web, it's absent.

JeffAbrahamson:
> > Sorry, there was confusion here. We definitely want the email contact in the press releases on the web.
> > What we'd also like is to be able to provide a contact phone number in the press releases we mail out. But the phone number should only display in the mails, it should not show up on the web.
> > That will require a new field in the model, I think. I think we discussed making this a single free-form field, so it can be "06-12-34-56-78" or "Jean-Yves Du Pont à 06-12-34-56-78" or whatever seems appropriate to the user. But when it displays on the web, it's absent.
> 
> I think this isn't the right PR ?

That would be correct...

JeffAbrahamson (approved):
LGTM.

PR #884, opened 2022-08-11
Benja mail campaign summary 870

### LINKED ISSUE #870: Make email sending summary
We want to know, at the very least (one line per TBSendable),
* How many things sent
* How many failed
* How many opened
* How many clicks

If PII-authorised, allow click-through to see details and when things happened with further click-though to see events.

Shriukan33:
This PR adds a dashboard that allow authorized users to see metrics about mail campaigns, as well as details about users with either a search field or a click through. The table are sortable / filterable dynamically, and paginated.

Tests included. 

Preview video here: 

https://user-images.githubusercontent.com/70256364/184135682-0adf6e52-c4c3-491f-9eb7-278c1719919d.mp4

closes #870


JeffAbrahamson:
I need more time to digest this one.  The behaviour looks like what we want.  Sending email currently doesn't require admin, it just requires (I think) one of the "can send" permissions.  I hope, we generally won't want to grant admin permission to too many people...

Shriukan33:
> I need more time to digest this one. The behaviour looks like what we want. Sending email currently doesn't require admin, it just requires (I think) one of the "can send" permissions. I hope, we generally won't want to grant admin permission to too many people...

Yes, I don't think we have any permission that is "is admin" right now. 
For mail sending we have 4 : self send (press|Email) and send (press|email).
the lat**t**er is the permission needed to send to mailing lists.

It's a bigger PR I agree, was a fun one to do though ! It's something we've meant to add for a while.

transport_nantes/dashboard/views.py
```diff
@@ -27,3 +34,63 @@ class SignatureView(LoginRequiredMixin, FormView):
...
     form_class = SignatureForm
+
+
+class EmailCampaignsDashboardView(PermissionRequiredMixin,
+                                  LoginRequiredMixin,
+                                  ListView):
+    """Dashboard of email campaigns"""
+    # Login & Permission
+    login_url = reverse_lazy("authentication:login")
+    permission_required = ('topicblog.tbe.may_send', 'topicblog.tbp.may_send')
```
JeffAbrahamson, line 45: Doesn't this mean the user must have both permissions?  But surely having only one could permit being able to view the send records for that one type?

transport_nantes/authentication/models.py
```diff
@@ -39,6 +41,17 @@ def __str__(self):
             conf=confirmed, auth=auth,
             commune=self.commune, cp=self.code_postal)
 
+    def get_all_send_records(self) -> list:
```
JeffAbrahamson, line 44: The authentication model shouldn't know anything about send records or the TB module without some extensive comment that explains why we really couldn't get around it.

transport_nantes/dashboard/templates/dashboard/email_campaign_details.html
```diff
@@ -0,0 +1,46 @@
+{% extends 'dashboard/email_campaigns.html' %}
```
JeffAbrahamson, line 1: The dashboard app may not have been a good idea -- perhaps its very tiny content should have gone in `asso_tn`.  But things that are specific to other apps surely should stay with those other apps.

So this PR probably shouldn't touch dashboard except to provide a link from the dashboard page to the TB dashboard(s).  The point of apps is to provide abstraction and containment.

Think what you would do if you decided an app was really useful and should be shared with others.  Or think of other apps you've used.  You'd include directions to add a couple hooks into it, mostly via a `urls.py` inclusion and a clearly documented API for calling it from outside the app.  But, otherwise, everything needed would be in one place.

transport_nantes/dashboard/urls.py
```diff
@@ -1,8 +1,15 @@
...
 
 app_name = 'dashboard'
 urlpatterns = [
-    path('', DashboardIndex.as_view(), name='index'),
-    path('signature/', SignatureView.as_view(), name='signature'),
+    path('', views.DashboardIndex.as_view(), name='index'),
+    path('signature/', views.SignatureView.as_view(), name='signature'),
+    path('email-campaigns/', views.EmailCampaignsDashboardView.as_view(),
+         name='email_campaigns'),
+    path('email-campaigns/user/<int:pk>/',
```
JeffAbrahamson, line 10: I think this should index on slug and user_id.  (Cf. other comments about why I don't understand the need for the complexity or abstraction of `EmailCampaign` objects and also think they are modeling something we explicitly don't care about.)

transport_nantes/dashboard/urls.py
```diff
@@ -1,8 +1,15 @@
...
+         views.UserSendRecordsDetailView.as_view(),
+         name='user_send_records'),
+    path('email-campaigns/<int:pk>/', views.EmailCampaignDetailView.as_view(),
```
JeffAbrahamson, line 13: This should index just on slug (for marketing emails).

transport_nantes/dashboard/urls.py
```diff
@@ -1,8 +1,15 @@
...
-from .views import DashboardIndex, SignatureView
+from dashboard import views
 
 app_name = 'dashboard'
 urlpatterns = [
-    path('', DashboardIndex.as_view(), name='index'),
-    path('signature/', SignatureView.as_view(), name='signature'),
+    path('', views.DashboardIndex.as_view(), name='index'),
+    path('signature/', views.SignatureView.as_view(), name='signature'),
+    path('email-campaigns/', views.EmailCampaignsDashboardView.as_view(),
```
JeffAbrahamson, line 8: I'd call this something like `'sent'`, but that's a detail.  And then it would let you choose (for marketing email) whether you want emails or press releases, and for transactional the various sorts of transactional emails.  A tab line at the top of the page is probably the right interface to do that, but I'm pretty open.

For transactional emails, I can't think of a pertinent grouping off hand.  We'd just like to see recent transactional mails sent.

transport_nantes/topicblog/migrations/0059_emailcampaign_and_more.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 4.0.4 on 2022-08-09 16:00
```
JeffAbrahamson, line 1: Migration 0059 already exists on origin/master.

JeffAbrahamson (changes requested):
This is mostly really good work, thanks!

There are two technical issues, which I've noted in the review comments.

The first is that I think you went in a slightly incorrect direction in abstracting what we might want to see and understand.  The primary entity (for marketing emails) is the slug.  Things like mailinglist are useful for debugging and for eventual unsubscribe but aren't otherwise very pertinent.  I don't think this will be too hard to fix (though surely a bit nitty in the doing), since it just means not creating `EmailCampaign` objects and instead doing a query on `slug`.

The other is a simple abstraction violation that's easy to correct.  You put a bunch of code in the dashboard app, but this is all about TopicBlog, and so the code should go there.

There's an item that I still don't fully understand, which is why this all works even though you seem to have defined it in terms of the admin interface!  That perhaps warrants code comments so that future devs understand the relationship between something that's not part of admin and the admin interface.

JeffAbrahamson:
A bit of an aside, wouldn't this be more easily done with a three `ListView`s (one each for marketing email and press objects and another one (or three) for transactional mails, all using the same list template (with `nav nav-tabs`) and using a query something like

```
queryset = SendRecordXXX.objects.all().\
    values_list('slug', flat=True).distinct()
```

Click-through goes to another `ListView` with query along the lines of
```
queryset = SendRecordXXX.objects.filter(slug=this_slug)
```
and click-through from there to a `DetailView` with
```
context['object'] = SendRecordXXX.objects.get(id=the_id)
```
and that permits click-through to a final `ListView` with
```
queryset = SendRecordXXX.objects.filter(recipient=this_recipient)
```
(and that last with click-through to the `DetailView` above)?

Things like search are impressive but in the spirit of not building what we don't yet need, maybe the first version can even do without.

JeffAbrahamson:
Where you used DataTable, is there a reason not to use bootstrap?  It avoids adding a dependency.

* [sorting](https://mdbootstrap.com/docs/b4/jquery/tables/sort/)
* [filtering](https://mdbootstrap.com/docs/standard/extended/table-filter/) and [w3school](https://www.w3schools.com/Bootstrap/bootstrap_filters.asp) which is a bit clearer
* [pagination](https://mdbootstrap.com/docs/b4/jquery/tables/pagination/) and [also](https://getbootstrap.com/docs/4.0/components/pagination/)

Django already offers support for pagination.

Shriukan33:
this PR has been open for a while, I had to redo most of it, I propose we close this one

PR #886, opened 2022-08-11
Add a reminder and a pre-visualization of sent object

### LINKED ISSUE #870: Make email sending summary

Shriukan33:
To help the person who's sending, we added a reminder before sending the
object.

Closes #870

### DIFF
--- a/transport_nantes/topicblog/templates/topicblog/topicblogbase_send_form.html
+++ b/transport_nantes/topicblog/templates/topicblog/topicblogbase_send_form.html
@@ -9,6 +9,16 @@
 
 {% block content %}
 
+    <div class="col-10 mx-auto mt-2">
+        <p>
+            Envoi du slug :
+            <a href="{% url base_model.viewbyslug_object_url tbe_slug %}" target="_blank">
+                {{ tbe_slug }}
+            </a>
+            <span class="text-muted small"><i class="fa-solid fa-up-right-from-square"></i></span>
+        </p>
+    </div>
+    
     <form id="send-email-form" class="col-10 mx-auto mt-2"
     action="{% url send_to_view tbe_slug %}"
     method="POST">{% csrf_token %}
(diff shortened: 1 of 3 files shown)


transport_nantes/topicblog/views.py
```diff
@@ -572,6 +572,11 @@ def get_context_data(self, **kwargs):
             raise ImproperlyConfigured("base_model must be a class with a "
                                        "send_object_url attribute")
         context["send_to_view"] = self.base_model.send_object_url
+        context["base_model"] = self.base_model
+        context["sent_object"] = self.base_model.objects.filter(
```
JeffAbrahamson, line 576: Not hugely important, but isn't this the object we are _about to send_, not something _already sent_?

transport_nantes/topicblog/templates/topicblog/topicblogbase_send_form.html
```diff
@@ -9,6 +9,16 @@
 
 {% block content %}
 
+    <div class="col-10 mx-auto mt-2">
+        <p>
+            Vous envoyez le contenu suivant :
```
JeffAbrahamson, line 14: I think it would be helpful for you to read more about human interface design and its metaphors.  This isn't the sort of thing that blocks PR approval, but it will surely be the subject of a change request at some point soon when someone sees a better way / gets annoyed with the verbosity.

All information doesn't have to come in the form of declarative sentences, and I think if you explore a bit you'll see that most signals between you and your web interface are far, far more subtle.  This box I'm typing in has two tabs, "write" and "preview", neither says "You can click here to see what your text looks like".  Etc.

Wouldn't it have been enough to show the slug, perhaps something like
```
Envoie du slug <a href={% url ...>
```
and the user has seen the web before, sees that there's an underline, understands that it's therefore a link, that it will open something, understands about hover or touch feedback, etc.  Since it opens in a new tab, you could even add a "new tab" superscript (`fas-external-link`, I think) to make that clear.

Attention is a very precious commodity.  Anything that draws attention away from the task at hand and focuses it on administration is usually a negative.

JeffAbrahamson (approved):
LGTM.

transport_nantes/topicblog/views.py
Shriukan33: It is the _object about to be sent_

transport_nantes/topicblog/templates/topicblog/topicblogbase_send_form.html
Shriukan33: Applied this change, true I don't use icons nearly enough.
I'd rather start from overly explicit to less explicit, but I should try to come with a balance !

PR #887, opened 2022-08-16
Benja press mention updates 867 based on mickael

Shriukan33:
This is based on Mickael's PR #877, so better merging the latter before this one.

I checked every items in #867 issue, and added what wasn't included already in #877

In this PR :
* Removal of existing item list in creation form
* Addition of a back button from creation form
* Addition of OG-less preview in detail view

### DIFF
--- a/transport_nantes/press/templates/press/press_create.html
+++ b/transport_nantes/press/templates/press/press_create.html
@@ -3,6 +3,10 @@
 {% block content %}
 <div class="container-fluid pl-5 py-3">
     <div class="col-md-8 col-sm-12 mx-auto">
+      <a href="{% url 'press:list_items' %}" class="btn navigation-button mb-2">
+        <i class="fa-solid fa-arrow-left"></i>
+        Retour à la liste des articles
+      </a>
       <h5 class="alert alert-info text-center" role="alert">
         N'oubliez pas qu'il faut toujours respecter les droits d'auteur.
       </h5>
--- a/transport_nantes/press/templates/press/press_create.html
+++ b/transport_nantes/press/templates/press/press_create.html
@@ -2,28 +2,6 @@
 {% load crispy_forms_tags %}
 {% block content %}
 <div class="container-fluid pl-5 py-3">
-  <div class="row">
-    <div class="col-sm-12 my-2">
-      <div class="accordion" id="accordion">
-        <div class="card">
-          <div class="card-header" id="headingOne">
-            <h2 class="mb-0">
-              <button class="btn btn-link btn-block text-left" type="button" data-toggle="collapse" data-target="#publication-liste" aria-expanded="true" aria-controls="publication-liste">
-                Liste des publications deja présente
-              </button>
-            </h2>
-          </div>
-          <div id="publication-liste" class="collapse"  aria-labelledby="headingOne" data-parent="#accordion">
-            <div class="text-center mt-3">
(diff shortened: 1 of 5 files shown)


mcmikashi (approved):
LGTM.

JeffAbrahamson:
If we don't have OG data, we should fall back on what the user provided.

Do you see a customer need being satisfied by displaying a bear URL and nothing more?

image

JeffAbrahamson:
In `/`, the images don't link through to the article, but they are the primary click target.

JeffAbrahamson:
In `/`, the images don't link through to the article, but they are the primary click target.

JeffAbrahamson:
I'm stil confused why we display two version of each article in preview without explaining why.  Why isn't this just a reactive thing, and if I want to preview both, I'll change my screen size.  Otherwise, I just trust that we developed a site that functions correctly and I don't spend my time verifying reactive design.

In this particular example, note that the horizontal image display (first version) is different between the two.

image

JeffAbrahamson:
In `/admin/new/` we are creating an item, not viewing, so that "return" arrow is really weird.  It should be an `fa-cross` or something similar to suggest cancel.

JeffAbrahamson:
At `/admin/update/<id>` we don't show fetched OG data.

transport_nantes/press/templates/press/press_create.html
```diff
@@ -3,6 +3,10 @@
 {% block content %}
 <div class="container-fluid pl-5 py-3">
     <div class="col-md-8 col-sm-12 mx-auto">
+      <a href="{% url 'press:list_items' %}" class="btn navigation-button mb-2">
+        <i class="fa-solid fa-arrow-left"></i>
```
JeffAbrahamson, line 7: We probably shouldn't use `fa-arrow-left` when we mean to cancel.  It's really for going back.  I'll make an issue.

JeffAbrahamson (approved):
LGTM.

Shriukan33:
> LGTM.

Do you want me to merge, in spite of your previous comments ? I'm confused

Shriukan33:
> I'm stil confused why we display two version of each article in preview without explaining why. Why isn't this just a reactive thing, and if I want to preview both, I'll change my screen size. Otherwise, I just trust that we developed a site that functions correctly and I don't spend my time verifying reactive design.
> 
> In this particular example, note that the horizontal image display (first version) is different between the two.
> 
> image

Difference top / down is, I think, due to the amount of text in the description. 
And I agree it's kinda weird that if one is mobile and the other is bigger version, then why both are displayed at all time instead of being responsive, I'm a bit confused.

JeffAbrahamson:
> > LGTM.
> 
> Do you want me to merge, in spite of your previous comments ? I'm confused

Your call.  Many of the items I noted were satellite to the PR.  They need doing, but don't necessarily invalidate the PR.

JeffAbrahamson:
> > I'm stil confused why we display two version of each article in preview without explaining why. Why isn't this just a reactive thing, and if I want to preview both, I'll change my screen size. Otherwise, I just trust that we developed a site that functions correctly and I don't spend my time verifying reactive design.
> > In this particular example, note that the horizontal image display (first version) is different between the two.
> > image
> 
> Difference top / down is, I think, due to the amount of text in the description. And I agree it's kinda weird that if one is mobile and the other is bigger version, then why both are displayed at all time instead of being responsive, I'm a bit confused.

We should do what's right and what we need now, not what we thought was right and what we thought we needed at some past time.

PR #895, opened 2022-08-17
Make the go to record button more obvious

### LINKED ISSUE #890: Make button "Passer à l'enregistrement" more visible
Once the user has seen all pages from the tutorial, they are presented the first page again which can be confusing. 
The button is present on top, but barely visible.

The button may be added on top or beneath the arrow instead ?

Shriukan33:
The button was too little noticeable annd would cause confusion
when user landed on the same page they've seen before, because they
didn't realise that the button was there.

Closes #890


JeffAbrahamson (approved):
LGTM.

PR #896, opened 2022-08-17
Prevent refresh on overscroll in recording page

### LINKED ISSUE #891: Prevent down scroll from refreshing a page while recording
Refreshing puts counters to 0 and may make the user think they lost all progress. We should prevent this refresh.

Shriukan33:
According to this
https://usefulangle.com/post/278/html-disable-pull-to-refresh-with-css
this property should prevent the refresh to happen when the user is
scrolling the page.

closes #891


JeffAbrahamson (approved):
LGTM.

PR #897, opened 2022-08-17
Apply Simon's suggestions on Mobilito design

Shriukan33:
Closes #892

image

image

PR #876, opened 2022-08-08
Switch color patterns to stick to style guidelines

Shriukan33:
Buttons now behave like the other buttons on the website.
We turn background to white and turn whatever color was in background
as the font color.

Closes #875

PR #898, opened 2022-08-23
Add an option to prompt the user about their location

### LINKED ISSUE #894: Add geolocation on demand
Instead of the current address form, we should have a button that prompts user to agree to share its location.

Shriukan33:
This is still optional, but people can use their phone's location
to get their current location easily.

This branch uses Simon's style suggestion as described in the #897 PR, I think it looks better than the #876 version. It gives more that "app" feeling while sticking to Mobilitain's colors. 

Closes #894


JeffAbrahamson:
I've two concerns as I play with this and read the code.

JeffAbrahamson:
Concern 1.  There was this thing where I was getting multiple pins on the map.  If that happens to real users, we're toast, the app is discredited.  It'll be hard to recover if that perception makes its way around.

JeffAbrahamson:
Concern 2.  We're using the map to model addresses.  Addresses model buildings.  But roads and paths are the connections between buildings.  It's completely reasonable that a user places himself on a bridge
image

or in a park
image

or along a country lane
image

and in these cases there isn't a useful address.  What we want is to be able to find where the user put the pin and to be able to display that.

JeffAbrahamson:
Playing around with Vigilo, I see that they provide addresses like "chemin du machin", but, notably, they somehow remember where the pin is.

Is there a way we can let the user place (or, optionally, auto-place via geo) the pin, we recover some representation of where the pin is, but we also let the user add a comment.  So if the app recovers "chemin du machin", the user might modify that to say "chemin du machine, entre la D145 et la premiere route transversale" or whatever makes sense?

Shriukan33:
> Concern 1. There was this thing where I was getting multiple pins on the map. If that happens to real users, we're toast, the app is discredited. It'll be hard to recover if that perception makes its way around.

I fixed that bug, or at least I couldn't reproduce it.

> Concern 2. We're using the map to model addresses. Addresses model buildings. But roads and paths are the connections between buildings. It's completely reasonable that a user places himself on a bridge 
>[...]
> and in these cases there isn't a useful address. What we want is to be able to find where the user put the pin and to be able to display that.

The only reliable data we could use that I can think of right now would be latitude / longitude pairs. Those never fail and are consistent all across the globe.




> Playing around with Vigilo, I see that they provide addresses like "chemin du machin", but, notably, they somehow remember where the pin is.
> 
> Is there a way we can let the user place (or, optionally, auto-place via geo) the pin, we recover some representation of where the pin is, but we also let the user add a comment. So if the app recovers "chemin du machin", the user might modify that to say "chemin du machine, entre la D145 et la premiere route transversale" or whatever makes sense?

you can actually modify what's in the address field while keeping your pin in place right now.

What happens currently is that when you pin a place, its lat / lng pairs are sent to a geocoding API, that in return, gives you an address with words (street, suburb, city, number, all that stuff). I use response's content to override what's in the field.
Changing the field's content doesn't move the pin however, so you can freely add anything into it and have the pin at the same place.

Shriukan33:
> > you can actually modify what's in the address field while keeping your pin in place right now.
> > 
> > What happens currently is that when you pin a place, its lat / lng pairs are sent to a geocoding API, that in return, gives you an address with words (street, suburb, city, number, all that stuff). I use response's content to override what's in the field.
> > Changing the field's content doesn't move the pin however, so you can freely add anything into it and have the pin at the same place.
> 
> What is being persisted in the model?  The lon/lat pair or the address or other?
> 

The address fields are what's saved.

JeffAbrahamson:
So if the user puts the pin at a spot on a road with no buildings, will we accurately reproduce that spot?

Shriukan33:
> So if the user puts the pin at a spot on a road with no buildings, will we accurately reproduce that spot?

Pin wont move if you change the address field if this is what you ask ? We save the address fields (postcode, street, city) in the model. If you wanted to replace precisly the pin after it's being saved, we should also save lat/lng.

JeffAbrahamson:
> > So if the user puts the pin at a spot on a road with no buildings, will we accurately reproduce that spot?
> 
> Pin wont move if you change the address field if this is what you ask ? We save the address fields (postcode, street, city) in the model. If you wanted to replace precisely the pin after it's being saved, we should also save lat/lng.

Consider the three cases I mentioned in [concern 2](https://github.com/transport-nantes/tn_web/pull/898#issuecomment-1229169472).

1. Can you offer a reasoned defence that you are modelling the correct thing with the address, that the user's click was just an indicator to get the thing you really want to model, the address?
2. How is the user likely to react on seeing his/her recording confirmation if we place the pin in a goofy spot (or don't place it at all, and the user can clearly see that the address points to somewhere else)?
3. How do you think we want to use this session data?  In what ways would we want to use the exact spot the user indicated?  In what ways would we want to use the address?

JeffAbrahamson:
> > Concern 1. There was this thing where I was getting multiple pins on the map. If that happens to real users, we're toast, the app is discredited. It'll be hard to recover if that perception makes its way around.
> 
> I fixed that bug, or at least I couldn't reproduce it.

What was the fix?  Is it somewhere in the diff 3532884..04f90d7?

Why are you confident you've fixed it?  Did you repro before making a change and then it doesn't repro after?

Shriukan33:
> > > Concern 1. There was this thing where I was getting multiple pins on the map. If that happens to real users, we're toast, the app is discredited. It'll be hard to recover if that perception makes its way around.
> > 
> > 
> > I fixed that bug, or at least I couldn't reproduce it.
> 
> What was the fix? Is it somewhere in the diff [3532884](https://github.com/transport-nantes/tn_web/commit/35328842df552a0338ea6eb3d9560ae5383f8c24)..[04f90d7](https://github.com/transport-nantes/tn_web/commit/04f90d73d3de47637f78b1db3b2f43b689f9d38b)?
> 
> Why are you confident you've fixed it? Did you repro before making a change and then it doesn't repro after?

I did squash the changes at some points, but the fix essentially is here : 

https://github.com/transport-nantes/tn_web/blob/04f90d73d3de47637f78b1db3b2f43b689f9d38b/transport_nantes/mobilito/static/mobilito/js/geolocation_map.js#L58-L64

Instead of sending again an event while the previous one isn't processed, I pass until next event has returned. (L83)

https://github.com/transport-nantes/tn_web/blob/04f90d73d3de47637f78b1db3b2f43b689f9d38b/transport_nantes/mobilito/static/mobilito/js/geolocation_map.js#L71-L83

Shriukan33:
Update : 

Removed 4 model fields : address, city, postcode, country and replaced with 3 new fields : Location, latitude, longitude. 

Switch geocoding provider to Google maps. 

:warning:  This change requires you to put an API key into your settings_local.py
The change has to be reflected on infra so it's deployable.
:warning: Do not forget to protect the API key with domain restriction, or anybody could use it for their own profit. (I messaged you on slack about this)

I created my own API key with a free account, and the performance and user experience is much much better, Google answers almost immediately, compared to nominatim ! :slightly_smiling_face:

JeffAbrahamson (approved):
LGTM.

PR #899, opened 2022-08-23
Make selected templates rely on the displayed object

### LINKED ISSUE #862: Make Template / email template pairs for TBEmail and TBPress
As found in this PR 

https://github.com/transport-nantes/tn_web/pull/857#issue-1316956928

The TBE and TBP have hardcoded templates in the view, and their .template_name attribute is actually the one rendered in the Emails we send. 

We agreed on writing the pairs like this : 
```

    template_config = {
        'topicblog/content_email.html': {
            'email_template': 'topicblog/content_email_client.html',
            'default_choice': True,
            'user_template_name': 'Classique',
            'active': True,
            "fields": {
                'slug': True,
                'title': True,
                'subject': True,
                'body_text_1_md': True,
            },
            "optional_fields_for_publication":
                template_config_default['optional_fields_for_publication'],
            "one_of_fields_for_publication":
                template_config_default['one_of_fields_for_publication'],
            "dependent_field_names":
                template_config_default['dependent_field_names'],
        },
    }
```

Shriukan33:
As explainned in #862, templates were hardcoded.
TBEmails would have for example "topicblog/content_email_client.html" as
their template_name attribute, and this would be overriden by the class
that would display it, for instance, TopicBlogEmailView, with the
hardcoded value "topicblog/content_email.html".

This commit fix that behavior and ensure that the current objects
are still accessible.
Indeed, many currently saved objects have some variation of a template
made for email clients in their template_name attribute, but these
wont work in a web environnement.
To prevent all current objects with invalid template_name attribute,
I added a check that verifies that the value in "template_name" is
also in its class' "template_config" dictionary.
In the event this check fails, then the default template is used, and
the object is edited so it has a valid template.

This acts as a fallback mecanism that will prevent the server from 500ing
if an incorrect template is provided.

Closes #862


JeffAbrahamson:
Commit message suggestion: "this commit fix*es* ..." and "wont" -> "won't" and "mec*h*anism".

What does "and ensures that the current objects are still accessible" mean?

The phrase "To prevent _all current objects_ with invalid ..." is French ("...tout objet actuel...").  I think you want to say "To prevent an object with invalid ..." or something like that.  (I'm actually not 100% sure what you mean, but I think I almost understand.)

JeffAbrahamson:
The code looks good.  Does it perhaps warrant a test?

Shriukan33:
> Commit message suggestion: "this commit fix_es_ ..." and "wont" -> "won't" and "mec_h_anism".
> 
> What does "and ensures that the current objects are still accessible" mean?
> 
> The phrase "To prevent _all current objects_ with invalid ..." is French ("...tout objet actuel..."). I think you want to say "To prevent an object with invalid ..." or something like that. (I'm actually not 100% sure what you mean, but I think I almost understand.)
> 
> Usually one is encouraged to write declarative commit messages and to avoid first person / narrative. It's not about what you did, it's about what the commit _is_.

Hmm I may have wrongly worded it.

Let me rephrase :

In the current build, all TB objects are displayed using their `.template_name` attribute. The problem is, in TBE and TBP, the `.template_name` attribute is set to the *email version* ! It only displayed right on web because the view was overriding `self.template_name` with an hardcoded value.

The new behavior will always use the object's `.template_name` attribute for web, and will look at its `email_template` counterpart  in the `template_config` dict 

However, it doesn't change the fact that all previously created objects are still using the email version as their `template_name` attribute, and trying to display them on web wont work.

To prevent the older objects from raising 500 because we try do display an email template in the web (That's what I meant by allow them to remain accessible), I added a check that verifies that the object's `.template_name` attribute is present in its class' tempalte_config. If it's not, the old template (which is the email version) is replaced with the default one.

JeffAbrahamson:
> > Commit message suggestion: "this commit fix_es_ ..." and "wont" -> "won't" and "mec_h_anism".
> > What does "and ensures that the current objects are still accessible" mean?
> > The phrase "To prevent _all current objects_ with invalid ..." is French ("...tout objet actuel..."). I think you want to say "To prevent an object with invalid ..." or something like that. (I'm actually not 100% sure what you mean, but I think I almost understand.)
> > Usually one is encouraged to write declarative commit messages and to avoid first person / narrative. It's not about what you did, it's about what the commit _is_.
> 
> Hmm I may have wrongly worded it.
> 
> Let me rephrase :
> 
> In the current build, all TB objects are displayed using their `.template_name` attribute. The problem is, in TBE and TBP, the `.template_name` attribute is set to the _email version_ ! It only displayed right on web because the view was overriding `self.template_name` with an hardcoded value.
> 
> The new behavior will always use the object's `.template_name` attribute for web, and will look at its `email_template` counterpart in the `template_config` dict to produce emails.
> 
> However, it doesn't change the fact that all previously created objects are still using the email version as their `template_name` attribute, and trying to display them on web wont work.
> 
> To prevent the older objects from raising 500 because we try do display an email template in the web (That's what I meant by allow them to remain accessible), I added a check that verifies that the object's `.template_name` attribute is present in its class' template_config. If it's not, the old template (which is the email version) is replaced with the default one.

That does make much more sense!

You might be able to take advantage of the fact that there's currently only _one_ template to propose a fix for the existing entries and so avoid the need to carry forward the introspection hack for the rest of time.

Shriukan33:
> You might be able to take advantage of the fact that there's currently only _one_ template to propose a fix for the existing entries and so avoid the need to carry forward the introspection hack for the rest of time.

It may seem hacky, but I really see it as a fallback if we someday delete a template.

What do you suggest ?

JeffAbrahamson:
> > You might be able to take advantage of the fact that there's currently only _one_ template to propose a fix for the existing entries and so avoid the need to carry forward the introspection hack for the rest of time.
> 
> It may seem hacky, but I really see it as a fallback if we someday delete a template.
> 
> What do you suggest ?

If that is your reasoning, then at the very least update the code and commit comments to explain this so that the reader is thinking "ah, this is a mechanism to permit retiring templates!" rather than "ah, this is legacy code designed to support templates created before this code".

Shriukan33:
> > > You might be able to take advantage of the fact that there's currently only _one_ template to propose a fix for the existing entries and so avoid the need to carry forward the introspection hack for the rest of time.
> > 
> > 
> > It may seem hacky, but I really see it as a fallback if we someday delete a template.
> > What do you suggest ?
> 
> If that is your reasoning, then at the very least update the code and commit comments to explain this so that the reader is thinking "ah, this is a mechanism to permit retiring templates!" rather than "ah, this is legacy code designed to support templates created before this code".

image

Alright, I'll rephrase all that ! :slightly_smiling_face:

Shriukan33:
@JeffAbrahamson Used my rephrase to make a new commit message, and added a comment in view to explain further what it does.

transport_nantes/topicblog/views.py
```diff
@@ -222,10 +226,25 @@ def get_context_data(self, **kwargs):
...
-        self.template_name = tb_object.template_name
+        # We check that the object's template is valid and configured, or
+        # fallback to a default template.
+        # Doing so we make sure that if a template is retired, we can keep
+        # serving the content without raising a 500 error.
+        if tb_object.template_name in tb_object.template_config:
+            self.template_name = tb_object.template_name
+        else:
+            logger.info(
+                f"Current template ({tb_object.template_name}) is not configured."
```
JeffAbrahamson, line 237: I'd say this is an error, no?

It's good we catch it, but we should get a rollbar alert if it ever happens.

JeffAbrahamson: After all, one should assume that changing the template could do really horrible things to what is displayed: different assumptions about what fields are displayed and in what juxtaposition, etc.

JeffAbrahamson (commented):
Overall looks good, modulo the info/error noted.

The commit comment continues to talk about "old objects" and not about the issue that I think you mean to discuss, which is objects (of any age) that have had their template or template config disappear behind their backs.  And that's something you want to catch, but it's nonetheless a pretty grave error, and I think we should be clear that we're saving ourselves from a 500, not serving good content.

Indeed, the more I think about it, the more I think we _should_ 500 if this ever happens.  Because if it happens, it's git-level stuff that needs to be rolled back asap, there's no benefit letting it quietly log or quietly log and rollbar.

Shriukan33:
It isn't a git stuff, it's the object's  `.template_name` attribute's value that is wrong (contains incorrect template). So this is fixable in the database.

If you really want to 500, then every TBEmail and TBPress need to be updated to use the correct template, or they'll raise a 500 every time you try to access them.

JeffAbrahamson:
Ah, I thought you meant the `.template_name` in the config.

It's trivial to update all the entries to have the same value, a simple `UPDATE WHERE`.  Let's just do that then.  (I can do it if you want, once you make clear what needs to happen.  The commit comment should make clear that the commit requires a simultaneous SQL command.  Or maybe there's a way to use the django migration mechanism to do updates like this so it's more atomic.)

JeffAbrahamson:
The main point here is the question "is this a good design, something we should tell our children about?"

If I understand you correctly, you're proposing that we should maintain code to handle the case where certain types of objects don't have a valid template name field.  But if they don't, bad things might happen anyway, such as we fall back to displaying things with the wrong template, which could be themed to something totally different.  So better just to enforce in code that the template name must be present.

And we can do that easily now, because we only have one template for anything (except TBItem, where there's one vestigial template, something I should clean up one of these days).

Shriukan33:
The update should be : 

update tbpress
set template_name = 'topicblog/content_press.html'
where template_name = 'topicblog/content_press_mail_client.html'

and update tbemail
set template_name = 'topicblog/content_email.html'
where template_name = 'topicblog/content_email_client.html'

So wherever there is a 'content_<type>_client' you replace it with 'content_<type>'

and yes, the commit and code should reflect those changes, I'll get to it a bit later, tell me if it looks ok sql wise ?

JeffAbrahamson:
> The update should be :
> 
> ```sql
> update tbpress
> set template_name = 'topicblog/content_press.html'
> where template_name = 'topicblog/content_press_mail_client.html'
> ```
> 
> and
> 
> ```sql
> update tbemail
> set template_name = 'topicblog/content_email.html'
> where template_name = 'topicblog/content_email_client.html'
> ```
> 
> So wherever there is a `'content_<type>_client'` you replace it with` 'content_<type>'`
> 
> and yes, the commit and code should reflect those changes, I'll get to it a bit later, tell me if it looks ok sql wise ?

LOoks almost right, just misses the Django name mangling.  I'll fix it and we'll test out together on beta.

There's not a migration mechanism for this?  I thought I once saw something, but my memory is vague and might be wrong.

JeffAbrahamson:
I think what I was thinking of is called a [data migration](https://docs.djangoproject.com/en/4.1/topics/migrations/#data-migrations).

Shriukan33:
> I think what I was thinking of is called a [data migration](https://docs.djangoproject.com/en/4.1/topics/migrations/#data-migrations).

added a migration for that, thanks, didn't know this was a thing ! Definitely better than using SQL queries in db.

Shriukan33:
I also removed the fallback to default mechanism to let it 500 if an invalid template is used. But at least it has logs now :)

JeffAbrahamson:
LGTM.

Issue #832, opened 2022-06-28
Improve mailing list dashboard


JeffAbrahamson:
This: https://www.mobilitains.fr/ml/list

* The frequency is too prominent, it should just be a parenthesised annotation.
* The state should be a checkbox and active should precede inactive.
* If the user has GDPR rights (I forget what we called it), they should be able to click through and see the user list.
* If the user has GDPR rights, they should be able to click through on individual users to see the send record summary for that user (over all lists, labelled by list).

Let's use this as an opportunity to reveal the newer elements of the data models, like `linked_article` and such.

Note that we intend to deprecate the `Petition` model in favour of `MailingList` with `mailing_list_type == PETITION`.  So this refactor shouldn't worry about the `Petition` model.

This should be pretty simple, since we aren't proposing editing, it's just three ListView subclasses.

PR #900, opened 2022-08-24
Apply cosmetic changes to ml/list

Shriukan33:
Made the table narrower and the frequency field not as big as before.
Also made the "État" field a checkbox instead of a string.

Part of #832

The other demands in the #832 issue are very reminiscent of what I've done in #884, what do you think ?


JeffAbrahamson (approved):
LGTM.

Issue #831, opened 2022-06-28
Improve dashboard usability


JeffAbrahamson:
Let's reformat the page to be a simple table, presented cleanly.

Maybe prefer fa-icons like on press mention rather than words?

Let's also add some text so that new users of the page get some reminder of what each thing is.  Ideally, we should take into account whether or not the user has the right to click before showing a thing to click.

Shriukan33:
> Maybe prefer fa-icons like on press mention rather than words?
> 
> Let's also add some text 

Text or icons ..?

JeffAbrahamson:
text, in the sense of "TBItem is a blog page ..." or "TBEmail represents an email that we will send to some list of users" etc.  We'll refine it with user feedback if/when we have users.

PR #908, opened 2022-08-25
Switch main dashbaord to a table

Shriukan33:
This table regroups everything the former dashboard was doing, but in a
more compact way.
It is responsive.

It doesn't handle permissions checks before displaying, this is coming a bit later so it's a lighter PR to digest. 

Part of #831

### DIFF
--- a/transport_nantes/dashboard/templates/dashboard/index.html
+++ b/transport_nantes/dashboard/templates/dashboard/index.html
@@ -1,46 +1,109 @@
 {% extends 'asso_tn/base_mobilitain.html' %}
 
-{% block content %}
+{% block styles %}
+	{{ block.super }}
+	<style>
+		tbody > tr > th[colspan="3"] {
+			text-align: center;
+			background-color: #35353547
+		}
+		thead > tr > th {
+			text-align: center;
+			font-weight: bold;
+			font-size: 1.2em;	
+		}
+	</style>	
+{% endblock styles %}
+
+{% block app_content %}
 <div class="container">
-	<div class="row pb-3">
-		<div class="col-sm-6">
-			<h3>Contact</h3>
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
The code looks good.

Can you think of many dashboard that have a big title "dashboard" at the top?  Usually things try to present themselves well enough that you don't have to say "this is a loaf of bread".

I think the internal names for things, based on our usage, is stuff like `TBItem` etc.  But that's not terribly useful to outsiders, so maybe we could use simple French words: `blog`, `mail`, `CP`, `carré`, `inscription`.  What would you call these things when talking to a non-dev?

It's a strange abstraction to a non-developer to think that an email is also a blog post and that a blog post is or isn't an email.  Think how a volunteer or non technical staffer would view these things.  On that subject, you've organised it the way our code is organised, but few non-dev's would find that logical, I expect.

```
blog     Articles de blog
mail     Marketing mails qui peuvent également être visionné sur web
CP       Communiqués de presse
carré    Un carré qui peut être présenté dans le rubrique de nos projets
inscription   Pages qui incite le visiteur de s'inscrire une liste de diffusion (y compris pétition)
presse   Mentions des Mobilitains dans la presse
diffusion   Listes de diffusion, de newsletters et pétitions
```

Why is the list view icon for mailinglists different from the list view icon for everything else?  Indeed, I'd say it should have a "+" icon, too, just inactive (because we haven't implemented that yet).

JeffAbrahamson (changes requested):
See long-ish comment in the conversation pain.  It's just about presentation, the code otherwise looks good.

Shriukan33:
Applied your suggestions : 
image

JeffAbrahamson (approved):
LGTM.

PR #918, opened 2022-09-04
Jeff 202209 mobilito

JeffAbrahamson:
Jeff mobilito changes for benja to approve.


transport_nantes/transport_nantes/settings.py
```diff
@@ -46,6 +46,7 @@
     'topicblog.apps.TopicBlogConfig',
     'dashboard.apps.DashboardConfig',
     'utm.apps.UtmConfig',
+    'gom_jabbar.apps.GomJabbarConfig',
```
Shriukan33, line 49: At this point of review (4bea2cfd3c890417951e3c788f09e742f0361173), it's not clear why we are duplicating the UTM-now-Visit app, you mention it's staging for another process but I think the commit comment could be clearer : 
- While I like the nerd reference to Dune, it's not very obvious at first glance that it aims to differentiate humans for robots unless you googled what gom_jabbar is
- I wish the the problem you're trying to solve doing this was specified in the commit message, because right now, I only know that this is a preliminary step to another change.

transport_nantes/gom_jabbar/models.py
```diff
@@ -60,3 +63,19 @@ class Visit(models.Model):
     user_is_authenticated = models.BooleanField(blank=True, default=False)
 
     timestamp = models.DateTimeField(auto_now=True)
+
+
+class Visitor(models.Model):
+    """Track visitors to the site.
+
+    A visitor is a set of statisticcs concerning a set of visits with
```
Shriukan33, line 71: statisticcs -> statistics

Shriukan33: Also, a single visitor instance has 2 IntegersFields.
I don't understand how this works ? Did you mean to increase that counter ? Like if one person visits the website and is human, you can tell they visited the website X times, X being Visitor.visits_from_humans ?

Please give me an example of how you would like to use this model, I'm having troubles seeing the use cases here

transport_nantes/press/views.py
```diff
@@ -128,7 +128,7 @@ def fetch_opengraph_data(request, url=None, is_view=False) \
 
     try:
         response = requests.get(url)
-        tree = html.fromstring(response.content)
+        tree = html.fromstring(response.content.decode("utf-8"))
```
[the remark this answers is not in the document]
JeffAbrahamson: Is the issue that you don't understand about unicode vs utf-8 encoding?

You might read [this](https://stackoverflow.com/a/643810/833300) and [this](https://stackoverflow.com/a/27939161/833300).  There's plenty more reading to do to really understand the topic.

I don't think it needs a comment to state "we fetched something over http, it's (probably) utf-8 encoded and we need to decode that".  That should be obvious to any experienced programmer who has had to work with text.

JeffAbrahamson:
I claim this should now be approvable up to this commit:
```
55d2efe|2022-09-13|Jeff Abrahamson| UTF-8 decode text from read web pages
```
I'm going to push off requesting approval on the UTM/Gom Jabbar part until a bit later.  You're quite right that it merits actually being used and better commented first.

Shriukan33:
I was confused because the prints and typo are present in 2cfc6e2d2a7676fceab412d4203712d6fa300d39, but this is fixed in 55d2efe ! :slightly_smiling_face: 
LGTM up to that point then !

JeffAbrahamson:
Closing this PR, having merged the commits through the utf bug.
Will submit a new PR with the utm -> gom jabbar changes later.

Issue #932, opened 2022-09-19
Geolocalisation doesn't work correctly on mobilito


JeffAbrahamson:
Clicking the arrow on the map localises (poorly, sometimes, but that's up to device GPS).  But it's modal.  Modal?  It's an on-off switch.  It should localise me and be done with it, not cause me to jump around with every GPS update, maybe I saw ok just as I teleport somewhere.

In addition, when it's wrong (always, for me, by 50 m or so, when I pull my phone out of my pocket and go to record), I should be able to tap the map to say "this is where I really am".  But I can't.  I can only go back into random teleport mode.

PR #936, opened 2022-09-26
Allow user to select on position on map without locating themselves

Shriukan33:
We used to ask for location permission when the user clicked on the "Je partage ma position" button. This was not ideal because the user couldn't use the map to set their location without giving us permission to track them.

The map is now centered on Nantes unzoomed by default, and user is now only located once they click the control button to locate themselves on the map.

### DIFF
--- a/transport_nantes/mobilito/static/mobilito/css/geolocation_form.css
+++ b/transport_nantes/mobilito/static/mobilito/css/geolocation_form.css
@@ -12,7 +12,6 @@ footer {
     display: none;
 }
 #map {
-    display: none;
     margin-bottom: 1em;
     height: 300px;
     width: 100%;
--- a/transport_nantes/mobilito/static/mobilito/css/geolocation_form.css
+++ b/transport_nantes/mobilito/static/mobilito/css/geolocation_form.css
@@ -18,3 +18,7 @@ footer {
     border: 1px solid black;
     color: black;
 }
+div.leaflet-control-locate > a > span {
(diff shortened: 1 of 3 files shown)


JeffAbrahamson (approved):
Please add an issue number before merging.

There's an ongoing bug that the map is not shown (and there's still a button offering to localise).  And that the arrow button is modal, which is thoroughly bizarre.  For example, if I click the arrow to localise myself, it gets it wrong, I click the map, I've now localised myself correctly, a second later the map jumps me back to the wrong spot.

Shriukan33:
I've fixed the modal behavior and "random teleport". 
Also I changed the icon to be more what I would expect from a button that brings me to my location.

JeffAbrahamson:
First commit comment:

--- 
Allow user **to** point to a position on **the** map ~~without giving~~**to provide** device location
    
We used to localise the user on clicking the
"Je partage ma position" button. This was not ideal because
it introduced an unuseful extra step (revealing the map)

and because the user should be able to position himself
on the map without triggering geolocalisation.
    
With this commit the map is centred on Nantes at wide
(departement) zoom.  The user geolocalises only on clicking
the location control.
    
**P**art of #932**.**

--- 

Note that _we_ did not ask for permission.  That is the browser.

JeffAbrahamson:
LGTM, modulo the commit message.

In the second commit, you're missing a period at EOL on the last line.  And it has an extra space.

PR #938, opened 2022-09-26
Benja mobilito session summary page 927

Shriukan33:
This PR adds a page that allows users to see session results : 
* In any case if the session is published (which is the default)
* Or only if you are either the author or allowed to see unpublished sessions

I find the summary page a bit empty, not very fun. I'd like to get suggestion to make it better looking.

Sessions are accessible using /mobilito/session/<uuid>

The UUID is retrievable using Admin panel for now.


JeffAbrahamson:
* Please fix spelling in our commit messages.
* "a URL" not "an url", unless you're into that turn-of-the-century "earl". ;-)
* I don't think you need to provide doc links for data migrations in commit comments.  It's good that you've learned this, but you can generally assume that competent developers know how to look up docs.  (It's good that you note it requires three migrations because of this, however.)  That said, your code doesn't provide that history: the migrations appear from nowhere.  It requires three migrations, which to my mind requires three commits.  But not a huge deal, one figures it out.
* The last commit needs tests.

JeffAbrahamson:
I thought I commented once more here, but I don't see it now...

For debugging purposes, it's important that we as devs can construct URLs.  But the way UUID's are stored in the db is as hex strings that are not how they are expected to be rendered in URLs.

image

http://localhost:8000/mobilito/session/38d2f91ba23346dfaf1c13dfe7b0e2ee

Shriukan33:
> * Please fix spelling in our commit messages.
> 
>     * "a URL" not "an url", unless you're into that turn-of-the-century "earl". ;-)
>

I couldn't find the spelling or "an url", mind point me to it please ?

Shriukan33:
> I thought I commented once more here, but I don't see it now...
> 
> For debugging purposes, it's important that we as devs can construct URLs. But the way UUID's are stored in the db is as hex strings that are not how they are expected to be rendered in URLs.
> 
> _image_
> 
> http://localhost:8000/mobilito/session/38d2f91ba23346dfaf1c13dfe7b0e2ee

I mentioned that you can actually fetch the UUID string from the admin panel, it's in its string format. 

image


Alternatively, you can also use Python : 
```python
>>> import uuid
>>> str(uuid.UUID("38d2f91ba23346dfaf1c13dfe7b0e2ee"))
'38d2f91b-a233-46df-af1c-13dfe7b0e2ee'
```

Shriukan33:
Added tests for ability to see published / unpublished Sessions.

JeffAbrahamson:
It is possible to go through the admin interface, but this cumbersome and requires that devs remember this.  The mapping goes both ways.  Sometimes we'll see something in the db and want to view the page.  Sometimes we'll see a page and want to check the db record.

And sometimes we'll want to do this in an automated way, such as
```
SELECT
  'https://www.mobilitains.fr/mobilito/session/' || session_uuid || '/'
FROM mobilito_session
WHERE start_timestamp > '2022-09-27';
```

Asking a dev to look these things up manually in the admin interface is a recipe for cursing.

And we can avoid this by having the same value in the db as in the URL.

One way, though it will require testing on postgres (and confirming in the docs to make sure it's a spec'd behaviour and not just a coincidental behaviour), would be to make the URL be `uuid.hex`.

JeffAbrahamson:
Perhaps use `.hex` (cf. issue).

Please don't name fields "uuid" any more than naming a field "id" is a good idea.  Inevitably you will have a second uuid for some reason, and then you'll hurt.

You could say `session_uuid`, for example.

Shriukan33:
I opened a new PR that uses SHA1 instead, is it made me redo almost all commits, I thought it would be easier to do another branch;

PR #944, opened 2022-09-28
Benja session summary sha1 edition 927

Shriukan33:
This PR is a follow-up to #938 
The main difference is that it uses sha1 instead of uuid. 

> This PR adds a page that allows users to see session results :
> 
>     In any case if the session is published (which is the default)
>     Or only if you are either the author or allowed to see unpublished sessions
> 
> I find the summary page a bit empty, not very fun. I'd like to get suggestion to make it better looking.

Sessions are accessible using /mobilito/session/**<session.session_sha1>**


JeffAbrahamson (approved):
If you hit the case where you modify the sha1, we should log it with a warning.  This is so unlikely, we want to be able to discover it.  It's also so unlikely we expect it never to happen.

Otherwise, LGTM.

Shriukan33:
Added warning logs, so we can see a duplicate sha1 happen IRL if this ever happens :)

PR #951, opened 2022-10-04
Add test that checks for missing migrations

Shriukan33:
We sometimes forget to add migrations to commits, this test will check for that.

### DIFF
--- a/transport_nantes/transport_nantes/tests.py
+++ b/transport_nantes/transport_nantes/tests.py
@@ -1,0 +1,8 @@
+from subprocess import run
+from unittest import TestCase
+
+
+class MissingMigrationsTest(TestCase):
+    def test_missing_migrations(self):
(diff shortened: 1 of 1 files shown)


Shriukan33:
(the failing test is confirmation that it works ! \ o / )

JeffAbrahamson:
This won't reject a commit.

It also doesn't pass its own test. ;-)

Shriukan33:
the test fails because on this branch, the migration is missing, so this is actually a passing test if you ask me :) I'd be worried if the test showed green while a migration is indeed missing ! 

And no, it doesn't reject the commit, but at least you know before merging that a migration file is missing.

Shriukan33:
@JeffAbrahamson  Can I merge this non-passing-tests-but-it's-actually-a-good-thing ? :)

JeffAbrahamson (approved):
LGTM.

Issue #927, opened 2022-09-19
Create a session summary page


JeffAbrahamson:
Initially, it's pretty much the same as the summary mail.

If the session is auth'd to the same user id as did the recording, then show the session summary regardless of the following.

If not, then only show if all of the following are true:
* The session auth timestamp is set.
* The session has closed.
* The session published boolean is set (field to add in this issue).

Initially, we'll set the published flag to `True` by default.  It's a way for us to take stuff offline quickly if we find some abuse.  Planning ahead.

Shriukan33:
Question about the URL path here : 
- Should we make the most basic URL with /mobilito/session/<id> with <id> being the object's database ID (Integer) ?
- From where do you access this link ? E-mail only or a dedicated page ?
- If I understand this correctly, you can actually share the session with anyone as long as it matches the 3 criteria (if you're not the author). Does the author still access their session if we toggle the published field to off ?
- We could use UUID to make URLs (to avoid db mapping with bruteforce), if we make it a primary key, we can either erase all data or we could do some gymnastic to transfer the id from id to uuid and the re-set foreign keys (https://blog.hexack.fr/en/change-the-primary-key-of-a-django-model.html)

Shriukan33:
> Question about the URL path here :
> 
>     * Should we make the most basic URL with /mobilito/session/ with  being the object's database ID (Integer) ?
> 
>     * From where do you access this link ? E-mail only or a dedicated page ?
> 
>     * If I understand this correctly, you can actually share the session with anyone as long as it matches the 3 criteria (if you're not the author). Does the author still access their session if we toggle the published field to off ?
> 
>     * We could use UUID to make URLs (to avoid db mapping with bruteforce), if we make it a primary key, we can either erase all data or we could do some gymnastic to transfer the id from id to uuid and the re-set foreign keys (https://blog.hexack.fr/en/change-the-primary-key-of-a-django-model.html). Wdyt ?

JeffAbrahamson:
> Question about the URL path here :
> 
>     * Should we make the most basic URL with /mobilito/session/ with  being the object's database ID (Integer) ?

No, because we don't want to reveal things like how much people use the service, how many observations there are, etc.



>     * From where do you access this link ? E-mail only or a dedicated page ?

How people find the link is a different PR.  This PR is about making the page.

We'll use it in mails, in social media, etc.

 
[...]

PR #952, opened 2022-10-04
Add missing migration

Shriukan33:
Part of #927


Shriukan33:
> LGTM, but please reference the correct commit in the git comment.

So I checked the commit and indeed, the 3 migrations are there about the sha1. 
The difference is that somewhere along the road we had a "editable=False" that makes read only a field (here, sha1) and dropped that. 
So the solution would actually be either to make the field read-only again (by adding editable=False), or to make the migration that makes it editable. 

What are you thoughts on it ? I don't think the sha1 should be editable once created, I hardly see in what case we would edit one.

Issue #928, opened 2022-09-19
Add report abuse button on session page


JeffAbrahamson:
This follows on to #927.

Do a quick survey of how such buttons are presented on other sites.  It needs to be easily findable if looking for it, but subtle enough not to steal attention.  Consider at the very least how twitter, facebook, stackoverflow and such do it.

If someone clicks it, use javascript to capture a phrase about what's wrong.  (Try some of them elsewhere to see how it works.  A reasonable site lets you unclick it.)

We need a model to track these reports.  As a first draft, let's do this:
* mobilito session id
* django session id
* django user id if available
* timestamp of report
* string reason for report

When a user visits a summary page, we should do a quick check if they've already reported the page.  If yes, we should make sure there's (subtle) visible feedback so that they don't feel a need to report it again.

For the moment, let's not worry about what to do with the data.  We don't have that problem yet.

Issue #173, opened 2021-08-12
Add the session cookie to TrackingProgression


JeffAbrahamson:
Help us do matching on events, see if users return / attempt frequently or if they're all different.

Shriukan33:
So i've played a bit with the cookies to see how it works. 
What do you suggest ? Creating a random string if none is present in the cookies and save it along the other information in the tracking model ? 
Should we set an expiry date ? Default is 14 days apparently 
Would a random 20 long alphanumerical string would do ?

JeffAbrahamson:
The cookie name has to be identifiable, otherwise you can't recover it.  Perhaps something like "tn_nav".  Django just represents it as a (key, value) pair in a dict, roughly.  As is natural in python.

Perhaps use `make_timed_token()` with a value of 6 months.

Shriukan33:
> The cookie name has to be identifiable, otherwise you can't recover it. Perhaps something like "tn_nav". Django just represents it as a (key, value) pair in a dict, roughly. As is natural in python.
> 
> Perhaps use `make_timed_token()` with a value of 6 months. That way you can check if it's legit using `token_valid()`.

Yes, the name has to be identifiable, I was thinking about the value of tn_nav :slightly_smiling_face: 
Cookies are key-value pairs. 
There is a built in function that set expiry date, it just defaults to 14 days but I'm pretty sure that can be modified. It also accept dates iirc.

JeffAbrahamson:
uaparser -- awesome!  I'm not sure what info we'll want of that eventually, but not collecting data is the best way not to have data.  From memory, I bet that's just a couple lines of code, so well worth it!

Field length of 50 is fine, of course.  It's just about being more than needed.  To much is enough...

PR #953, opened 2022-10-05
Benja report button 928

Shriukan33:
This PR adds a report session feature under session summaries.
Authenticated users may see and click the report button, which, once it effectively reported, turns into a normal text. 

The form to report is modal and you can click out of it.


transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,18 @@ def save(self, *args, **kwargs) -> None:
             super().save(*args, **kwargs)
         except ValidationError as e:
             logger.error(f"Error while saving event : {e}")
+
+
+class SessionReport(models.Model):
+    """A report on a session."""
```
JeffAbrahamson, line 117: This comment doesn't seem to add much value over the class name?

Shriukan33:
> The git comment on [05eee7c](https://github.com/transport-nantes/tn_web/commit/05eee7cef56d651d784bd2a954979033d09bbd17) is wrong: you're removing "static", not adding it back.

Meant adding back to tracked folders, that's ambiguous haha

transport_nantes/mobilito/models.py
Shriukan33: # This is a comment 
sort of vibe haha, added it just to have a docstring, but that wasn't exactly necessary.

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +443,41 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
         return self.render_to_response(context)
+
+
+@csrf_protect
+def report_session(request: HttpRequest, **kwargs) -> HttpResponse:
+    """Report a session as inappropriate"""
+
+    if request.method == 'POST':
+        if not request.user.is_authenticated:
+            return HttpResponse(status=401)
```
JeffAbrahamson, line 454: By putting the bar so high -- how could a conscientious user authenticate at this point -- we lose a lot of information (most reports) and potentially frustrate users ("I reported spam and the site errored" -- the average user won't realise it's a 4xx and will tell his friends that the site errored without explanation).

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +443,41 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+        session_sha1 = kwargs['session_sha1']
+        try:
+            session_object = get_object_or_404(Session, session_sha1=session_sha1)
+        except Http404:
```
JeffAbrahamson, line 458: This deserves an explanation why you believe this can't possibly happen in the normal course of affairs.  (I believe you're right, but it needs to be documented.)

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +443,41 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+                    'session': session_object,
+                    'reporter_user': request.user,
+                    'reporter_session_id': request.session.session_key,
+                    'textual_report': request.POST.get('report-abuse-text')
+                })
+            if created:
+                logger.info(f"{request.user} reported session "
+                            f"{session_object.id}")
+            else:
+                logger.info(f"{request.user} tried to report session "
```
JeffAbrahamson, line 477: This also deprives us of quite a lot of information.  A page that is reported multiple times is a stronger signal, not something about which we should fret.

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,18 @@ def save(self, *args, **kwargs) -> None:
             super().save(*args, **kwargs)
         except ValidationError as e:
             logger.error(f"Error while saving event : {e}")
+
+
+class SessionReport(models.Model):
```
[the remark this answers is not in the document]
JeffAbrahamson: Now that I've finished reading the code, I understand that this is a spam/abuse report about a session, so my remaining comments on this model are likely unuseful here.  A reader of the model file would likely have similar confusions.  The importance of documentation.  WTF is a report?  Only you know, as far as this file is concerned.

JeffAbrahamson (changes requested):
Also, I don't see how I can _use_ this feature.  Screenshot in converation.

JeffAbrahamson:
How would one signal this session?

image

transport_nantes/mobilito/models.py
JeffAbrahamson: Or
```
Represent user spam/abuse reports on mobilito recording sessions.

We are publishing user-provided data to the internet.
Despite our efforts, it may happen that some sessions
are inappropriate, off brand, and/or shouldn't appear
publicly.  This mechanism lets visitors flag such inappropriate sessions with an optional note to tell us
why they believe the session is inappropriate.
```

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,18 @@ def save(self, *args, **kwargs) -> None:
...
+    """A report on a session."""
+    session = models.ForeignKey(Session, on_delete=models.PROTECT)
+    reporter_session_id = models.CharField(max_length=200)
+    reporter_user = models.ForeignKey(User,
```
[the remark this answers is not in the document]
Shriukan33: user A makes a session. 
User B reports it, and is saved as reporter_user.
So we can tell who reported what.

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,18 @@ def save(self, *args, **kwargs) -> None:
...
+                                      on_delete=models.PROTECT,
+                                      null=True, blank=True)
+    timestamp = models.DateTimeField(auto_now_add=True)
```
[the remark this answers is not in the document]
Shriukan33: Auto_now_add is indeed creation timestamp.
Should I rename it creation_timestamp ? That's effectively the time of report.

JeffAbrahamson:
In private browsing session (so I'm not logged in):
image

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
Shriukan33: This is the feature you asked in #928, I'll make a more verbose description to make it clearer.

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
JeffAbrahamson: When _using_ the variable, you'd surely like your usage to self-document what the thing is.  And at that point, the reader is not looking at the model definition.

So, yes, `creation_timestamp` makes much more sense.

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
JeffAbrahamson: Fair enough.  Although, as noted below, in a private session (so not auth) I still don't see a button to report the session.

transport_nantes/mobilito/views.py
Shriukan33: Non-auth user wouldn't be able to signal the session, should I add a button that redirects to login perhaps ? Like you clic "Signaler un abus" and it redirects to login ?

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,18 @@ def save(self, *args, **kwargs) -> None:
             super().save(*args, **kwargs)
         except ValidationError as e:
             logger.error(f"Error while saving event : {e}")
+
+
+class SessionReport(models.Model):
+    """A report on a session."""
+    session = models.ForeignKey(Session, on_delete=models.PROTECT)
+    reporter_session_id = models.CharField(max_length=200)
```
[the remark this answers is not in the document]
JeffAbrahamson: Ah, so you're hoping readers will magically understand when you mean mobilito session and when you mean django session? ;-)

transport_nantes/mobilito/views.py
Shriukan33: Just in case someone uses curl to post, not very likely I agree, but at least the event is handled

JeffAbrahamson:
> > The git comment on [05eee7c](https://github.com/transport-nantes/tn_web/commit/05eee7cef56d651d784bd2a954979033d09bbd17) is wrong: you're removing "static", not adding it back.
> 
> Meant adding back to tracked folders, that's ambiguous haha

For that edge case of "ambiguous" in which the sense is "actually means the opposite".  LOL.

transport_nantes/mobilito/views.py
Shriukan33: Multiple users can report the same session. 
A single user can only report once.

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
Shriukan33: Fair enough, I was copying too much issue's fields, without thinking about out of context reviewer :/

transport_nantes/mobilito/views.py
JeffAbrahamson: No, you should just let them signal it.  Most people will never click the button no matter how egregious the content.  Of those who do, few will both to auth, provide an email address, wait for a mail, click through, and presumably remember what charitable act they were in the middle of to click the button and provide an explanation or confirmation of the spam report.

Do you have _any_ examples of where such a workflow has worked in the past with passers-by?

You'll note in the issue that non-auth users are explicitly permitted to flag pages.

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
Shriukan33: noted, will apply this change

transport_nantes/mobilito/views.py
JeffAbrahamson: CSRF protect

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
Shriukan33: You need to be auth and not the author to have the option to report, although I see that we might want anon user to see the button except it makes them login.

transport_nantes/mobilito/views.py
JeffAbrahamson: And I agree with you that it's highly unlikely and that it's reasonable to 4xx if it happens.  But you _do_ need to explain clearly your reasoning in the code.

transport_nantes/mobilito/views.py
JeffAbrahamson: Can you think of other sites that have actions you can/should only do once?  How do they communicate to you that you've done the action and can't do it again?

How do they know who you are and why do you think they make those decisions?

transport_nantes/mobilito/views.py
Shriukan33: How do we prevent spams then ? One could refresh and click report over and over

transport_nantes/mobilito/views.py
Shriukan33: Hmm right, the CSRF protect will prevent the post from originating from somewhere else.
Should I get rid of this try/except block altogether ? It might confuse readers with a very unlikely event

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
JeffAbrahamson: You mean you were copying this part?  ;-)

*    mobilito session id
*   django session id
*   django user id if available
*   timestamp of report
*   string reason for report

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
JeffAbrahamson: So, in other words, nearly _nobody_ who sees the report will be presented with the option to flag it, would even see that such an option is possible?  How does this help us?

transport_nantes/mobilito/views.py
JeffAbrahamson: From the issue,

> When a user visits a summary page, we should do a quick check if they've already reported the page. If yes, we should make sure there's (subtle) visible feedback so that they don't feel a need to report it again.

Shriukan33: I did implement that mechanism, except that to check if one has already reported, I enforced login.
If you already reported, a string saying "Session signalée" is displayed. 
If you're anon, I can't tell if you already did or not ...

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
JeffAbrahamson: Do you believe that "session report" has immediate clear meaning to the reader with only the context that this has something to do with mobilito?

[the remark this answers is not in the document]
JeffAbrahamson: I'm only saying that you need to document your work.  I know exactly what you are doing and _I_ misread it.  If that's not a signal...

transport_nantes/mobilito/views.py
JeffAbrahamson: No, you should **document** what you have done.

transport_nantes/mobilito/views.py
JeffAbrahamson: You have the django session id.

> a string saying "Session signalée" is displayed
How does stack overflow signal to you that you've already voted on something?  (They require login for reasons other than preventing you from voting twice.  What is that reason, do you suppose?  Why does that apply to them and not to us?)

Shriukan33: On SO, upvotes are modal, you can give an upvote and then take it back by clicking again.
You can't upvote or do anything really without an account and some reputation.
We don't have such system, an ill-intentioned person could decide to create multiple accounts and make reports, but that's unlikely, for now.

Django session can be very easily refreshed, all you have to do is close browser, start a private navigation, and it's done. 
Just tested it and first GET, the session ID is actually None (in get_context_data)
A first refresh and you have a session ID. 
Close browser and go incognito, same partition plays again, with None, then a brand new session ID.

I don't know if we could identify someone that never got to the website, as their django_session_id would be None in the first get. Also, incognito mode would allow the same user to report multiple times easily.

JeffAbrahamson: That seems like a bug if a visitor can have no session id.  We're supposed to set that on arrival if it's not already set.

JeffAbrahamson: Why does SO have the auth-only voting system that they have?  What does it give them?

Why is that not applicable to us?

What happens if someone signals abuse abusively?

JeffAbrahamson:
The commit comment on 47c3754 is incorrect (and has at least one spelling error).  I suspect it's also not been addressed since last week's review, but correct me if I'm wrong there. ;-)

Shriukan33:
> The commit comment on [47c3754](https://github.com/transport-nantes/tn_web/commit/47c375439ecfba7e99787b86255de7647b96248c) is incorrect (and has at least one spelling error). I suspect it's also not been addressed since last week's review, but correct me if I'm wrong there. ;-)

Mind elaborating please ? 
About the anon users able to report, I haven't found a way to both allow the anon users to report and prevent spams.
The session key can be refreshed very easily, so it can't be used for that purpose.

JeffAbrahamson:
> > The commit comment on [47c3754](https://github.com/transport-nantes/tn_web/commit/47c375439ecfba7e99787b86255de7647b96248c) is incorrect (and has at least one spelling error). I suspect it's also not been addressed since last week's review, but correct me if I'm wrong there. ;-)
> 
> Mind elaborating please ? About the anon users able to report, I haven't found a way to both allow the anon users to report and prevent spams. The session key can be refreshed very easily, so it can't be used for that purpose.

Why can't anon users flag potential spam?  If we find a spammer, we can ignore the person.

I'm still confused why you think that users sometimes have empty session ids, we don't see that elsewhere.  Cf. my question about session_id vs tn_session_id.

Shriukan33:
> > > The commit comment on [47c3754](https://github.com/transport-nantes/tn_web/commit/47c375439ecfba7e99787b86255de7647b96248c) is incorrect (and has at least one spelling error). I suspect it's also not been addressed since last week's review, but correct me if I'm wrong there. ;-)
> > 
> > 
> > Mind elaborating please ? About the anon users able to report, I haven't found a way to both allow the anon users to report and prevent spams. The session key can be refreshed very easily, so it can't be used for that purpose.
> 
> Why can't anon users flag potential spam? If we find a spammer, we can ignore the person.
> 
> I'm still confused why you think that users sometimes have empty session ids, we don't see that elsewhere. Cf. my question about session_id vs tn_session_id.

So turns out we have two sessions : 
`request.session.session_key` -> :x: This one is evaluated to None in get_context_data
`request.session.get('tn_session')` -> :o: this is set before get_context_data

However, opening a new incognito window is enough to refresh tn_session.
We'd have no way to block someone from reporting, as we couldn't identify them.

Now if you're ok with that risk, I can make it available to anon and save the tn_session in reporter_django_session_id  ?

JeffAbrahamson:
What if we're not ok with that risk?  Then we have no way to flag abuse without putting legitimate flaggers through a process so difficult that nearly no one will ever do it.

What if we allow that risk?  What bad can happen that didn't happen in the first case?

Shriukan33:
> What if we're not ok with that risk? Then we have no way to flag abuse without putting legitimate flaggers through a process so difficult that nearly no one will ever do it.
> 
> What if we allow that risk? What bad can happen that didn't happen in the first case?

From my POV : 
Worst case scenario if anon can't report a session :
- Someone does a bullsh*t session, noticeable enough by a visitor, visitor doesn't want to go through login.
- We remain with an unreported session, and the best thing we can hope is that statistical analysis will detect it as outlier (numbers of click / min too high, too focused on one specific spot, these kind of metrics.) 
- Grand total : 1 session that is moonshine, and 0 to X not frustrated enough visitors to bother reporting.

Worst case scenario if anon can report a session :
- Legit sessions and fake ones can be reported anonymously
- One person can create an infinite amount of reports in the database
- We can't trust that 1 report = 1 person ( it's easy to create a new account, but at least requires to have a new email everytime)
- We can't stop the behavior unless we disallow all anons to report.
- It takes only one to take down the point of this system. 

I can't think of a website with anonymous allowed to vote for content, you need at least _some way_ to identify the reporter.

Now maybe we could find something halfway by adding a captcha in the reporting form ? That would at least prevent basic automated selenium scripts to post massively.

Shriukan33:
Report button is now accessible to anonymous users, we also save their IP address and tn_session.

.gitignore
```diff
@@ -10,7 +10,6 @@ Pipfile*
 # Django
 db.sqlite3
 media/
-static/
```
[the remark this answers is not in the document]
JeffAbrahamson: This isn't addressed.  By doing this, the root-level static folder shows as untracked instead of ignored.  It should be ignored.  Use appropriate regex's to do what you mean and not what you said.

JeffAbrahamson (changes requested):

JeffAbrahamson:
Apparently I reviewed this and forgot to hit the final save / request changes button.  Sorry.  And I was wondering why nothing was changing...

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,26 @@ def save(self, *args, **kwargs) -> None:
...
+    session = models.ForeignKey(Session, on_delete=models.PROTECT)
+    reporter_django_session_id = models.CharField(max_length=200)
+    reporter_user = models.ForeignKey(User,
+                                      on_delete=models.PROTECT,
+                                      null=True, blank=True)
+    creation_timestamp = models.DateTimeField(auto_now_add=True)
+    textual_report = models.TextField(blank=True, null=True)
+
+    def __str__(self):
+        return (f'Report on {self.session.user.user.email} - '
```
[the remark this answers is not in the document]
Shriukan33: session objects have a user attribute, which is a MobilitoUser. 
We're not referencing to the person that is reporting, but to the reported person, which has at the minimum provided an email to login and create the session.

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,26 @@ def save(self, *args, **kwargs) -> None:
             super().save(*args, **kwargs)
         except ValidationError as e:
             logger.error(f"Error while saving event : {e}")
+
+
+class SessionReport(models.Model):
```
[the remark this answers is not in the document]
Shriukan33: Changed to InappropriateFlag

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,26 @@ def save(self, *args, **kwargs) -> None:
...
+    with an optional note to tell us why they believe the session is 
+    inappropriate.
+    """
+    session = models.ForeignKey(Session, on_delete=models.PROTECT)
+    reporter_django_session_id = models.CharField(max_length=200)
+    reporter_user = models.ForeignKey(User,
+                                      on_delete=models.PROTECT,
+                                      null=True, blank=True)
+    creation_timestamp = models.DateTimeField(auto_now_add=True)
+    textual_report = models.TextField(blank=True, null=True)
```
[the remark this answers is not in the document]
Shriukan33: Changed to report_details

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,26 @@ def save(self, *args, **kwargs) -> None:
...
+
+    We are publishing user-provided data to the internet.
+    Despite our efforts, it may happen that some sessions
+    are inappropriate, off brand, and/or shouldn't appear
+    publicly.  This mechanism lets visitors flag such inappropriate sessions
+    with an optional note to tell us why they believe the session is 
+    inappropriate.
+    """
+    session = models.ForeignKey(Session, on_delete=models.PROTECT)
+    reporter_django_session_id = models.CharField(max_length=200)
```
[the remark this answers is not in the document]
Shriukan33: I improved the documentation by adding to the class' docstring a description of each field and a note about the lingo

Also I do think it's important to have the reporter_ prefix to differentiate the session and the person reporting the session, I think it's more explicit that way.

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +443,44 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+            session_object = get_object_or_404(Session, session_sha1=session_sha1)
+        except Http404:
+            logger.error(f"{request.user.email} tried to report a "
+                         "non-existing session")
+            raise Http404
+        if session_object:
+            _, created = SessionReport.objects.get_or_create(
+                session=session_object,
+                reporter_user=request.user,
+                reporter_django_session_id=request.session.session_key,
```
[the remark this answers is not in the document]
Shriukan33: # request.session.session_key
request.session.session_key is a django-generated key to identify a SessionStore instance in our backend (we have django.contrib.sessions enabled in INSTALLED_APPS [settings.py])
https://docs.djangoproject.com/en/4.1/topics/http/sessions/#using-sessions-out-of-views

This is the object you access when we use `request.session` in views.

It is saved when modified : https://docs.djangoproject.com/en/4.1/topics/http/sessions/#when-sessions-are-saved

# request.session.get('tn_session')
This is our homemade session identifier
We implemented this a year ago through this issue : https://github.com/transport-nantes/tn_web/issues/173

It seems that, that day, we reinvented the django session by adding to _the existing session_ another id, which we named tn_session. 
It's added to the session through our middleware 
https://github.com/transport-nantes/tn_web/blob/9f4a321535399805e0cc512cf177a565a64cfcc6/transport_nantes/asso_tn/middleware/sessionCookie.py#L4

It basically serves the same purpose, except that it's already initialized when you're in the view (while django session saves only if the session is modified, and while it's not, it has no session_key).

So basically we have a session, that might or might not have a session_key (it's primary key once saved), but that session always have a "tn_session" key, thanks to our middleware. tn_session can't exist without django session, reverse isn't true. 
tn_session in the end, is an attribute of a django session.

transport_nantes/mobilito/tests.py
```diff
@@ -4,7 +4,7 @@
 from django.test import TestCase, Client
 from django.urls import reverse_lazy
 
-from .models import Session, MobilitoUser
+from .models import Session, MobilitoUser, SessionReport
 
 
 class SessionViewTests(TestCase):
```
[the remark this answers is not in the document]
Shriukan33: I did precise scenario in comments and in assert messages the expected outputs.

If there are blurry lines for you, please tell me which one and I'll rephrase so it sounds clearer.
Right now I don't see it

Shriukan33:
@JeffAbrahamson Ready for round 2

.gitignore
```diff
@@ -11,6 +11,7 @@ Pipfile*
 db.sqlite3
 media/
 static/
+!transport_nantes/*/static/
```
JeffAbrahamson, line 14: Why is this correct?

Rather than excluding what you don't want and hoping you didn't miss anything, please write a single regex that matches what you do want.

transport_nantes/mobilito/models.py
```diff
@@ -111,3 +111,50 @@ def save(self, *args, **kwargs) -> None:
...
+    """Represent user spam/abuse reports on mobilito recording sessions.
+
+    We are publishing user-provided data to the internet.
+    Despite our efforts, it may happen that some sessions
+    are inappropriate, off brand, and/or shouldn't appear
+    publicly.  This mechanism lets visitors flag such inappropriate sessions
+    with an optional note to tell us why they believe the session is
+    inappropriate.
+
+    Notes on lingo :
```
JeffAbrahamson, line 126: It's good that you comment these words.  Why don't you include the comments next to the definitions?

JeffAbrahamson: The commit message for this model talks about the SessionReport model, which apparently does not exist.  Either the commit message or the commit is therefore wrong...

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +452,50 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+
+@csrf_protect
+def report_session(request: HttpRequest, **kwargs) -> HttpResponse:
+    """Report a session as inappropriate"""
+
+    if request.method == 'POST':
+        if request.user.is_authenticated:
+            user = request.user
+        else:
+            user = None
```
JeffAbrahamson, line 465:
```
user = None
if request.user.is_authenticated:
    user = request.user
```

or, better,

```
user = request.user if request.user.is_authenticated else None
```

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +452,50 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+
+        session_sha1 = kwargs['session_sha1']
+        try:
+            # In the event that somehow, in spite of CSRF protection, a user
```
JeffAbrahamson, line 469: The goal of CSRF is not to avoid flagging non-existent sessions, it is to avoid outsiders doing it without going through our website.

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +452,50 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+            # tries to report a session that doesn't exist, we don't want to
+            # crash the app, so we catch the exception, log it and return a 404
+            session_object = get_object_or_404(
+                Session, session_sha1=session_sha1)
+        except Http404:
+            logger.error(f"{user or 'Anonymous User'} tried to report a "
```
JeffAbrahamson, line 475: This will not provide you with useful information for debugging.

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +452,50 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+                         "non-existing session")
+            raise Http404
+        if session_object:
```
JeffAbrahamson, line 478: Under what circumstances could this be False?

Under those circumstances, you've decided you can silently ignore the case.  Why?

transport_nantes/mobilito/views.py
```diff
@@ -436,3 +452,50 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+                    'reporter_user': user,
+                    'reporter_tn_session_id': request.session.get('tn_session'),
+                    'report_details': request.POST.get('report-abuse-text'),
+                    'reporter_ip_address': client_ip,
+                })
+            if created:
+                logger.info(f"{user or 'Anonymous User'} reported session "
+                            f"{session_object.id}")
+            else:
+                logger.info(f"{user or 'Anonymous User'} tried to report session "
```
JeffAbrahamson, line 495: Language: are these flags or reports?

JeffAbrahamson: If we can tell that the person has already flagged this session, then why did we permit them to do so?

Also, again, log enough that you can debug if you need to.  You are teasing your future self by not leaving useful information in the log, only a T/F.

transport_nantes/mobilito/templates/mobilito/session_summary.html
```diff
@@ -19,5 +31,61 @@ <h1>
             et {{ mobilito_session.public_transport_count|default:"0" }} transports collectifs.
         </p>
         
+        {% if request.user != mobilito_session.user.user %}
+            {% if user_has_reported_this_session %}
+                <span id="reported" class="small text-muted">
+                    Session signalée
```
JeffAbrahamson, line 37: Can we please make this more subtle?

First, these aren't necessarily abuses, and we don't want to get into discussions about the circumstances under which something should be flagged.

Second, this steals attention.  It should be visible if you need it and peripheral otherwise.

This should a light-toned icon like a flag, which changes tone if already flagged.

JeffAbrahamson (changes requested):
These five commits alternate in language: report, flag, etc.  This needs to be made uniform in order for the whole to be easily understandable.

.gitignore
Shriukan33: static/ bans every static folder (including js and css files)
!transport_nantes/*/static/ adds an exception (with `!`)  for static folders inside our apps. 
As you know, transport_nantes/ folder is the base directory, inside it we have all our apps, and most have a subdirectory named static. 

What's wrong with this, and how is a regex better than the .gitignore syntax ?

transport_nantes/mobilito/views.py
Shriukan33: I agree. I hardly see how this could happen to be honest.

transport_nantes/mobilito/models.py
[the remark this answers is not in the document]
JeffAbrahamson: sure, it's just that it's a flag in the model...

transport_nantes/mobilito/views.py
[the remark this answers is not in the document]
JeffAbrahamson: > So basically we have a session, that might or might not have a session_key (it's primary key once saved), but that session always have a "tn_session" key, thanks to our middleware. tn_session can't exist without django session, reverse isn't true.

Indeed, and that's a huge difference, and the reason we have tn_session_id.

JeffAbrahamson:
Branch has conflicts.

JeffAbrahamson (approved):
LGTM.

Tests should typically be introduced on the same commit as the code being tested.  Otherwise there's scant reason to believe the code is correct.  "Untested code is broken by design."

We have a lot of untested code, it's been a joint effort to arrive at that.  Life...

Issue #199, opened 2021-09-20
Get SES working


JeffAbrahamson:
We use Amazon SES for mail.

TODO:
* [x] @JeffAbrahamson gives @Shriukan33 AWS access to admin
* [x] @Shriukan33 gets "thank you for your donation" mails working, cf #49 when starting to send mail
* [ ] @Shriukan33 consults with GJ on text of said mail so that we all agree.  That mail should be a TBv2 item with "view on web" link.

JeffAbrahamson:
In making mail work, we'll also want to make newsletters work, which means TBv2 integration.

Shriukan33:
Added :

- [ ] Add mobilitains.fr as verified domain to send mails from in AWS. See https://us-east-2.console.aws.amazon.com/sesv2/home?region=us-east-2#/verified-identities

Shriukan33:
We don't send emails on donation, only show a template : 
https://github.com/transport-nantes/tn_web/blob/a8a8cceb4ece8d3810ac255607881da2c7302684/transport_nantes/stripe_app/views.py#L205-L210

JeffAbrahamson:
That's what I thought.  And yet this issue has been 0-urgent since 2021-11-21...

PR #957, opened 2022-10-10
Add mail sending on donation

Shriukan33:
Donors will receive an email thanking them for their donation.

Part of #199

Deployed on beta  :
image


JeffAbrahamson (approved):
LGTM.

The mail sending could be more DRY, but for another day...

Issue #921, opened 2022-09-15
Stop confusing publication constraints with save constraints


JeffAbrahamson:
There's (almost) nothing that should prevent saving a TBObject.  But we still get stopped on what are really publication constraints: you can't have this without that.

It's completely normal to want to save one's work, even to continue later on another device.

PR #958, opened 2022-10-11
Move image field validation to publication instead of saving

### LINKED ISSUE #922: TBObjects losing images
Create a TBPress.  Fill out first page.  Save.
Modify to add social media (OG) data.  Save.  Refused, because header image is missing.

But the header image was provided in the first version.

Providing the header image causes the OG image to disappear.

Shriukan33:
This commit allow the TBObjects to keep the images on edit.

One should be able to save an unfinished work, but not publish it. The clean() method used to enforce the header image validation leading to a form_invalid() call when saving an unfinished work. This would prevent the images from persisting, as we retrieve the images in the form_valid() method.

Closes #922 #921

### DIFF
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -439,15 +439,6 @@ class TopicBlogObjectSocialBase(TopicBlogObjectBase):
         verbose_name="Notes libres pour éditeurs",
         blank=True, null=True)
 
-    def clean(self):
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
`ModuleNotFoundError: No module named 'compressor'`

JeffAbrahamson (approved):
LGTM.

PR #962, opened 2022-10-18
Secure MAPS API key and transfer the API call to server

### LINKED ISSUE #954: Properly secure maps API key
The google maps API call currently happens in `mobilito/js/geolocation_map.js`.  The `onMapClick(e)` handler is called when the user clicks the leaflet map, which generates an event that includes longitude and latitude.  We then, in javascript, call `getGeocodedAddress()` which calls a google maps geocoding endpoint and returns the result.

This code has two bugs.

1/ First, the function `getGeocodedAddress()` blindly returns whatever the google maps geocoding endpoint returns.  This is inappropriate management of that abstraction.  Clients of this function shouldn't have to know anything about its implementation.  If we someday switch to a new service, such as OSM, callers should be none the wiser.  The function should state its input and output contracts and stick to them.

In passing, the comment about the 50 rps limit is about the google call and not about this function per se, so that documentation should be inside the function.  Moreover, it should be clearer: 50 rps from this client or globally using this key?  The function should have a comment that explains what the _function_ does.

This also means we can't IP-restrict the key, since it will be used from the user's IP and not from ours.  An IP restriction to our hosts would guarantee google request failure.

2/ Second, the google maps geocoding endpoint only needs longitude and latitude to function, so it doesn't need to be called from client code.  As a result, the client code doesn't need to know the API key value, and so we don't need to leak the key to the world.  We should establish a CSRF-protected ajax endpoint with a clear API and contract, and `getGeocodedAdress()` should simply call that.  That endpoint will make its own ajax call to google, wait on its future, and return when it becomes valid.

Shriukan33:
To avoid disclosing our API key and have a better control over the API calls, we now transfer the API call to our server. Clients make requests to our server, and if CSRF token is valid, the server makes the API call and returns the result to the client.

Closes #954

### DIFF
--- a/transport_nantes/mobilito/static/mobilito/js/geolocation_map.js
+++ b/transport_nantes/mobilito/static/mobilito/js/geolocation_map.js
@@ -112,15 +112,17 @@ map.on('click', onMapClick);
 
 // Function to get user's location with geocoding, limited 50 requests per second.
 async function getGeocodedAddress(lat, lng) {
-    let api_key = JSON.parse(document.getElementById('maps_api_key').textContent);
-    let url = `https://maps.googleapis.com/maps/api/geocode/json?latlng=${lat},${lng}&key=${api_key}`;
+    let url = "/mobilito/ajax/get-address/";
     let response = await $.ajax({
         url: url,
-        type: 'GET',
+        headers: {"X-CSRFToken": getCookie("csrftoken")},
+        type: 'POST',
         dataType: 'json',
-        success: function (data) {
-            console.log(data);
+        data: {
+            lat: lat,
+            lng: lng,
         },
+        mode: 'same-origin',
         error: function (err) {
             console.log(err);
         }
--- a/transport_nantes/mobilito/urls.py
+++ b/transport_nantes/mobilito/urls.py
@@ -16,4 +16,6 @@ urlpatterns = [
          name='session_summary'),
 
     path('ajax/create-event/', views.create_event, name='event_creation'),
+    path('ajax/get-address/', views.ReverseGeocodingView.as_view(),
+         name='geocoding'),
 ]
--- a/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
+++ b/transport_nantes/asso_tn/templates/asso_tn/base_mobilitain.html
@@ -95,6 +95,25 @@
 					<noscript><p><img src="//wa.mobilitains.fr/matomo.php?idsite=3&amp;rec=1" style="border:0;" alt="" /></p></noscript>
 					<!-- End Matomo Code -->
 					{% endif %}
+					<script type="text/javascript">
+						// From django docs
+						// https://docs.djangoproject.com/en/dev/howto/csrf/#s-acquiring-csrf-token-from-cookie
+						function getCookie(name) {
+							let cookieValue = null;
+							if (document.cookie && document.cookie !== '') {
+								const cookies = document.cookie.split(';');
+								for (let i = 0; i < cookies.length; i++) {
+									const cookie = cookies[i].trim();
+									// Does this cookie string begin with the name we want?
+									if (cookie.substring(0, name.length + 1) === (name + '=')) {
+										cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
+										break;
+									}
+								}
+							}
+							return cookieValue;
+						}
+					</script>
 				{% endcompress %}
 
 
--- a/transport_nantes/mobilito/static/mobilito/js/plus_minus_buttons.js
+++ b/transport_nantes/mobilito/static/mobilito/js/plus_minus_buttons.js
@@ -1,19 +1,3 @@
-function getCookie(name) {
-    let cookieValue = null;
-    if (document.cookie && document.cookie !== '') {
-        const cookies = document.cookie.split(';');
-        for (let i = 0; i < cookies.length; i++) {
-            const cookie = cookies[i].trim();
-            // Does this cookie string begin with the name we want?
-            if (cookie.substring(0, name.length + 1) === (name + '=')) {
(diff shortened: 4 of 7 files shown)


transport_nantes/mobilito/urls.py
```diff
@@ -16,4 +16,6 @@
          name='session_summary'),
 
     path('ajax/create-event/', views.create_event, name='event_creation'),
+    path('ajax/get-geocoded-address/', views.MapsGeocodingView.as_view(),
```
JeffAbrahamson, line 19: Kind of a funny function name: it's getting an address, not a specially geo-encoded address.

JeffAbrahamson (approved):
LGTM.

transport_nantes/mobilito/static/mobilito/js/geolocation_map.js
```diff
@@ -110,17 +110,36 @@ function onMapClick(e) {
 }
 map.on('click', onMapClick);
 
+// From django docs
```
JeffAbrahamson, line 113: This is the third instance of this function to be introduced in our code base.  It should not be: we should realise from the second instance that it is useful (really, the library doesn't have a get cookie function?) and put it somewhere common.

JeffAbrahamson, line 113: It's good that you site that it comes from django docs, but you should include a URL.

JeffAbrahamson: Finally, how is this function different from `request.COOKIES.get()`?  This requires documentation.

transport_nantes/mobilito/urls.py
```diff
@@ -16,4 +16,6 @@
          name='session_summary'),
 
     path('ajax/create-event/', views.create_event, name='event_creation'),
+    path('ajax/get-address/', views.MapsGeocodingView.as_view(),
+         name='geocoding'),
```
JeffAbrahamson, line 20: Why is the name "geocoding" mnemonic for a function that fetches the user's physical address from GPS data?

JeffAbrahamson (changes requested):
Please grammar check the commit message (cf. google).

Is the commit message at the right level of detail?  It spends 1/3 of its length explaining that there's a CSRF-protected exchange.  Is that what is pertinent in reading commit comments to understand how program structure and logic has evolved?

transport_nantes/mobilito/static/mobilito/js/geolocation_map.js
Shriukan33: > It's good that you site that it comes from django docs, but you should include a URL.

I was unsure about it since that answer : 
https://github.com/transport-nantes/tn_web/pull/938#issuecomment-1258399627
> I don't think you need to provide doc links [...]  It's good that you've learned this, but you can generally assume that competent developers know how to look up docs.

> Finally, how is this function different from request.COOKIES.get()? This requires documentation.
request.COOKIES.get is a django function on HttpRequest objects, while this function is usable in Javascript.
It parses document.cookies, which returns a string, to get the cookie that is relevant to me.

transport_nantes/mobilito/urls.py
Shriukan33: going from lat,lng to address is actually reverse geocoding, I'll think of another name. 
What about `MapsReverseGeocodingView` ? Or `UserLocationView` ? I'm open to suggestions here

Shriukan33: Went with `ReverseGeocodingView`

transport_nantes/mobilito/static/mobilito/js/geolocation_map.js
JeffAbrahamson: How is it different from `request.COOKIE`?

JeffAbrahamson: When you copy code from somewhere, you should absolutely credit from whence you took it.  That is both politeness and honesty.  Explaining in detail what it does would be a different story, since it's obvious in this case.

transport_nantes/mobilito/urls.py
JeffAbrahamson: Is that clear than "address_lookup_from_geo"?

Shriukan33: > Is that clear than "address_lookup_from_geo"?

PEP8 tells us we should use CamelCase for class names.
AddressLookupFromGeoView isn't, to my mind, clearer than ReverseGeocodingView. 

Reverse Geocoding describes the process to go from lat,lng pair to a fully worded address. This is what this view does, I even omitted the Maps part because it's irrelevant, no matter the service we use, inputs and outputs are the same nature.

Address_lookup_from_geo otoh, is a bit confusing to me : address lookup ? What kind of address ? IP ? and geo ? What's geo ? Has to do with Geography but other than that ..

Shriukan33:
Ready for review again

JeffAbrahamson (approved):
LGTM.

Don't we have a base js library?

Shriukan33:
> LGTM.
> 
> Don't we have a base js library?

what do you mean by base js library ?

JeffAbrahamson:
> > LGTM.
> > Don't we have a base js library?
> 
> what do you mean by base js library ?

A js library that we include everywhere, that contains stuff like this, the basic utilities we expect to need more or less everwhere.  Seems like it would be practical.

Shriukan33:
> > > LGTM.
> > > Don't we have a base js library?
> > 
> > 
> > what do you mean by base js library ?
> 
> A js library that we include everywhere, that contains stuff like this, the basic utilities we expect to need more or less everwhere. Seems like it would be practical.

I used the base_mobilitain.html template for this, that's the closest thing we have.

PR #963, opened 2022-10-25
Benja photo contest improvement 960

Shriukan33:
This PR adds a voting system on the photo contest (aka operation pieton)

Any user that want to vote, logged in or not, is offered once to receive updates about the operation pieton by email.

This branch doesn't offer a List view for submissions, and is about the voting system's design itself.

I haven't included tests yet, to spare me the work of redoing them if you have any feedback and changes to make.
Once the design approved, I'll update this PR with tests for a final approval to merge. 

For the operation pieton voting system to be public, we need to add OG data and a list view.


transport_nantes/photo/models.py
```diff
@@ -63,6 +69,81 @@ class CategoryType(models.TextChoices):
         verbose_name="Date de soumission",
         auto_now_add=True,
     )
+    # Photo entries are pending (null) until they are accepted (True)
+    # or rejected (False)
```
JeffAbrahamson, line 73: A useful comment would not explain what a boolean is (;-)) so much as explain the context, that this boolean expresses the outcome of a judging process, for example.

transport_nantes/photo/models.py
```diff
@@ -63,6 +69,81 @@ class CategoryType(models.TextChoices):
...
+
+
+class Vote(models.Model):
+    """
+    Represents a vote for a photo entry.
+
+    The same user can vote several times on the same entry, but only the last
+    vote is taken into account.
+
+    Field descriptions:
```
JeffAbrahamson, line 123: Describe fields with the fields, not in summary above.  This is like those charts where there's a key on the side and you have to keep going back and forth between the key and the data in order to understand, rather than having the explanation in the data and so being able to focus on one thing only.  It's additional and unuseful congnitive load.

transport_nantes/photo/models.py
```diff
@@ -63,6 +69,81 @@ class CategoryType(models.TextChoices):
...
+    - user: the user who voted, can be none if user is anonymous
+    - tn_session_id : the tn session we set with SessionMiddleware. It's a
+    fallback mechanism in case the user is anonymous.
```
JeffAbrahamson, line 126: I'm not convinced that tn_session_id is a fallback mechanism.  It's really a primary way of tracking visitor data without needing authenticated users and so respecting privacy.  We never allow ourselves to make persistent conclusions about users based on their non-logged-in behaviour (even if we do sometimes need to map the two together, ephemerally, as here).

transport_nantes/photo/forms.py
```diff
@@ -30,3 +31,59 @@ class Meta:
...
             'submitted_photo': forms.FileInput,
         }
+
+
+class AnonymousVoteForm(forms.Form):
+    """
+    Form to collect anonymous votes the first time they vote.
+    """
+    CHOICES = [("upvote", 'Pour'), ("downvote", 'Contre')]
+    vote_nature = forms.ChoiceField(
```
JeffAbrahamson, line 41: This means something very different in English than in French.  It suggests a choice for greenery, for the natural order.  I'm afraid we need to rename this in order to avoid a persistent miscommunication on what is intended.  Just calling it "vote" would work.

transport_nantes/photo/static/photo/js/vote_system.js
```diff
@@ -0,0 +1,120 @@
+function getCookie(name) {
```
JeffAbrahamson, line 1: This would be instance four in our code base of exactly the same function?

transport_nantes/photo/static/photo/js/vote_system.js
```diff
@@ -0,0 +1,120 @@
...
+            applySelectedVoteStyles(element);
+        },
+        error: (data) => {
+            console.log("Error");
+            console.log(data);
+        }
+    });
+}
+/**
+ * This function is called on the first vote for a user (anon or not)
```
JeffAbrahamson, line 43: Why do we need a different function for the first vote than for later votes?  This deserves explanation.

transport_nantes/photo/static/photo/js/vote_system.js
```diff
@@ -0,0 +1,120 @@
...
+        type: 'post',
+        data: data,
+        mode: 'same-origin',
+        success: () => {
+            console.log("Success");
+            applySelectedVoteStyles(element);
+            $('#anon-vote-div').modal('hide');
+        },
+        error: (data) => {
+            console.log("Error");
```
JeffAbrahamson, line 60: How would you debug something that leaves such a trace?

transport_nantes/photo/static/photo/js/vote_system.js
```diff
@@ -0,0 +1,120 @@
...
+    $.ajax({
+        url: window.location.href,
+        headers: {"X-CSRFToken": getCookie("csrftoken")},
+        type: 'post',
+        data: {
+            vote_nature: vote_nature,
+        },
+        mode: 'same-origin',
+        success: () => {
+            console.log("Success");
```
JeffAbrahamson, line 33: How will we make use of this?  How does this suffice to debug anything?

JeffAbrahamson (approved):
A few comments about comments needing improvement.

Most things need tests, in the same commit as the function is introduced (otherwise one has to assume that the commit will need changing with the introduction of tests).

JeffAbrahamson:
I hit the wrong button, that was supposed to be "changes requested". ;-)

transport_nantes/photo/models.py
Shriukan33: I added this comment because we usually expect Bools to be either false or true, and this one can be null. 
Also "accepted" sounds to me like we judged it, but I'll word it anyway, comments are cheap hehe

transport_nantes/photo/models.py
Shriukan33: How would you describe its use then ? I used it as a way to identify a user that isn't logged in, it's kind of a default in absence of user account.

transport_nantes/photo/forms.py
Shriukan33: I don't want to name it vote because it would name conflict with the Vote object, hence the "nature" I appended to make a difference. 
It's important that the name doesn't make you question whether we're talking about the instance or its boolean

transport_nantes/photo/static/photo/js/vote_system.js
Shriukan33: 066960a18bfe86819aa60577be33c2de17f1cf5b

This was moved in base template in this commit, are you sure you are looking at the correct head ? (6ccba43c7b90726c75f11feb36294877bf1e99ed)

transport_nantes/photo/static/photo/js/vote_system.js
Shriukan33: It was dev sanity check haha, gotta remove it (as well as the "error" one)

transport_nantes/photo/static/photo/js/vote_system.js
Shriukan33: That's a leftover apparently, I moved into the previous commit this function to our base template

transport_nantes/photo/static/photo/js/vote_system.js
Shriukan33: We don't need the same information the first time and the next times. 
The first time we have vote_nature and at least consent box, and if not authenticated, also email and captcha.  These are retrieved with a modal form that appears when you click on any vote button when you never voted. 

The second form, once you voted a first time, doesn't require a modal and doesn't send the same info, just the vote_nature. Doing so you can vote without filling a modal and it's quicker.

transport_nantes/photo/models.py
JeffAbrahamson: It's usually better to err on the side of too much commentary than too little.  Though even better to hit the right Goldilocks ratio. ;-)

transport_nantes/photo/models.py
JeffAbrahamson: Yes, but it's not a fall-back is all.  It's a first class identifier of site visitors.

transport_nantes/photo/forms.py
JeffAbrahamson: vote_value ?

Shriukan33: Deal

Shriukan33:
I've applied suggested changes (Rename of vote_nature, documentation, moving getCookie to main template, got rid of useless console.log statements, comments closer to fields..)

Shriukan33:
Tests are absent for now, I don't intend to merge the branch without it, this is purely to check with you if this all behaves, at least at surface level, like you intended it

JeffAbrahamson:
> I've applied suggested changes (Rename of vote_nature, documentation, moving getCookie to main template, got rid of useless console.log statements, comments closer to fields..)

Console.log statements don't have to be useless, these just happened to be. ;-)  Removing was one option, the other was to make them useful.

JeffAbrahamson:
image

I don't get the UI.
* You don't need words, just left and right arrows.
* The vote buttons are empty, how would one know what to do?

JeffAbrahamson:
And, whether I vote left or right, I'm asked if I'm voting for or against.  That's excessive.  It would suffice to have a thumb up and a thumb down, and now there's nothing left to ask.

JeffAbrahamson:
image

Modulo spacing (should be uniform), I'd go with something lighter.  Here, the two spans are removed and `s/fa-solid/fa/; s/thumbs/thumbs-o/;`

JeffAbrahamson:
<i class="fa fa-thumbs-o-down my-auto" aria-hidden="true"></i>

JeffAbrahamson:
image

This form is heavy.
* The vote intention is obvious from the thumb.
* That it is a vote is also obvious from the thumb.
* Why we are asking for email and captcha is totally opaque.  We could motivate that, but we didn't.
* The opt-out should be much smaller.  The golden rule of interface design is understatedness.  Put some huge 100 pt signs around your flat explaining every day objects and see how it feels after 30 seconds.  "I'm a doorknob, you can turn me to operate the door!"  "This is a door, you can use it to access the next room."  Etc.

Shriukan33:
> image
> 
> I don't get the UI.
> 
>     * You don't need words, just left and right arrows.
> 
>     * The vote buttons are empty, how would one know what to do?

How weird ! 
image
Responsive : 
image

In responsive I simply used arrows to save on space, while on larger screens I added "Photo suivante/précédente" because it looked so small

Shriukan33:
> This form is heavy.
> 
>     * The vote intention is obvious from the thumb.
> 
>     * That it is a vote is also obvious from the thumb.
> 
>     * Why we are asking for email and captcha is totally opaque.  We could motivate that, but we didn't.
> 
>     * The opt-out should be much smaller.  The golden rule of interface design is understatedness.  Put some huge 100 pt signs around your flat explaining every day objects and see how it feels after 30 seconds.  "I'm a doorknob, you can turn me to operate the door!"  "This is a door, you can use it to access the next room."  Etc.

So your suggestions : 
- Removing header and radio select
- Say something along the lines of "Captcha pour vérifier que vous êtes bien un humain" in the field name ?
- Reduce font size on the consent checkbox 


Correct ?

Shriukan33:
Also, you would like all buttons to be evenly spaced, including vote buttons ? I did a group for buttons, but that can change

JeffAbrahamson:
> > image
> > I don't get the UI.
> > ```
> > * You don't need words, just left and right arrows.
> > 
> > * The vote buttons are empty, how would one know what to do?
> > ```
> 
> How weird ! image Responsive : image
> 
> In responsive I simply used arrows to save on space, while on larger screens I added "Photo suivante/précédente" because it looked so small

The thumbs eventually showed up.  Maybe was a caching thing.

Put the four buttons, with appropriate negative space, near the centre.  Always.  Never with text.

JeffAbrahamson:
> > And, whether I vote left or right, I'm asked if I'm voting for or against. That's excessive. It would suffice to have a thumb up and a thumb down, and now there's nothing left to ask.
> 
> That's because this form is loaded blank at page load, and changing it would require some JS mumbo jumbo to change it to whatever we clicked. I can do that, it just makes the code a bit more complicated and I went for the easier solution, especially since this form is only showed once

Why does the form have to say anything at all about how I'm voting?  I'm obviously voting, because I clicked a thumb.  Now you want to verify that I'm human, ok, so say that: "Merci de votre avis, qui nous sera encore plus precieux si nous nous surs que vous etes humain."  The email address is an extra benefit at this point.  We should just record

* opinion expressed
* captcha succeed/fail
* tn_session_id
* email if provided

Now we need to decide what to send the person if they provide an email, of course.

JeffAbrahamson:
> Also, you would like all buttons to be evenly spaced, including vote buttons ? I did a group for buttons, but that can change

Yeah, just four buttons with negative space to be uncrowded.

JeffAbrahamson:
> > This form is heavy.
> > ```
> > * The vote intention is obvious from the thumb.
> > 
> > * That it is a vote is also obvious from the thumb.
> > 
> > * Why we are asking for email and captcha is totally opaque.  We could motivate that, but we didn't.
> > 
> > * The opt-out should be much smaller.  The golden rule of interface design is understatedness.  Put some huge 100 pt signs around your flat explaining every day objects and see how it feels after 30 seconds.  "I'm a doorknob, you can turn me to operate the door!"  "This is a door, you can use it to access the next room."  Etc.
> > ```
> 
> So your suggestions :
> 
>     * Removing header and radio select
> 
>     * Say something along the lines of "Captcha pour vérifier que vous êtes bien un humain" in the field name ?
> 
>     * Reduce font size on the consent checkbox
> 
> 
> Correct ?

Correct.

And make clear that the email is optional, even if we'd really like it.

Shriukan33:
> > > And, whether I vote left or right, I'm asked if I'm voting for or against. That's excessive. It would suffice to have a thumb up and a thumb down, and now there's nothing left to ask.
> > 
> > 
> > That's because this form is loaded blank at page load, and changing it would require some JS mumbo jumbo to change it to whatever we clicked. I can do that, it just makes the code a bit more complicated and I went for the easier solution, especially since this form is only showed once
> 
> Why does the form have to say anything at all about how I'm voting? I'm obviously voting, because I clicked a thumb. Now you want to verify that I'm human, ok, so say that: "Merci de votre avis, qui nous sera encore plus precieux si nous nous surs que vous etes humain." The email address is an extra benefit at this point. We should just record
> 
>     * opinion expressed
> 
>     * captcha succeed/fail
> 
>     * tn_session_id
> 
>     * email if provided
> 
> 
> Now we need to decide what to send the person if they provide an email, of course.

So you want to register invalid forms ?

Captcha is only for anonymous users' first vote, adding a boolean captcha fail field and even accepting to save invalid form completely remove the need of a captcha at all. 
If we're going to save it no matter what, we should only put the non mandatory email field and the consent checkbox for anonymous users. 
Unless you want to identify those who failed the captcha and do something with it ?

JeffAbrahamson:
> > > > And, whether I vote left or right, I'm asked if I'm voting for or against. That's excessive. It would suffice to have a thumb up and a thumb down, and now there's nothing left to ask.
> > > 
> > > 
> > > That's because this form is loaded blank at page load, and changing it would require some JS mumbo jumbo to change it to whatever we clicked. I can do that, it just makes the code a bit more complicated and I went for the easier solution, especially since this form is only showed once
> > 
> > 
> > Why does the form have to say anything at all about how I'm voting? I'm obviously voting, because I clicked a thumb. Now you want to verify that I'm human, ok, so say that: "Merci de votre avis, qui nous sera encore plus precieux si nous nous surs que vous etes humain." The email address is an extra benefit at this point. We should just record
> > ```
> > * opinion expressed
> > 
> > * captcha succeed/fail
> > 
> > * tn_session_id
> > 
> > * email if provided
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now we need to decide what to send the person if they provide an email, of course.
> 
> So you want to register invalid forms ?
> 
> Captcha is only for anonymous users' first vote, adding a boolean captcha fail field and even accepting to save invalid form completely remove the need of a captcha at all. If we're going to save it no matter what, we should only put the non mandatory email field and the consent checkbox for anonymous users. Unless you want to identify those who failed the captcha and do something with it ?

1.  You're confusing data we capture and data we count.  We can be very sure we don't get bad votes, and so no one votes.  Or we can be relatively accepting and sort it out later as necessary.  This is our classic story: we have _no_ data and we're worried that we might collet the wrong data?

2.  We have to make a case for why we want email addresses, we can't just ask and assume that people will provide them.  We don't have an account and an ecosystem that people are keen to join, we're just some NGO asking for your email.  Yes, we want your email.  But we can't  just throw a form under people's noses and expect them to salivate and say "yes, that's what I was hoping for".

JeffAbrahamson:
> Understood.
> 
> Do we keep the captcha ? If so, should I add a boolean `captcha_success` with True if the user goes through form_valid and False if they go through form_invalid ?

Yes, I think we want to record captcha success as part of the vote: user succeeded captcha or didn't have to do one (for whatever reason, like is logged in).  If the user fails the captcha, perhaps we don't need to record the vote, chalk it up as a robot.

How we actually _count_ votes is a subject for another day.

Shriukan33:
So if logged in or successful captcha -> Bool = True
If failed captcha 
>  perhaps we don't need to record the vote, chalk it up as a robot.
But we still create a Vote in database, right ?

JeffAbrahamson:
> So if logged in or successful captcha -> Bool = True If failed captcha
> 
> > perhaps we don't need to record the vote, chalk it up as a robot.
> 
> But we still create a Vote in database, right ?

I think we could argue either way.  In general, we want to gather contacts of people who might be open to having conversations with us, supporting us.  Here, however, the goal is at least as much to encourage people to browse and to share.  If we're too aggressive on collecting email, we could lose the share, too.

Now, if the person fails the captcha, there's an argument.  We can certainly record the vote (and, probably, not count it...).  We'll want to keep an eye on the table to make sure it's not being robo-filled.

Shriukan33:
I've added the changes you asked : 
* Anon users are presented a smaller form : No header, no radio button, a small text to tell why we would like the email, smaller checkbox
* Failed captcha are flagged with the `captcha_succeeded` boolean

Rememeber to re-do your migrations, as I've modified 0010 to add the captcha field.

JeffAbrahamson:
* There's an erroneous leading space in the commit comment of 0126e9d27 ("Populate PhotoEntry ...")
* Whenever you use filter functions with first/last/etc., you _must_ specify an order.
* As you know, there are still no tests. ;-)

JeffAbrahamson:
image

Here, too, I'm confused.

* I'm not logged in, so I'm being to prove I'm human.  OK, but that's the captcha.  But what you are asking for, immediately, is my email.  What should a user understand that we want?
* Why is there a close box in the upper right?  This should be as light weight as possible, like the abuse flag.
* Why is there a "vote" button?  Didn't I vote by clicking the thumb?
* "si nous nous sûrs" ? ;-)

Seeing it rendered this way, I think we've been approaching it wrong, the text and flow need to be lighter:

* Merci !<br>Votre avis sur la photo nous sera encore plus précieux si nous pourrions être confiants que vous êtes humain.  Ca ne vous embête pas trop ?
* [captcha]
* (smaller) Avec votre accord, nous pourrions vous tenir au courant du prix populaire et des avancées de la marche à pied.
* [email] [ok]

We'll figure out the right text as we go...

Shriukan33:
Change log : 
  * Removed whitespace before commit message
  * Added order_by('timestamp') to last() queries
  * Modal form :
      * Changed header text
      * Moved captcha to first position in the form
      * Captcha is now not required (free to press OK without filling it).
      * Added small text after captcha and before email
      * Changed the button text from "Voter" to "OK"
      * Removed the consent box in anonymous forms.
        Filling the mail address field is considered as consent.

> Why is there a "vote" button? Didn't I vote by clicking the thumb?

The vote is only cast once you click OK on the modal form, otherwise we couldn't get the consent or captcha before posting.

Shriukan33:
Rebased to master

transport_nantes/photo/views.py
```diff
@@ -85,3 +91,257 @@ def get_context_data(self, **kwargs):
...
+
+class PhotoView(FormView):
+    """
+    View to display a single photo and up/down vote it
+    """
+    template_name = 'photo/single_entry.html'
+    form_class = AnonymousVoteForm
+
+    def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
+        """Overriding here so we can return HttpResponseForbidden if needed"""
```
JeffAbrahamson, line 104: This is assuredly the wrong way to do this.  If there's no way to throw a 403 from `get_context_data()` (where you determine that this is the right action) and have django do the right thing, then define a 403 exception yourself and catch it here.  Don't override return types, it's twisted and makes your code harder to understand.

transport_nantes/photo/views.py
```diff
@@ -85,3 +91,257 @@ def get_context_data(self, **kwargs):
...
+
+    def get_initial(self):
+        """Set the initial value in the form"""
+        initial = super().get_initial()
+        initial["photoentry_sha1_name"] = self.kwargs.get("photo_sha1")
+        return initial
+
+    def post(self, request, *args, **kwargs):
+        """
+        Handle POST requests: instantiate a form instance with the passed
```
JeffAbrahamson, line 182: Of course the `post()` function handles POST requests.  Perhaps you could say something more useful here in this important first line, which provides a summary of what the function does?

transport_nantes/photo/views.py
```diff
@@ -85,3 +91,257 @@ def get_context_data(self, **kwargs):
...
+                subscribe_user_to_list(user, mailing_list)
+                logger.info(
+                    f"Subscribed user {user.email} to list {mailing_list}.")
+            except MailingList.DoesNotExist:
+                logger.error(
+                    "Mailing list operation-pieton does not exist")
+
+        # Vote nature is a constant among the forms
+        vote_value = self.request.POST.get("vote_value", None)
+        if vote_value == "upvote":
```
JeffAbrahamson, line 262: Why doesn't the form set the correct True/False for you from the outset?  Not a huge deal, just wondering...

More pertinent, the pythonic way do this would be to have a dict that does the mapping, you do an assignment through the dict and on key error you return the appropriate response.  That response is _not_ 403, which means you understood the response but refuse to authorise it.

transport_nantes/photo/views.py
```diff
@@ -85,3 +91,257 @@ def get_context_data(self, **kwargs):
...
+                mailing_list = MailingList.objects.get(
+                    mailing_list_token="operation-pieton")
+                subscribe_user_to_list(user, mailing_list)
+                logger.info(
+                    f"Subscribed user {user.email} to list {mailing_list}.")
+            except MailingList.DoesNotExist:
+                logger.error(
+                    "Mailing list operation-pieton does not exist")
+
+        # Vote nature is a constant among the forms
```
JeffAbrahamson, line 325: Please use consistent language.

JeffAbrahamson (changes requested):
Several of the commits in this chain need tests.

transport_nantes/photo/views.py
Shriukan33: Missed that one. Thanks !

transport_nantes/photo/views.py
Shriukan33: Javascript sends the string "true" for `True` in it POST data, it was easier and more idiomatic for me to write out "upvote".

transport_nantes/photo/views.py
Shriukan33: Fixed that with a custom error

transport_nantes/photo/views.py
Shriukan33: I reworded the docstring to include this information in a clearer way

Shriukan33:
I'm aware tests are missing, I'm going to write them now that we're closer to the expected design :)

JeffAbrahamson:
Why is displaying the word "None" the right thing to do?

image

JeffAbrahamson:
image

I find this form confusing.  You say my vote is accepted and you'd like my permission to keep in touch.  Great!  But you start by asking for captcha and not for email.  And then you tell me you don't even need my email, which gives lie to the captcha being about keeping in touch.

JeffAbrahamson:
Not using the review tool, because no tests, so not ready for formal review...

JeffAbrahamson:
Also, my vote wasn't recorded.

```
sqlite> SELECT * FROM photo_vote;
sqlite>
```

Shriukan33:
> Also, my vote wasn't recorded.
> 
> ```
> sqlite> SELECT * FROM photo_vote;
> sqlite>
> ```

Indeed there is a miswording here, the vote is only cast one you click OK

Shriukan33:
Ok so I figured out what was happening on your screenshot. The photo you were trying to vote on was not accepted so it has this weird spot where is shouldn't have been votable anyway (not accepted photos are accessible by authorized users and the author themselves.)

Now you should see an alert "Cette photo n'est pas elligible au vote pour le moment" when the photo isn't accepted, and when it's accepted, the proper forms should be displayed. 
I tweaked the form's text so it's less confusing

Last, I added tests for the whole app, not just this last view.

JeffAbrahamson:
> Ok so I figured out what was happening on your screenshot. The photo you were trying to vote on was not accepted so it has this weird spot where is shouldn't have been votable anyway (not accepted photos are accessible by authorized users and the author themselves.)
> 
> Now you should see an alert "Cette photo n'est pas elligible au vote pour le moment" when the photo isn't accepted, and when it's accepted, the proper forms should be displayed. I tweaked the form's text so it's less confusing

Cool, that makes sense.

Why is this error message better than a simple 404 (it isn't found).  In other words, just making "accepted" part of the search criteria seems to sort the issue quite cleanly.


> Last, I added tests for the whole app, not just this last view.

Yeah!

Shriukan33:
> Why is this error message better than a simple 404 (it isn't found). In other words, just making "accepted" part of the search criteria seems to sort the issue quite cleanly.

Sending back a 404 would simply result in nothing happening in the front-end (buttons wouldn't change style), I thought that the feedback would be better if there was a clear message appearing. 

This should be a very rare occasion where someone would get to their submission and try to upvote it while we didn't accept it, I don't think it's worth sweating over it.

JeffAbrahamson:
If this is the same view that an auth'd person would see (the submitter of the photo), then just remove the vote buttons if it's not votable.  Why create an error when none exists (if the image is viewable).

If this the only purpose being here is to vote and you can't do that because the photo isn't accepted, then just error.

There's no reason to start a dialog.  The web has well-established procedures for what happens when you go to view something that isn't accessible.

Shriukan33:
Rebased on master + removed vote buttons for not accepted photos

Shriukan33:
@JeffAbrahamson

JeffAbrahamson:
There's something vaguely unbelievable about having a commit that adds tests for eight previous commits, and amazingly none of the eight previous commits requires modifications in light of the tests.

An experience dev says "when was the last time I wrote code that had no errors found in unit tests even after I'd tested it by hand?"  And then one suspects that the tests are light.

One thing I noticed, matching tests to commits, is that some commits never got tested.  The util functions, for example.  And this would be obvious if the tests accompanied the commits.

Shriukan33:
> and amazingly none of the eight previous commits requires modifications in light of the tests.

This is because I edit the history to make it look no changes are required. :slightly_smiling_face:

JeffAbrahamson:
The forward/backward buttons inactivate to a colour that appears not to be in our palette.  Do we have css defined for inactive buttons?  Cf. page 17 and also page 16 of _guide visuel mobilitains_.

JeffAbrahamson:
> > and amazingly none of the eight previous commits requires modifications in light of the tests.
> 
> This is because I edit the history to make it look like no changes are required. slightly_smiling_face

Aha!  Even better would be to edit this history so that the tests appear with the code.  This makes the reviewer's job easier and avoids missing tests.

JeffAbrahamson:
I'm seeing two strange effects on voting.
* After the very first vote, the opposite vote button is unreactive.  This auto-fixes, apparently, when changing to the next image, even after changing back.
* An anonymous voter can get no dialogue and no vote feedback.  In the logs, we see a 403 on the POST.

JeffAbrahamson (changes requested):
Please test untested code.

Please investigate why votes sometimes 403 (with nothing overly evident for the user) and why no tests have caught this.

Shriukan33:
> The forward/backward buttons inactivate to a colour that appears not to be in our palette. Do we have css defined for inactive buttons? Cf. page 17 and also page 16 of _guide visuel mobilitains_.

The color is in the palette, it's just a reduced opacity. 
I could use the mobilitains-gris color, but it looks awful 
image

.disabled class makes the opacity .65, which is an common pattern, I think. 
What color would you suggest instead ? The visual guide doesn't mention _disabled_ buttons.

Shriukan33:
> I'm seeing two strange effects on voting.
> 
>     * After the very first vote, the opposite vote button is unreactive.  This auto-fixes, apparently, when changing to the next image, even after changing back.
> 
>     * An anonymous voter can get no dialogue and no vote feedback.  In the logs, we see a 403 on the POST.

I've tried to reproduce and couldn't get the unresponsive effect nor the 403.

A possible reason would be because you tried to vote a photo that wasn't flagged as accepted, but it's strange because vote buttons are not available when the photo isn't accepted.

For clarity's sake, I also removed the arrows on not accepted photos, they would always be disabled anyway on these.

See videos for my attempts to reproduce the error : 

https://user-images.githubusercontent.com/70256364/201898829-01c584fe-3df9-4d8f-a2b1-51bfe6b04925.mp4


https://user-images.githubusercontent.com/70256364/201898835-6135e459-f955-4df9-83f9-ec76f5d2eacf.mp4

Shriukan33:
> Please investigate why votes sometimes 403 (with nothing overly evident for the user) and why no tests have caught this.

They do : 

https://github.com/transport-nantes/tn_web/blob/7b51c878c02aac939ec9a1df0ad5486b337754e4/transport_nantes/photo/tests.py#L140-L153

JeffAbrahamson:
I'm sitting here voting repeatedly on two images, the only two that are accepted.  I am a new user, no cookies prior to my first photo visit.  The vote feedback is subtle, a dotted border appears, which appears to be selection rather than vote, as it isn't persistent if I refresh the page or go to next then prev photo.  The background colour doesn't change.

image

Meanwhile, the user receives no feedback besides the subtle border change.

The server is 403'ing (the server understands the request but refuses to authorize it).  This is happening because the CSRF cookie is not set, but I didn't debug beyond that.

Querying db.sqlite3, no votes are being recorded to the photo_vote table, which is at least consistent with a 4xx on voting.

Shriukan33:
> I've not been able to get a decent screen capture of it, because just trying to capture the screen image requires interacting with the snapshotter, which deselects.

I've been using flameshot for that, it has an option to add a countdown, so it lets you do whatever for x seconds before freezing the screen to let you capture the desired area 

https://doc.ubuntu-fr.org/flameshot#installation

Shriukan33:
> I'm sitting here voting repeatedly on two images, the only two that are accepted. I am a new user, no cookies prior to my first photo visit. The vote feedback is subtle, a dotted border appears, which appears to be selection rather than vote, as it isn't persistent if I refresh the page or go to next then prev photo. The background colour doesn't change.

Reproducing those steps inside vagrant still works on my ends, is your HEAD at e87c1715 ? 
I've force pushed a rebase on master just now, so no code change. I want to make sure you have all migrations done and code at latest version. 
New head is at 12e09415

JeffAbrahamson:
```
╭╴ ((edc03e1...) $%)╶╮
╰ jeff@morning:transport_nantes $ git fetch
╭╴ ((edc03e1...) $%)╶╮
╰ jeff@morning:transport_nantes $ git checkout origin/benja-photo-contest-improvement-960
HEAD is now at edc03e1 Add tests to photo app
╭╴ ((edc03e1...) $%)╶╮
╰ jeff@morning:transport_nantes $ git rebase master
HEAD is up to date.f
╭╴ ((edc03e1...) $%)╶╮
╰ jeff@morning:transport_nantes $ glt -10
edc03e1|2022-11-03|Benjamin Mourgues| (HEAD, origin/benja-photo-contest-improvement-960) Add tests to photo app
a3947bd|2022-10-19|Benjamin Mourgues| Add PhotoView
e41c658|2022-10-19|Benjamin Mourgues| Add utils function for Vote model
d078073|2022-10-19|Benjamin Mourgues| Add permissions to PhotoEntry model
505a179|2022-10-18|Benjamin Mourgues| Add Unique constraint to PhotoEntry sha1_name
53dc04e|2022-10-18|Benjamin Mourgues| Populate PhotoEntry sha1_name fields
ac3b5ad|2022-10-18|Benjamin Mourgues| Add sha1_name to PhotoEntry model
e29629c|2022-10-26|Benjamin Mourgues| Create vote model
af03de7|2022-10-18|Benjamin Mourgues| Add descriptive fields to photo entries
ccc3f32|2022-11-09|Jeff Abrahamson| (origin/master, origin/HEAD, master) Move mobilito flag from h1 to bottom of page.
╭╴ ((edc03e1...) $%)╶╮
╰ jeff@morning:transport_nantes $ 
```

Shriukan33:
> ```
> ╭╴ ((edc03e1...) $%)╶╮
> ╰ jeff@morning:transport_nantes $ git fetch
> ╭╴ ((edc03e1...) $%)╶╮
> ╰ jeff@morning:transport_nantes $ git checkout origin/benja-photo-contest-improvement-960
> HEAD is now at edc03e1 Add tests to photo app
> ╭╴ ((edc03e1...) $%)╶╮
> ╰ jeff@morning:transport_nantes $ git rebase master
> HEAD is up to date.f
> ╭╴ ((edc03e1...) $%)╶╮
> ╰ jeff@morning:transport_nantes $ glt -10
> edc03e1|2022-11-03|Benjamin Mourgues| (HEAD, origin/benja-photo-contest-improvement-960) Add tests to photo app
> a3947bd|2022-10-19|Benjamin Mourgues| Add PhotoView
> e41c658|2022-10-19|Benjamin Mourgues| Add utils function for Vote model
> d078073|2022-10-19|Benjamin Mourgues| Add permissions to PhotoEntry model
> 505a179|2022-10-18|Benjamin Mourgues| Add Unique constraint to PhotoEntry sha1_name
> 53dc04e|2022-10-18|Benjamin Mourgues| Populate PhotoEntry sha1_name fields
> ac3b5ad|2022-10-18|Benjamin Mourgues| Add sha1_name to PhotoEntry model
> e29629c|2022-10-26|Benjamin Mourgues| Create vote model
> af03de7|2022-10-18|Benjamin Mourgues| Add descriptive fields to photo entries
> ccc3f32|2022-11-09|Jeff Abrahamson| (origin/master, origin/HEAD, master) Move mobilito flag from h1 to bottom of page.
> ╭╴ ((edc03e1...) $%)╶╮
> ╰ jeff@morning:transport_nantes $ 
> ```

Do you do anything differently ? Maybe because I did modify some migrations, you'd need to redo you migrations from scratch ?


https://user-images.githubusercontent.com/70256364/203322306-ea0366c3-f982-4939-b80c-7d5bb16e61ca.mp4

JeffAbrahamson (approved):
This is a fatigue LGTM.  I think this still has issues, including the famous "let me develop this, then test it later", but re-reviewing a 9-commit 1300 LOC diff is just too painful to keep repeating.  We'll fix bugs later.

PR #974, opened 2022-10-30
Mobilito icons 968

JeffAbrahamson:


.gitignore
```diff
@@ -8,9 +8,9 @@ venv.vagrant
 Pipfile*
 
 # Django
-db.sqlite3
-media/
-static/
+/transport_nantes/db.sqlite3*
+/transport_nantes/media/
+/transport_nantes/static/
```
Shriukan33, line 13: We don't have these paths for media and static, they are transport_nantes/<app_name>/static

Shriukan33 (changes requested):
LGTM for images, the gitignore thing however I think I do understand the intention, I don't think it does that. 

You tell git to ignore transport_nantes directory /media and /static subdirs, but we don't have those anyway ? Also we don't want to ignore the app's static subdirs, as it makes JS and CSS ignored.

.gitignore
JeffAbrahamson: Exactly, and we very much want content in `transport_nantes/*/static` to  not be ignored.

What we want ignored is the stuff that collectstatic accumulates in `transport_nantes/static/` or that we test upload to `transport_nantes/media`.

Shriukan33: Hmm, I see, my statics don't go in those but maybe that's my local settings. LGTM then !

Shriukan33 (approved):

JeffAbrahamson:
Where do you put your static and media files?  Those are the usual places to configure for dev.

JeffAbrahamson:
We should probably do something to make this more likely to be uniform among devs.

Shriukan33:
> Where do you put your static and media files? Those are the usual places to configure for dev.

That's maybe because I don't run collectstatic locally, but everything stays in app subdirs

JeffAbrahamson:
> > Where do you put your static and media files? Those are the usual places to configure for dev.
> 
> That's maybe because I don't run collectstatic locally, but everything stays in app subdirs

That seems valid, but then it's also unaffected by this change!

Issue #966, opened 2022-10-27
Inheriting views in mobilito/views.py is weird


JeffAbrahamson:
Inheriting TutorialView from AddressFormView and RecordingView just so you can hav access to the tutorial choices has really weird code smell.

That commonly needed functionality should be pushed out into a simple non-view class (still in views.py) that encapsulates the behaviour you want to share.

PR #976, opened 2022-10-30
Refactor mobilito to avoid weird template view inheritance

JeffAbrahamson:
Several mobilito template views were inheriting from a common template view class, which was really strange and difficult to understand.  It did this in order to share code that handled tracking which tutorials had been seen, which to view next, and when to save state for the user.  In doing that, it ended up requiring some slighly strange overrides to the standard get() / get_context_data() from the django library.  Brief, it was a bit twisted and not really a natural pattern for django.

This commit refactors the code responsible for tracking tutorial view state to a separate class, thus simplifying the view code.  In the process, refactor the view code to use the new tutorial state class.

Part of #966.

### DIFF
--- a/transport_nantes/mobilito/tests.py
+++ b/transport_nantes/mobilito/tests.py
@@ -1,10 +1,38 @@
 from datetime import datetime, timezone
 
 from django.contrib.auth.models import User, Permission
-from django.test import TestCase, Client
+from django.test import TestCase, Client, RequestFactory
 from django.urls import reverse_lazy
+from http.cookies import SimpleCookie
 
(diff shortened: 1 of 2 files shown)


JeffAbrahamson:
The tests on this need to be substantially better.  I was hitting my head against the wall trying to get simple things like `request.COOKIES` to work properly whilst testing.  So that's something we need to come back to very soon.

In the mean time, I tried to test all the code paths that I could think of, though I'm sure I missed cases along the auth/no-auth boundaries especially.

JeffAbrahamson:
Closing, as this was included in another PR.

PR #987, opened 2022-11-07
Benja listview photo contest 986

Shriukan33:
Based on #963 

This add a simple, responsive and paginated view to all entries in photo contest.

Paginated to 4 items per page to save on loading time.
Added a small hover effect for desktop users.

Couldn't get the infinite loading to properly work in an hour, so dropped it.

### DIFF
--- a/transport_nantes/photo/views.py
+++ b/transport_nantes/photo/views.py
@@ -6,7 +6,7 @@ from django.http import (HttpRequest, HttpResponse,
                          HttpResponseForbidden, HttpResponseRedirect)
 from django.shortcuts import get_object_or_404
 from django.urls import reverse_lazy
-from django.views.generic import (TemplateView, CreateView, FormView)
+from django.views.generic import (TemplateView, CreateView, FormView, ListView)
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.core.exceptions import ObjectDoesNotExist
 
@@ -265,7 +265,7 @@ class PhotoView(FormView):
                 logger.error(
                     "Mailing list operation-pieton does not exist")
 
-        # Vote nature is a constant among the forms
+        # vote value is a constant among the forms
         vote_value = self.request.POST.get("vote_value", None)
         if vote_value == "upvote":
             vote_value = True
@@ -330,7 +330,7 @@ class PhotoView(FormView):
                 logger.error(
                     "Mailing list operation-pieton does not exist")
 
-        # Vote nature is a constant among the forms
+        # vote value is a constant among the forms
         vote_value = self.request.POST.get("vote_value", None)
         if vote_value == "upvote":
             vote_value = True
@@ -353,3 +353,9 @@ class PhotoView(FormView):
         )
 
         return HttpResponse(status=200)
+
+
+class PhotoListView(ListView):
+    queryset = PhotoEntry.objects.filter(accepted=True)
+    allow_empty = True
+    paginate_by = 4
--- a/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
+++ b/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
@@ -622,3 +622,19 @@ span.centered {
 .border-blue-light {
     border-color: var(--mobilitains-bleu);
 }
+.thumbnail {
+    position: relative;
+    display:block;
+    z-index:2;
+    cursor: pointer;
+    transition-property: all;
+    transition-duration: 0.3s;
+    transition-timing-function: ease;
+    -webkit-transition-property: all;
+    -webkit-transition-duration: 0.3s;
+    -webkit-transition-timing-function: ease;
+}
+.thumbnail:hover {
+    transform: scale(1.1);
+    z-index:3;
+}
--- a/transport_nantes/photo/templates/photo/photoentry_list.html
+++ b/transport_nantes/photo/templates/photo/photoentry_list.html
@@ -1,0 +1,49 @@
+{% extends 'asso_tn/base_mobilitain.html' %}
+
+{% block app_content %}
+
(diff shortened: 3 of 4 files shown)


JeffAbrahamson:
Please remove the import of ipware in view.py.

transport_nantes/photo/events.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Returns the number of upvotes for the given photo entry.
+    """
+    upvoting_users = User.objects.annotate(
+        latest_action=Subquery(
+            Vote.objects.filter(
+                photo_entry=photo_entry,
+                user=OuterRef('pk'),
+                captcha_succeeded=True,
+            ).order_by('-timestamp').values("vote_value")[:1]
```
JeffAbrahamson, line 37: This the same efficiency as `.first()`?

transport_nantes/photo/templates/photo/single_entry.html
```diff
@@ -0,0 +1,144 @@
...
+{% load static %}
+{% load compress %}
+{% load crispy_forms_tags %}
+
+{% block localscripts %}
+    {% compress js %}
+        <script src="{% static 'photo/js/vote_system.js' %}" defer></script>
+        <script type="text/javascript">
+            {% comment %}Those values are used in vote_system.js{% endcomment %}
+            {% if has_voted %}
```
JeffAbrahamson, line 12: This presents four different js files to the caching system, which increases the chances of cache miss.

Maybe we need to do that, but why aren't these values simply arguments to the function?

transport_nantes/photo/views.py
```diff
@@ -85,3 +96,266 @@ def get_context_data(self, **kwargs):
...
+
+        return context
+
+    def set_next_and_previous_arrows(self, context: dict, photo: 'PhotoEntry') -> None:
+        """
+        Set the next and previous photo arrows
+
+        context is passed by reference.
+        """
+        all_accepted_photos = list(PhotoEntry.objects.filter(accepted=True))
```
JeffAbrahamson, line 171: This fetches all photos on every view in order to check which are previous and next.  Even if the user never clicks those buttons.

Can we not be lazy and have links for showing the next and the previous photo, with argument the current photo?  Then in those functions we do a query with the right order() and a first() or last().

There's a subtly about user feedback if it's already the first or the last photo.  You can get around that by making the list circular.

transport_nantes/photo/views.py
```diff
@@ -85,3 +96,266 @@ def get_context_data(self, **kwargs):
...
+                mailing_list = MailingList.objects.get(
+                    mailing_list_token="operation-pieton")
+                subscribe_user_to_list(user, mailing_list)
+                logger.info(
+                    f"Subscribed user {user.email} to list {mailing_list}.")
+            except MailingList.DoesNotExist:
+                logger.error(
+                    "Mailing list operation-pieton does not exist")
+
+        # Vote nature is a constant among the forms
```
JeffAbrahamson, line 333: Vote nature?
Consistent vocabulary.

JeffAbrahamson (changes requested):

transport_nantes/photo/templates/photo/single_entry.html
Shriukan33: I count 1 file and 1 inline script. 

I don't understand what you mean, these variables are state. They are used in various vote_system.js functions to make it run properly. 

Could you please expand on the cache miss issue ? Do you suggest we don't compress the JS ?

transport_nantes/photo/views.py
Shriukan33: Photo entries aren't very numerous so this can be a concern, I think this is a premature optimization. 
If we went for that, then an ajax call would probably work, but I don't think this is necessary at this point.

transport_nantes/photo/views.py
Shriukan33: Fixed

transport_nantes/photo/views.py
JeffAbrahamson:
```
def view_to_show_image(image_sha1):
    this_image = photos.get(sha1_name=image_sha1)
    if not this_image.accepted:
        # 404 here.
    try:
        next_image = photos.filter(
            accepted=True, sha1_name__gt=image_sha1).order_by("sha1_name").first()
    catch photo.DoesNotExist:
        # This must exist, barring race conditions, because this_image is accepted.
        next_image = photos.filter(accepted=True).order_by("sha1_name").first()
    try:
        prev_image = photos.filter(
            accepted=True, sha1_name__lt=image_sha1).order_by("-sha1_name").first()
    catch photo.DoesNotExist:
        # This must exist, barring race conditions, because this_image is accepted.
        next_image = photos.filter(accepted=True).order_by("-sha1_name").first()

```

JeffAbrahamson:
This looks fine, but can't be accepted until `HEAD^` is accepted.

transport_nantes/photo/views.py
Shriukan33: Do you want me to refactor ?

transport_nantes/photo/templates/photo/single_entry.html
JeffAbrahamson: Do you still not understand this?

transport_nantes/photo/views.py
JeffAbrahamson: Yes, that was kinda the point...  ;-)

Whenever you have to disable an element that you display or usually display, it's good user interface design to ask yourself if a slight change in metaphor could make the flow easier.

JeffAbrahamson (approved):
LGTM, though at some level, this is a fatigue approval: 10 commits, 1300 LOC diff, multiple reviews.

Nothing is overly broken, though I still think we could do better on the left/right arrow and disabling arrows bit.

PR #996, opened 2022-11-09
Move mobilito flag from h1 to bottom of page.

### LINKED ISSUE #995: Move mobilito flag from h1 to bottom

JeffAbrahamson:
The flag doesn't work for me in dev, neither before not after this commit.  But since it works on beta and tests pass, I'm going to assume I didn't break anything. ;-)

Closes #995.


Shriukan33:
I've found why it doesn't display very well on dev, it's because there is a case that wasn't handled properly : when you are the author, the dropdown's content went blank because of conditions. 
I'll leave that to another PR and LGTM this one :)

JeffAbrahamson:
> I've found why it doesn't display very well on dev, it's because there is a case that wasn't handled properly : when you are the author, the dropdown's content went blank because of conditions. I'll leave that to another PR and LGTM this one :)

Well done on finding the bug.

PR #999, opened 2022-11-17
Benja make tbepanels 978

Shriukan33:
This PR introduces TopicBlogPanels

Panels are short html snippets that can be embedded in other topicblog contents through our markdown (using `[[panel:]]((panel-slug))`) or using the templatetag synthax `{% load panels %}{% panel 'panel-slug' %} `

The panels work the same way all other topicblog items work with a publication / versioning system.

The provided template is email client-friendly, as well as mobile friendly (it removes the image on smaller screens)

### DIFF
--- a/transport_nantes/topicblog/test_tn_links.py
+++ b/transport_nantes/topicblog/test_tn_links.py
@@ -24,32 +24,35 @@ class TnLinkParserTest(TestCase):
         self.assertEqual(self.parser.transform('[dog]((hat))'), '[dog]((hat))')
         self.assertEqual(self.parser.transform('[dog:cat]'), '[dog:cat]')
         self.assertEqual(self.parser.transform('[[hello'), '[[hello')
-        self.assertEqual(self.parser.transform('[[hello]]((goodbye))'), '[[hello]]((goodbye))')
-        self.assertEqual(self.parser.transform('[[hello:goodbye'), '[[hello:goodbye')
-        self.assertEqual(self.parser.transform('[[hello:goodbye]'), '[[hello:goodbye]')
+        self.assertEqual(self.parser.transform(
+            '[[hello]]((goodbye))'), '[[hello]]((goodbye))')
+        self.assertEqual(self.parser.transform(
+            '[[hello:goodbye'), '[[hello:goodbye')
+        self.assertEqual(self.parser.transform(
+            '[[hello:goodbye]'), '[[hello:goodbye]')
         self.assertEqual(self.parser.transform('dog [[hello:goodbye]] cat'),
                          'dog [[hello:goodbye]] cat')
 
     def test_bad_url(self):
         self.assertEqual(self.parser.transform(
             '[[action:Do something!]]((does-not-exist))'),
-                         don.action_button(reverse('topic_blog:view_item_by_slug',
-                                                   args=['does-not-exist']),
-                                           'Do something!'))
+            don.action_button(reverse('topic_blog:view_item_by_slug',
+                                      args=['does-not-exist']),
+                              'Do something!'))
 
     def test_buttons(self):
-        self.assertEqual(self.parser.transform('[[don:]]((give!))'), \
+        self.assertEqual(self.parser.transform('[[don:]]((give!))'),
                          don.bouton_don('give!'))
-        self.assertEqual(self.parser.transform('[[don:large]]((give!))'), \
+        self.assertEqual(self.parser.transform('[[don:large]]((give!))'),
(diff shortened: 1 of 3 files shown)


JeffAbrahamson:
Modifies tn_links.py without modifying test_tn_links.py.

Leaves a comment about a bug (that it apparently fixes), merely adding a comment about the comment.

JeffAbrahamson:
Three of the seven commits have no issue tag.

Shriukan33:
Updated tests and commit messages

transport_nantes/topicblog/test_tn_links.py
```diff
@@ -24,42 +24,54 @@ def test_broken_links(self):
...
 
-    ## Disabled due to captcha matching.
-    ## Think of better way to test [[news:*]]((...)).
-    #def test_news(self):
-    #    settings.csrf_token = 'some text that must exist'
-    #    self.assertEqual(self.parser.transform('[[news:kangaroo]]((aardvark))'), \
-    #                     render_inclusion_tag_to_html('newsletter',              \
-    #                                                  'show_mailing_list',       \
-    #                                                  {'mailinglist': 'kangaroo'}))
+    def test_news(self):
```
JeffAbrahamson, line 58: This isn't related in any way to the task at hand, even though it's a good fix.  It should really be a separate commit (with or without issue... ;-) and with commit message about test the tn-md news tag.

Meanwhile, aside, you have added the "panel" directive but haven't tested it.

transport_nantes/topicblog/urls.py
```diff
@@ -10,137 +10,215 @@
...
-    path('la/<slug:the_slug>/', views.TopicBlogLauncherView.as_view(),
-         name='view_launcher_by_slug'),
-    path('mlp/<slug:the_slug>/', views.TopicBlogMailingListPitchView.as_view(),
-         name='view_mlp_by_slug'),
 
-    path('e/i/<str:token>', views.beacon_view,
-         name='beacon_view'),
-
-    path('admin/t/view/<int:pkid>/', views.TopicBlogItemViewOne.as_view(),
+topicblogitem_urls = [
```
JeffAbrahamson, line 14: This is a good commit, but why is it in this PR?  This PR is long and has some minor issues, whereas this commit would have just been a simple approve (and so not needed to be re-read on next iteration).

JeffAbrahamson (changes requested):

transport_nantes/topicblog/urls.py
Shriukan33: This is because I've used the new url list format in this PR, creating a "topicblogpanel_urls" list in the next commit. So we would have to merge this change before TBPanels.

I can open a separate PR, but I thought this was a well isolated commit, enough to be reviewed once and approved without having to look at it twice. :)

transport_nantes/topicblog/test_tn_links.py
Shriukan33: I did actually test it, see : 
https://github.com/transport-nantes/tn_web/blob/6c5fa78caa585ce96b7b3b82e1a7073ab969b29f/transport_nantes/topicblog/templatetags/test_templatags.py#L202-L215

I didn't put it in the test_tn_links because to the difference of the other md stuff, the rendering aspect does rely on a database object witch is created in setUp. It was out of place in test_tn_links.py, in my opinion.

Shriukan33: Also, made issue + separate commit for the news tag, as requested :)

transport_nantes/topicblog/urls.py
JeffAbrahamson: It would have been approved already if separate, and you would have quietly rebased this one on it so as to use it...

transport_nantes/topicblog/test_tn_links.py
JeffAbrahamson: Testing the markdown and testing the templatetag are two different things.  The markdown must be tested.

Shriukan33: > {% tn_markdown "[[panel:]]((test-panel))" %} 

tn_markdown is the tag we use to make text go through our parser, it's **not** related to tbpanels, which have their own template tag, namely {% panel 'slug-here' %}

Shriukan33:
Added comments about where the test is located

Shriukan33:
Requesting re-review

JeffAbrahamson (approved):
LGTM.

PR #998, opened 2022-11-15
Benja fix report flag display

Shriukan33:
Fix bug discovered in #996 

Changes :
* Dropdown is not empty anymore when you are the sessio's author
* Text color is back to its original color in the modal (used to be white on white because of #990 )

Issue #981, opened 2022-11-03
Add map to mobilito session reports


JeffAbrahamson:

Shriukan33:
No assignees, are you handling it ?

Otherwise, what do you mean by map ? An actual embedded map on the web visualization ?

Shriukan33:
Should this map appear in the email summary as well ? That would mean we need to generate a PNG from a map, which can complicate things.

I don't think we should aim to get everything in the mail given the complexity layer it adds, simply incentivize clicking on the web version button, at least for now.

Shriukan33:
I did some research about how to go from lat/lng tuple to rendered png, there are services that provide that but it looks quite complex and long for that purpose. 

https://print.get-map.org/about/api/
https://github.com/hholzgra/ocitysmap/

Or maybe having a tile render server of our own 
https://github.com/enzet/map-machine#tile-server

There is no easy way if we want to add that into emails, I think.
Maybe we should encourage clicking to discover more info about the session, instead of trying to render it before sending. That would be much simpler implementation wise, not sure this is worth the hassle. 

What do you think ?

PR #1003, opened 2022-11-22
Benja mobilito map 981

Shriukan33:
Based on #998

This commit adds a map based on the latitude and longitude recorded for a mobilito session.
This is the same map we use for geolocation form.

Part of #981


JeffAbrahamson (approved):
LGTM.

Issue #993, opened 2022-11-08
post-session view update


JeffAbrahamson:
Rather than going to a thank you page, we should send the person to the session page (which might still have a toast saying "awesome, well done".  But that page is then directly shareable.

The session page, for everyone, should have a button for making new sessions.  We want this to be viral.

JeffAbrahamson:
Remaining:
* Remove "thanks" page if no longer needed.
* Modifications to session page to encourage sharing.
* A thank-you on the session page (e.g., post with an extra "thank-you" argument so that the argument isn't present when shared).

Shriukan33:
How do you want to encourage sharing ? I think we usually avoid twitter/FB/Google links on our website. Although the integration here would make sense, but I think this would force us to have a cookie banner ?

JeffAbrahamson:
> How do you want to encourage sharing ? I think we usually avoid twitter/FB/Google links on our website. Although the integration here would make sense, but I think this would force us to have a cookie banner ?

Twitter is easy enough, it can be done with a GET.
Facebook is harder, but a bit of searching proposes solutions.

This jQuery plugin looks promising:
https://panzi.github.io/SocialSharePrivacy/

[This article](https://www.business2community.com/social-media-articles/why-most-share-buttons-go-against-your-privacy-policy-02199051) suggests that (in some contexts, at a certain point in time) 2/3 of users shared via copy/paste rather than share buttons.  But share buttons remain important.

JeffAbrahamson:
> > A thank-you on the session page (e.g., post with an extra "thank-you" argument so that the argument isn't present when shared).
> 
> Like a django message, it would be displayed once and never again ?

Yup.  Post to the URL rather than GET it, just so the arg isn't in the URL.  The arg has the thank you.  You don't even need a different handler for GET vs POST, you just check for the arg.  You don't care how it arrives.

Shriukan33:
I added the plugin to see how it looks, and damn it shows it's not been updated for 10 years already : 
![image](https://user-images.githubusercontent.com/70256364/204299365-8654c293-5c16-48af-885c-a0dded3ea4ce.png)

However it does the job, no resources fetch to FB, Twitter & co. Clicking on it enables the button and we can share on the different platforms.

I'm going to see if there is a way to have more modern images.

Shriukan33:
The inline version show better, I removed most of icons to keep it light : 
[...]

PR #1004, opened 2022-11-22
Redirect to session summary instead of thank you page

Shriukan33:
Part of #993

The issue states that we need a button to make another session, and I agree on this but we have several PR pending on session summaries, I don't want this to be a merge nightmare so I'm postponing this change for later, it's not hard anyway.


JeffAbrahamson:
> Part of #993
> 
> The issue states that we need a button to make another session, and I agree on this but we have several PR pending on session summaries, I don't want this to be a merge nightmare so I'm postponing this change for later, it's not hard anyway.

That last seems wise. ;-)

JeffAbrahamson (approved):
LGTM.

PR #1005, opened 2022-11-23
Make unsub button less visible

### LINKED ISSUE #991: Unsub button is too visible on gmail
on mobile unsub button renders like so : 
image

Font should be smaller, and the button not as visible

Shriukan33:
The unsub class was drawing attention to the unsub button.

Closes #991


JeffAbrahamson (approved):
LGTM.

Issue #1006, opened 2022-11-23
Add a chronometer to mobilito recording


Shriukan33:
I think that a chronometer is a good way to know how long you've been recording and can help toward recording longer sessions if coupled with a "goal" bar. 

Say we set a bar that fills in 5 minutes with an indication that it would be nice if you reached that point ?

Shriukan33:
What about simply adding the chronometer for now ?

Shriukan33:
Maybe filling another bar on top of the other could achieve the desired effect, like first bar is green, second yellow, last red-ish 
All those with some text that cheers for more.

JeffAbrahamson:
We might want to test with coloured pencils on paper first.  It's way faster. ;-)

PR #1007, opened 2022-11-23
Add a simple chronometer to mobilito recording screen

Shriukan33:
Part of #1006

image


Shriukan33:
I toyed a bit with Bootstrap's progress bar, and came with something that looks decent

In this example, I've set each threshold to 3 seconds, but we can have different durations for each !

mobilito_progress_bar

Shriukan33:
Added cheer texts

mobilito_progress_bar_cheer

PR #1010, opened 2022-11-28
Add temporary message to thank user for making a Mobilito session

Shriukan33:
The message framework allows us to push ephemeral messages to the users. Sharing the link or even reloading the page will remove the message.

image

Part of #993

### DIFF
--- a/transport_nantes/mobilito/views.py
+++ b/transport_nantes/mobilito/views.py
@@ -18,6 +18,7 @@ from typing import Union
 from asso_tn.utils import make_timed_token, token_valid
 from authentication.views import create_send_record
 from django.conf import settings
+from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.core.mail import EmailMultiAlternatives
 from django.db.models import Q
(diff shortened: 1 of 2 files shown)


transport_nantes/mobilito/views.py
```diff
@@ -282,19 +283,8 @@ def post(self, request: HttpRequest, *args, **kwargs) \
...
-                - mobilito_session.start_timestamp
-            ).total_seconds() // 60
-        )
-        request.session['number_of_pedestrians'] = number_of_pedestrians
-        request.session['number_of_bicycles'] = number_of_bicycles
-        request.session['number_of_cars'] = number_of_cars
-        request.session['number_of_public_transports'] = \
-            number_of_public_transports
+        messages.add_message(
+            request, messages.SUCCESS, 'Merci de votre participation !')
```
JeffAbrahamson, line 287: Do you think you could think of something more dry and corporate?  ;-)

I'm not good at this in French.  Do you have any fun friends who could suggest a phrase?

JeffAbrahamson (approved):
LGTM, but let's find something more motivating than "merci de votre comprehension"...

transport_nantes/mobilito/views.py
Shriukan33: I'm doing either cheesy or Puppy on acid texts, or dry corporate, there is no in-between :laughing:

You're underselling yourself, you do have a quite good french, only some misses on spelling but nothing to be ashamed of. You do speak it probably better than my french-born-and-raised-in-France dad to be honest :laughing:

JeffAbrahamson: Good thing your dad doesn't hang out on github. ;-)

Shriukan33:
Rebased this on #1012, I'm not very inspired for something that says thanks without being too cheesy, let's merge this and make a new commit once we know what we want to be there :)

JeffAbrahamson:
I already LGTM'd this.

It now has merge conflicts, so back to you...

PR #1011, opened 2022-11-28
Add social sharing button on mobilito session

Shriukan33:
Tracker blocking extensions like Ghostery and Disconnect will block the twitter button if enabled.

Part of #993

image

### DIFF
--- a/transport_nantes/mobilito/static/mobilito/css/session_summary.css
+++ b/transport_nantes/mobilito/static/mobilito/css/session_summary.css
@@ -13,3 +13,12 @@ a.dropdown-item {
  #dropdownMenuLink {
     color: var(--body-text);
 }
+.settings_info {
+    display: none !important;
+}
+li.help_info > div.info {
+    z-index: 99999 !important;
+}
+div.dummy_btn.tweet {
+    width: auto !important;
+}
--- a/transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
+++ b/transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
@@ -3,6 +3,43 @@
 {% load static %}
 {% load compress %}
 
+
+{% block head %}
+    {{ block.super }}
+    {# https://panzi.github.io/SocialSharePrivacy/ #}
+    <script type="application/x-social-share-privacy-settings">
+        {
+            "path_prefix":"http://panzi.github.io/SocialSharePrivacy/",
+            "layout":"line",
+            "services":
+                {
+                    "buffer":{"status":false},
+                    "facebook":{"status":false},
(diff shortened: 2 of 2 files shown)


JeffAbrahamson:
Why do tracker-blocking extensions block the twitter button?
This is the one button that's really easy to do as a simple GET + fa icon...

JeffAbrahamson (changes requested):
This is too local.  If we're going to add sharing, we should make everything shareable, not just mobilito pages.

We are adding a dependency on someone's github repo and on an unversioned bit of code.  Why is this a good idea?

If some bit of this causes privacy issues, then the whole thing should be unacceptable to us.

Shriukan33:
> Why do tracker-blocking extensions block the twitter button? This is the one button that's really easy to do as a simple GET + fa icon...

Because the button that is loaded (the actual share on twitter button) is fetched on twitter's servers, apparently. 



> This is too local. If we're going to add sharing, we should make everything shareable, not just mobilito pages.

Can be quite easily done, we'd need to add to base template the scripts, and the div where appropriate. 



> We are adding a dependency on someone's github repo and on an unversioned bit of code. Why is this a good idea?
> 
> If some bit of this causes privacy issues, then the whole thing should be unacceptable to us.

I don't know what to tell you, you wanted to have social sharing buttons that didn't send data to third parties. That's what it does, as so far as I can see.

What's your requested change ? Getting rid of if or making it part of default template ?

JeffAbrahamson:
> > We are adding a dependency on someone's github repo and on an unversioned bit of code. Why is this a good idea?
> > If some bit of this causes privacy issues, then the whole thing should be unacceptable to us.
> 
> I don't know what to tell you, you wanted to have social sharing buttons that didn't send data to third parties. That's what it does, as so far as I can see.
> 
> What's your requested change ? Getting rid of if or making it part of default template ?

You know about the npm incident, because of depending on 7 lines of code?

https://www.theregister.com/2016/03/23/npm_left_pad_chaos/

When you see that you're about to put something like this

    tps://panzi.github.io/SocialSharePrivacy/javascripts/jquery.socialshareprivacy.min.autoload.js

in your code, you should stop and think: what happens if this person pushes a bug or deletes his repo?  Is this dependence worth the risk?

Alternatives, depending on the size and complexity of the thing included, are forking the repo so that you upgrade on _your_ schedule, or just copying the code (what should have happened with that short npm snippet) or rewriting.

In this case, we're looking at ten year old code that's not particularly complex and that's no longer maintained.  We get very little benefit from maintaining a dependency.  It sounds like it's also simple enough that we can extract the little bit that we need and so be done with it.

Shriukan33:
I'll go with a simpler implementation using FA icons and plain urls, as discussed in the issue :)

PR #1012, opened 2022-11-29
Benja merge cleanup 971

Shriukan33:
This PR does a cleanup of the #997 PR.

Mostly, it's improvements on responsiveness and code quality on front-end, and a minor change in the backend to use a non GUI backend for matplotlib.

### DIFF
--- a/transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
+++ b/transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
@@ -48,6 +48,9 @@
 
 
 {% block app_content %}
+
+{# Used to get the session's sha1 in Javascript #}
+{{ mobilito_session.session_sha1|json_script:'mobilito_session_sha1'}}
 <div class="container">
     <div class="d-flex flex-wrap col-12">
         
--- a/transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
+++ b/transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
@@ -49,132 +49,127 @@
 
 {% block app_content %}
 <div class="container">
-    <div class="row">
-	<div class="col-lg-12">
-	    {# name of recorder, some day #}
-	    <p>{{ mobilito_session.start_timestamp|timesince:mobilito_session.end_timestamp }}
-		à partir de {{ mobilito_session.start_timestamp|timezone:'Europe/Paris'|date:"l j F Y à G:i" }}.
-	    </p>
-	    <p><b>{{ mobilito_session.location }}</b></p>
-
-        <style>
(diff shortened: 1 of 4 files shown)


Shriukan33:
Actually there is a line that was deleted and that is important for flag feature, I'm re-adding it as well as adding a selenium test to catch it in the future.

Shriukan33:
Added the aforementioned test !

transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
```diff
@@ -48,133 +48,131 @@
...
 
-        {% with start=mobilito_session.start_timestamp end=mobilito_session.end_timestamp %}
-                <div class="dropdown show">
-                    <a href="#" role="button" id="dropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                        <i class="fa-solid fa-ellipsis ml-md-3"></i>
-                    </a>
-                    <div class="dropdown-menu dropdown-menu-right" aria-labelledby="dropdownMenuLink">
-                        <a class="dropdown-item" href="#"
-                        {% if user_has_reported_this_session or request.user == mobilito_session.user.user %}
-                            style="pointer-events: none; color: grey;"
```
JeffAbrahamson, line 108: Would this be better as mobilitain-grey?

JeffAbrahamson (approved):
LGTM.

Comments, for future commits:
* the flag doesn't display for me on beta
* the map is rather tall and, as the screen shrinks, stays that tall, meaning on mobile it takes quite a lot of space even though it's not the main point

transport_nantes/mobilito/templates/mobilito/mobilito_session_summary.html
Shriukan33: This is just the font color when one has already flagged a session or is unable to. I don't believe it warrants a special color other than the internet wide grey for disabled stuff :slightly_smiling_face:

Shriukan33:
> LGTM.
> 
> Comments, for future commits:
> 
>     * the flag doesn't display for me on beta


It's still an ellipsis for now, and it does display but it's easily miss-able : 
image

I think the flag icon will be more noticeable while the color scheme is quite discreet, need to find a balance here.

> 
>     * the map is rather tall and, as the screen shrinks, stays that tall, meaning on mobile it takes quite a lot of space even though it's not the main point

I can limit its height, it didn't hit me as "too big" to be honest but maybe it is. I mean, I can understand you may find it too big.

JeffAbrahamson:
Tracking map size at #1013, flag at #1027.

Shriukan33:
Thanks for opening issues ! :heart_eyes:

PR #1028, opened 2022-11-30
Benja create presse release index 1019

Shriukan33:
This PR adds a view to list all published press released on a single page, in the same fashion @mcmikashi designed the press app. 
I reused his template, tweaked it a bit to fit topicblog and here it is. 

Deployed to beta, but as I'm not sure this will stay, here is a screenshot : 

image

(on the screenshot you see thrice the same image and nearly same title, they do have different slugs however, so this works !)

### DIFF
--- a/transport_nantes/topicblog/urls.py
+++ b/transport_nantes/topicblog/urls.py
@@ -73,6 +73,9 @@ topicblogemail_urls = [
          name='send_email'),
 ]
 topicblogpress_urls = [
+    path("communiques-de-presse/",
+         views.TopicBlogPressIndex.as_view(),
+         name="press_releases_index"),
     path('p/<slug:the_slug>/',
          views.TopicBlogPressView.as_view(),
          name='view_press_by_slug'),
--- a/transport_nantes/topicblog/views.py
+++ b/transport_nantes/topicblog/views.py
@@ -10,7 +10,7 @@ from django.core import mail
 from django.core.exceptions import ImproperlyConfigured
 from django.db import IntegrityError
 
-from django.db.models import Count, Max
+from django.db.models import Count, Max, Subquery, OuterRef, F
 from django.dispatch import receiver
 from django.http import (Http404, HttpResponseBadRequest,
                          HttpResponseServerError, FileResponse)
@@ -1080,8 +1080,31 @@ class TopicBlogPressSend(PermissionRequiredMixin, LoginRequiredMixin,
     success_url = "/"
 
 
+class TopicBlogPressIndex(ListView):
+    model = TopicBlogPress
+    template_name = 'topicblog/topicblogpress_index.html'
+    context_object_name = 'press_releases_list'
+    paginate_by = 10
+
+    def get_queryset(self) -> list:
+
+        # Subquery preparation
+        latest_press_release = TopicBlogPress.objects.filter(
+            slug=OuterRef('slug'),
+            publication_date__isnull=False).order_by('-publication_date')
+
+        queryset = TopicBlogPress.objects.annotate(
+            latest_press_release=Subquery(latest_press_release.values(
+                'publication_date')[:1])).filter(
+            publication_date=F('latest_press_release')).order_by(
+            '-publication_date')
+
+        return queryset
+
+
 ######################################################################
 # TopicBlogLauncher
+
 
 class TopicBlogLauncherViewOnePermissions(PermissionRequiredMixin):
     """Custom Permission class to require different permissions
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,7 @@ autopep8==1.5.7
 awscli==1.19.59
 backcall==0.1.0
 backports.zoneinfo==0.2.1
+beautifulsoup4==4.11.1
 billiard==3.6.4.0
 boto3==1.17.59
 botocore==1.20.59
@@ -116,6 +117,7 @@ selenium==4.1.0
 six==1.14.0
 sniffio==1.2.0
 sortedcontainers==2.4.0
+soupsieve==2.3.2.post1
 sqlparse==0.3.0
 stripe==2.55.0
 toml==0.10.2
--- a/transport_nantes/topicblog/models.py
+++ b/transport_nantes/topicblog/models.py
@@ -1,12 +1,16 @@
 from datetime import datetime, timezone, timedelta
 import logging
+from typing import Union
 
+import bs4
+from bs4 import BeautifulSoup as bs
 from django.contrib.auth.models import User
 from django.forms import ValidationError
 from django.urls import reverse
 from django.db import models
 
 from mailing_list.models import MailingList
+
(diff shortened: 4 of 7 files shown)


JeffAbrahamson:
You're missing markdown parsing...

JeffAbrahamson:
The issue prominently includes "Look at other sites for examples of how this is typically done, then do something simple, quick, and relatively least common denominator".  That's curiously absent. ;-)

Here are a few examples:
* [GP France](https://www.greenpeace.fr/communiques-de-presse)
* [AI France](https://www.amnesty.fr/rapports)
* [WWF](https://www.worldwildlife.org/about/news-press)
* [Sierra Club](https://www.sierraclub.org/press-releases)

The visual aspect of GP and AI is appealing, but, as you note, it depends on a history of using images that work in this context.  We haven't.  We could push against the grain, or we could accept that.

The WWF page's formatting is probably the most adaptable to our needs.

Wdyt?

Shriukan33:
> The issue prominently includes "Look at other sites for examples of how this is typically done, then do something simple, quick, and relatively least common denominator". That's curiously absent. ;-)
> 
> Here are a few examples:
> 
>     * [GP France](https://www.greenpeace.fr/communiques-de-presse)
> 
>     * [AI France](https://www.amnesty.fr/rapports)
> 
>     * [WWF](https://www.worldwildlife.org/about/news-press)
> 
>     * [Sierra Club](https://www.sierraclub.org/press-releases)
> 
> 
> The visual aspect of GP and AI is appealing, but, as you note, it depends on a history of using images that work in this context. We haven't. We could push against the grain, or we could accept that.
> 
> The WWF page's formatting is probably the most adaptable to our needs.
> 
> Wdyt?

I'm ok with WWF's !
Let me get back to it.

Shriukan33:
So I redid this issue and came to something that does display the Markdown, but it has its flaws : 

image

As you can see, the markdown is sometimes even bigger than the title, which is sober. 
And the truncate can cut in the middle of a markdown, like there : 

image

I think we could maybe create a "first paragraph" property like described in this [SO topic](https://stackoverflow.com/a/69622100/16279937) ?

transport_nantes/topicblog/urls.py
```diff
@@ -10,137 +10,188 @@
 """
 
 app_name = 'topic_blog'
-urlpatterns = [
```
JeffAbrahamson, line 13: Why is commit a488753e part of this PR?  It appears completely unrelated?

It's important not to put irrelevant stuff in PRs.  It just slows down review.  This is something you wrote over a week ago, it was a no brainer approve, and it keeps riding on other PRs, which just delays it getting approved.

transport_nantes/topicblog/views.py
```diff
@@ -1080,9 +1080,44 @@ class TopicBlogPressSend(PermissionRequiredMixin, LoginRequiredMixin,
...
+    model = TopicBlogPress
+    template_name = 'topicblog/topicblogpress_index.html'
+    context_object_name = 'press_releases_list'
+    paginate_by = 10
+
+    def get_queryset(self) -> list:
+
+        # Get the list of slugs that have been published at least once
+        all_published_slugs = (
+            TopicBlogPress.objects.filter(publication_date__isnull=False)
```
JeffAbrahamson, line 1093: What guarantees that you select the slug with the maximum `publication_date`?

JeffAbrahamson: And, I see no tests.

transport_nantes/topicblog/views.py
```diff
@@ -1080,9 +1080,44 @@ class TopicBlogPressSend(PermissionRequiredMixin, LoginRequiredMixin,
...
+            .values_list('slug', flat=True)
+            .distinct()
+        )
+
+        # Get the ids of the most recently published press releases for each
+        # slug
+        all_published_press_releases = []
```
JeffAbrahamson, line 1100: This should happen server-side.  Yes, we are small enough that this will fit in memory with small transfer time.  But it's not good programming practice to leave time bombs like this.  It's like a programmer in 1980 assuming that two digits is good for a year, or a programmer more recently assuming that 32 bits is enough for a timestamp.  It works the day you write it, but it's predictably going to fail some day unless your software isn't used.

JeffAbrahamson (changes requested):
Needs tests.

Commit comments must separate paragraphs with a blank line.

Less important, because we can fine-tune it:  Text size in the list gives everything the same importance.  Also, the preview is quite long and can take the better part of a page.  We should limit it to a line or two and in slightly smaller type.

transport_nantes/topicblog/urls.py
Shriukan33: https://github.com/transport-nantes/tn_web/pull/1053

transport_nantes/topicblog/views.py
Shriukan33: Did you, by any chance, miss the changes in `test_topic_blog_part_2.py` ? 

Also, the point of the all_published_slugs isn't to get the currently published items, only their slugs. It's a first step.

transport_nantes/topicblog/views.py
Shriukan33: It does happen in Django, so server-side. 
A list of integer is quite reasonable, especially since we are not likely to hit hundreds of millions of published items.

You are trying to optimize a non-issue here, in my opinion.
Yes, the code might not scale if you published millions of press releases, but before then, all this would probably change.

Shriukan33:
Tests were already present
Rebased on master
Increased title's font size a bit
Reduced the number of chars before ellipsis

transport_nantes/topicblog/views.py
JeffAbrahamson: > It does happen in Django, so server-side. A list of integer is quite reasonable, especially since we are not likely to hit hundreds of millions of published items.
> 
> You are trying to optimize a non-issue here, in my opinion. Yes, the code might not scale if you published millions of press releases, but before then, all this would probably change.

I'm trying to optimize a junior developer so that he grows and makes fewer mistakes, writes less code that flags him out as not understanding the finer points of his craft.

JeffAbrahamson:
> So I redid this issue and came to something that does display the Markdown, but it has its flaws :
> 
> image
> 
> As you can see, the markdown is sometimes even bigger than the title, which is sober. And the truncate can cut in the middle of a markdown, like there :
> 
> image
> 
> I think we could maybe create a "first paragraph" property like described in this [SO topic](https://stackoverflow.com/a/69622100/16279937) ?

Yeah, _sigh_, this is annoying.  The SO article argues (essentially) that we render the markdown, then extract the first paragraph, which I think we'd then to strip of all html style (so extract only text) and then render.

Shriukan33:
(Just in case, I'm requesting review here)

JeffAbrahamson (approved):
LGTM.

Issue #1027, opened 2022-11-30
Sort out mobilito content flagging


JeffAbrahamson:
* Make the mobilito ellipsis a flag and position it so it's findable but doesn't steal.
* Probably do something more like what SO does, where the flag presents an easy dialog.

![image](https://user-images.githubusercontent.com/1220683/204751473-efa656a9-5b85-41b4-8ab2-42c3523ad8a8.png)

Shriukan33:
I agree the flag should be more visible. 
I'm not sold on the fact we need to improve the flag system just yet : Yes the radio select is friendlier, but do we want to spend more time on this ? If we get more flags, i'd be more enclined to get back on it, for now it's good enough to me.

PR #1029, opened 2022-11-30
Move flag session button under the address field in mobilito

Shriukan33:
The most likely reason someone might report a session is because of the free field that is user provided.
The color is dark enough so it doesn't scream click me, but you can still see if it if you look for it.
Part of #1027


JeffAbrahamson (approved):
LGTM.

JeffAbrahamson:
Re-open #1027 (let's see if github does...), this only sorts half of that issue.

It's reasonable to deprioritise the other half, but the issue remains.

PR #1037, opened 2022-12-01
Better log GCP api calls

### LINKED ISSUE #1034: log GCP API calls
As we do with SES, create a new log channel for GCP and log every request: a log line with the request, a log line with the response (info if 200, warn otherwise, error on communication error).

Shriukan33:
In case of a problem with reverse geocoding issues (like badly configured API keys), we were not logging enough information for debugging purpose.

Maps API may not return a formatted address, and that case wasn't handled in logs and would leave the user without a clue about why their screen wasn't responding to their clicks.

This commit adds logs for reverse geocoding and add an alert to the destination of the user in the event we don't get a formatted address from the reverse geocoding service, informing them that the request has been processed but couldn't give a satisfying answer at the moment.

Closes #1034


transport_nantes/mobilito/views.py
```diff
@@ -725,7 +725,16 @@ def post(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
                 f"latlng={lat},{lng}&key={MAPS_API_KEY}"
             )
             response = requests.get(url)
-            return JsonResponse(response.json())
```
JeffAbrahamson, line 728: You must info log all requests we send and all responses received.  After that, info log success, warn or error on non-success.

The purpose of logging is understanding what happened in production.  You do not have the ability to teleport to the user, back in time, with development tools, to find out what happened.

transport_nantes/mobilito/static/mobilito/js/geolocation_map.js
```diff
@@ -99,12 +99,16 @@ function onMapClick(e) {
...
-        $('#id_longitude').val(`${e.latlng.lng}`)
+        if (data.status === "OK") {
+            let addr = data.results[0].formatted_address;
+            marker = L.marker(e.latlng).addTo(map)
+                .bindPopup(`${addr}`).openPopup();
+            $('#id_location').val(`${addr}`)
+            $('#id_latitude').val(`${e.latlng.lat}`)
+            $('#id_longitude').val(`${e.latlng.lng}`)
+        } else {
+            alert("La localisation n'est pas disponible pour le moment.")
```
JeffAbrahamson, line 110: This needs a pointer to documentation so that any future dev can understand that the  specification of this service says that "data.status will be 'OK' on success, and anything else on failure with no further nuance of meaning".

That would be a rather rubbish specification for an service, however, so a future dev would be right in wondering if the error was the dev who wrote our code, and so go off searching.  You need to cut that search short by explaining here that the foreign service really provides such a poor interface and providing a pointer to it.

JeffAbrahamson (changes requested):
Traffic to/from GCP must log to a separate log file.  That means configuring a separate log file as part of this commit.

Further texture needs to be logged to the GCP log with a summary to our main log.  Think about what the dev exploring a problem will encounter and need.

For example, yesterday when I was seeing og bugs, the logs said nothing at all, not that it was trying, not that it was succeeding, not that it was failing.  That's unhelpful at best.

JeffAbrahamson:
I once suggested that you type your commit comments to google doc first in order to get spelling and grammar correction.  It wasn't a joke: people who don't know you will judge you on your prose and on the quality of your code.  Tools help you improve.

(The commmit comment on this PR is a bit challenging to read. ;-)

transport_nantes/mobilito/static/mobilito/js/geolocation_map.js
Shriukan33: Added link to maps' docs.

They actually have other status codes, but the ultimate result on the user's end is that they don't get a formatted address.

transport_nantes/mobilito/views.py
Shriukan33: Done in the new logger, added details into the reverse geocoding-specific log file.

transport_nantes/transport_nantes/settings.py
```diff
@@ -202,6 +202,12 @@
             # 'when': 'd',
             # 'utc': True,
             # 'backupCount': 10,
+        },
+        'rgeocoding_handler': {
+            'level': 'INFO',
+            'class': 'logging.FileHandler',
+            'filename': settings_local.LOG_DIR + "rgeocoding.log",
```
JeffAbrahamson, line 209: Let's not make a separate log for every GCP service.  Perhaps call it tn_gcp.log?

Shriukan33:
Renamed to gcp, thanks !

PR #1038, opened 2022-12-01
Limit mobilito's session summary's map's height

### LINKED ISSUE #1013: Fiddle with mobilito /session/ map size
It takes more vertical space than its importance merits.  It's no more important than the text, the real content is the plots and the information about modal share.  Yet on many screens it steals attention and even displaces the real content below the fold.

Perhaps consider reducing to 1/3 its current height?

Shriukan33:
The map was taking too much space especially on smaller devices. This commit limits the height to 23% of the screen's height.

It's a compromise to be still able to interact with it while keeping it small enough.

image

Closes #1013


JeffAbrahamson (approved):
LGTM.

PR #1053, opened 2022-12-05
Re-organize topicblog urlspattern

### LINKED ISSUE #1001: Re-organize the TB urls
I would like them to be grouped by object instead of vaguely grouped by functionality. 

I think this is better because we usually work on one particular item, and it's easier to have a full picture of it that way. 

It makes adding new TB objects easier, you simply copy-paste a previous object and modify its short name, most of the time.

Shriukan33:
The list became too long and it was hard to find a topicblog object's related urls. By organizing the urls by topicblog object, it is easier to duplicate for future topicblog objects, and easier to find the urls for a topicblog object.

Closes #1001


JeffAbrahamson (approved):
LGTM.

PR #1055, opened 2022-12-05
Fix too permissive query in photo app

### LINKED ISSUE #1054: Only the first anon user is presented the modal form
If another "Anonymous User" has been through the modal form, it doesn't show up for another Anonymous User (because the server thinks that Anonymous User has already voted, while it's a different Anonymous User ...)

Shriukan33:
The query used to be too permissive :
Q(user=user) could be interpreted as Q(user=<User>) but also Q(user=None).
As every anonymous profile cast a vote with user=None, the direct effect was that as long as one anonymous visitor had voted for a photo, every anonymous was marked as having voted in the past.

Closes #1054

### DIFF
--- a/transport_nantes/photo/views.py
+++ b/transport_nantes/photo/views.py
@@ -136,7 +136,9 @@ class PhotoView(FormView):
             # we need to check if the user has already voted for this photo
             # using its session id as well.
             if Vote.objects.filter(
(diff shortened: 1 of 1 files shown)


JeffAbrahamson (approved):
LGTM.

Issue #1049, opened 2022-12-04
photo gallery: varying image sizes lead to visual bounce


JeffAbrahamson:
Arrow and like buttons bounce around from one photo to the next.  We should move them above the photo so that they are stable.

PR #1056, opened 2022-12-05
Photo gallery : move next/previous arrows and vote buttons on top

Shriukan33:
The different shapes and sizes would move the buttons around, this fixes this behavior.

Part of #1049


JeffAbrahamson (approved):
LGTM.

PR #1058, opened 2022-12-05
Make arrows circular in photo gallery

### LINKED ISSUE #1045: photo gallery: circular arrows
The arrows should model a circular queue so that they don't disable at end of sequence.  We have no incentive to say "no" to our visitors.

Shriukan33:
First photo's previous is the last photo,
and the last photo's next is the first photo.

Closes #1045

### DIFF
--- a/transport_nantes/photo/tests.py
+++ b/transport_nantes/photo/tests.py
@@ -1,10 +1,18 @@
 from pathlib import Path
 
 from django.contrib.auth.models import User
+from django.contrib.staticfiles.testing import StaticLiveServerTestCase
 from django.core.files.images import ImageFile
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import Client, TestCase
 from django.urls import reverse
+from selenium import webdriver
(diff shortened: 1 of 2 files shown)


JeffAbrahamson (approved):
LGTM, more or less.

Beginning developers tend to pull data from databases to manipulate it and then do something.  If those developers become more mature, they learn that they should do some of this work in the database.

I would really like you to learn sooner rather than later that doing things like `list(PhotoEntry.objects.filter(...))` is an anti-pattern.

Shriukan33:
So I did find a query that would still hit the database once.
The performance seem to be a bit reduced (from ~115ms to load a photo:details page to 130ms) but at least it doesn't load into memory every photo ID like it did.

I also added Selenium tests to make sure we can actually click, fill the forms and vote on photo pages, as it involves some JS it seemed to be the best solution.

Shriukan33:
Re-requesting review

JeffAbrahamson (approved):
LGTM.

PR #1064, opened 2022-12-07
Fix failing Mobilito Test

### LINKED ISSUE #1063: Mobilito Selenium test sometimes fail
While working on an unrelated app (photo), I noticed that a mobilito test failed, but not all the time.
It seems to have to do with selenium sometimes having time to load certain elements, so it might be vulnerable to CPU's performance.

Shriukan33:
I tried waiting for the button to be available and visible, but while the test seemed to fail less often, it still failed sometimes. I'm not sure about what exactly causes the error to raise, but adding some wait seems to fix it.

Closes #1063


JeffAbrahamson:
The failing test is telling you something.  Pay attention to it.  If you tell it to be quiet, it will be quiet, but it won't fix the problem.

https://en.wikipedia.org/wiki/Race_condition

PR #1066, opened 2022-12-12
Add Twitter and Facebook social sharing buttons

Shriukan33:
The base templates are not specific enough to allow a single addition of the buttons. So we add them in the content templates.

Adding it on base_mobilitain.html was unsuccessful, as the buttons would show in undesired places (like the dashboard)

### DIFF
--- a/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
+++ b/transport_nantes/asso_tn/static/asso_tn/mobilitains.css
@@ -638,3 +638,8 @@ span.centered {
     transform: scale(1.1);
     z-index:3;
 }
+#share-buttons a:visited,
+#share-buttons a:link {
+    color: inherit;
+    text-decoration: none !important;
+}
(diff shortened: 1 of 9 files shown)


Shriukan33:
Django CI shows a fail, but the error message is : 
```
ERROR: test_vote_anon (photo.tests.TestVotes)
sqlite3.OperationalError: database table is locked: photo_vote
```

Locally these tests pass OK no problem, while also using sqlite3
Also, this PR doesn't change the backend code for `Vote`, so I'm a bit confused...

JeffAbrahamson:
> Partagez cette page sur les réseaux sociaux !

"This is a door!"

Why not simply one of these before the `fa-twitter` and `fa-facebook`?

```
fa-share
fa-share-square-o
fa-share-alt-square
fa-share-alt
fa-share-square
```

Shriukan33:
> Why not simply one of these before the `fa-twitter` and `fa-facebook`?

Before as in place of the text?

JeffAbrahamson:
> > Why not simply one of these before the `fa-twitter` and `fa-facebook`?
> 
> Before as in place of the text?

Like this (for example, first attempt): `[share] [twit] [fb]`

PR #1070, opened 2022-12-13
Make mobilito text mobilitains-gris

### LINKED ISSUE #1014: mobilto : white text gray
The text from the template is white but should be mobilitain-gray.

Shriukan33:
Contrast checker says it's ok, as the ratio is 6.17:1 The WCAG AAA requirement is 7:1, so we're close but not there yet.

Closes #1014


JeffAbrahamson (approved):
LGTM.

PR #1071, opened 2022-12-13
Benja add mobilito content graphs to email 1061

### LINKED ISSUE #1061: Add mobilito content graphs to email that we send
Even though the two return images with which we're not entirely happy, we should get the graphs in the email nonetheless (and improve the graphs after).

Shriukan33:
This PR updates the mobilito session's email template to match the color used in the app, and also adds the graphs in the email (so they aren't web-only) 

Closes #1061


JeffAbrahamson (approved):
LGTM.

PR #1075, opened 2022-12-14
Move arrows on the same level as the image in the photo gallery

Shriukan33:
The initial design intended to make the arrows slightly overflowing the image, but on lower resolutions this rendered poorly, making images even smaller.

Instead, I embedded entirely the arrows.
The arrows do not jump visually anymore and are steadily placed.

image

image

Closes #1072

PR #1076, opened 2022-12-14
Benja remove downvote button 1073

### LINKED ISSUE #1073: Remove downvote and make the like look like "Heart" button
The thumbs down button would draw attention to a negative feeling, instead of expressing those, we would like to only leave love or indifference as choices.

Shriukan33:
Based on PR  #1075

Remove dislike button, and replace like with heart

The dislike button was drawing attention to negative feelings
instead of expressing those, we would like to only leave love
or indifference as choices.

Clicking again on the heart is equivalent to what used to be
the dislike button.

image

Closes #1073


Shriukan33:
@JeffAbrahamson Now that arrows are merged, there is this one (previously draft because based on arrows)

Issue #1077, opened 2022-12-15
Debug CI : test_vote_anon (photo.tests.TestVotes) fails


Shriukan33:
There is probably a race condition in this test that make it fail regularly for no obvious reason, the test pass 100% locally. Time to track what's up.

PR #1078, opened 2022-12-15
Benja debug ci 1077

Shriukan33:
This is based on PR #1076 because this has to do with like button, so to save me a merge conflict, I did base it on latest change to that part of the code. 

Fix abusive database access in photo tests

The lambda function used to poll the database every 0.5s for 5 to
check if the vote had been processed. This was causing too many
database queries that would sometimes raise table locked errors
on the CI server.

This commit fixes it by checking the feedback that comes after the
vote has been processed : once the vote processed, the server
returns a 200 response and the "success" callback is executed.
This function will change the class list of the vote button by
toggling several classes.
It goes like this :
- Default state, the user never liked the photo. Class is absent.
- User likes the photo, css classes are added
- User unlikes the photo, css classes are removed <-- this is where
we polled the database, and where we now check the class list.

Closes [#1077](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

### DIFF
--- a/transport_nantes/photo/tests.py
+++ b/transport_nantes/photo/tests.py
@@ -291,6 +291,33 @@ class TestVotes(StaticLiveServerTestCase):
         self.anon_browser.implicitly_wait(5)
         self.auth_browser.implicitly_wait(5)
 
+    # This class is an adaptation from Selenium's documentation:
+    # https://selenium-python.readthedocs.io/waits.html#explicit-waits
+    # This has been implemented in place of a polling on the database
+    # because the polling was raising occasionally a table locked error
+    # on the CI server. (#1077)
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
Heh, that's called a spin loop, and you almost always want to avoid them, because they want to maim and kill you. ;-)

Shriukan33:
@JeffAbrahamson This one fixes the often failing CI test we got these last days !

JeffAbrahamson (approved):
LGTM, except: the documentation from the issue about what you're doing and the bug you're fixing would go quite well as comments in the test, too.

Issue #1022, opened 2022-11-30
Permit mobilito users to modify their location data


JeffAbrahamson:
For a period of, say, a week after creating a session, permit a user to modify the location field.  Do this the way github does for issue subjects.

Shriukan33:
Maybe we could do this from the session summary, with a button?

JeffAbrahamson:
> Maybe we could do this from the session summary, with a button?

Sorry, had missed a key word in the subject line, indeed.

Shriukan33:
Question, what's the goal behind the time limit? Why not allowing older than a week sessions to be updated?

Shriukan33:
> > Question, what's the goal behind the time limit? Why not allowing older than a week sessions to be updated?
> 
> Maybe it's silly. I'm inspired by sites like SO that don't let you change comments after too much time has gone by (there: 5 minutes), which is different. But if you come back a month later and propose to change your location data, do you _really_ likely remember the details?

It's an easy to remove constraint anyway, we can let that be for now.

PR #1079, opened 2022-12-15
Allow users to edit their MobilitoSession's location field

Shriukan33:
This view allows users to edit their mobilito sessions' string location field using an edit button.

The edit button is only visible to the user who created the mobilito session, or the persons with the permission to edit mobilito sessions.

Part of #1022

### DIFF
--- a/transport_nantes/mobilito/forms.py
+++ b/transport_nantes/mobilito/forms.py
@@ -1,5 +1,7 @@
 from django import forms
 
+from mobilito.models import MobilitoSession
+
 
 class AddressForm(forms.Form):
     location = forms.CharField(label='Localisation', max_length=1000,
@@ -8,3 +10,9 @@ class AddressForm(forms.Form):
                                  required=False)
     latitude = forms.FloatField(label='Latitude', widget=forms.HiddenInput(),
                                 required=False)
+
+
+class LocationEditForm(forms.ModelForm):
+    class Meta:
+        model = MobilitoSession
+        fields = ['location']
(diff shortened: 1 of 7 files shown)


JeffAbrahamson:
If I can't edit the text, then perhaps don't invite me to?
image

JeffAbrahamson:
What does this message add to the simple fact that the message has, visibly, changed to whatever I just changed it to?
image

Shriukan33:
I liked the positive feedback for the confirmation message :)

I admit I probably shouldn't have let the button be for older sessions, the only plus is can think of is that you can know why you can edit a session or not, as not not explained anywhere the 1 week rule

Envoyé à partir de Outlook pour Android<https://aka.ms/AAb9ysg>

JeffAbrahamson:
Aside from the messaging, it looks good.

I'm finding the user experience a bit jumpy.

Look at what happens on gh:
image
image

But for us we have lots of space (at least, in this particular case) and yet we reduce the width and crowd in some buttons and then it looks like
image
image

with, by the way, an errant and no longer correct message, because I've now started modifying the text and nothing has been saved yet
image

JeffAbrahamson:
Hey, look, I hit cancel, and yet:
image

JeffAbrahamson:
> I liked the positive feedback for the confirmation message :) I admit I probably shouldn't have let the button be for older sessions, the only plus is can think of is that you can know why you can edit a session or not, as not not explained anywhere the 1 week rule

Does anything still do this?  It feels very 1998 to me...

It's generally better to seek a good user experience than to make a clear experience that's wrong. ;-)  Inciting people to take actions that you have no intention of letting them complete is rarely, rarely the right thing to do...

Shriukan33:
* Removed messaging
* Removed button if past a week
* Changed Modifier to Sauvgarder
* Text input now takes all available space

JeffAbrahamson (approved):
LGTM.

PR #1085, opened 2022-12-21
Save lat/lng in form's field before reverse geocoding

### LINKED ISSUE #1084: Fill lat/lng fields when using GPS on mobilito

Shriukan33:
If the reverse geocoding failed, we would not have the lat/lng in the form's fields, while we had access to this information. By setting the field's values before the reverse geocoding, we can be sure to have the lat/lng even if the reverse geocoding failed.

Closes #1084


JeffAbrahamson (approved):
LGTM.

PR #1090, opened 2022-12-25
Photo 1089

### LINKED ISSUE #1089: The user-visible comment fields in photo/ are broken by design
* They are not available in admin, so are not directly editable.
* No test confirms that they display at all, let alone display the correct content or that the content is rendered from markdown.

JeffAbrahamson:
Closes #1089 and #1092.


Shriukan33:
LGTM aside that PEP8 recommends naming constants with ALL_CAPS (https://peps.python.org/pep-0008/#constants)

You seem to use `k_the_const` more often than not. Is it the C++ way?

Shriukan33 (approved):

JeffAbrahamson:
> LGTM aside that PEP8 recommends naming constants with ALL_CAPS (https://peps.python.org/pep-0008/#constants)
> 
> You seem to use `k_the_const` more often than not. Is it the C++ way?

I think I'm confusing with some element of the google style guide from years ago when I had to deal with that.  I'll fix the constants in a new PR in a bit.

Thanks for reminding me!

JeffAbrahamson:
In C++, I'd have used camel-case: `kSomeConstant`.

Issue #1062, opened 2022-12-06
Add mobilito map in summary emails


Shriukan33:
I did some research about how to go from lat/lng tuple to rendered png, there are services that provide that but it looks quite complex and long for that purpose.

https://print.get-map.org/about/api/
https://github.com/hholzgra/ocitysmap/

Or maybe having a tile render server of our own
https://github.com/enzet/map-machine#tile-server

There is no easy way if we want to add that into emails, I think.
Maybe we should encourage clicking to discover more info about the session, instead of trying to render it before sending. That would be much simpler implementation wise, not sure this is worth the hassle.

What do you think ?

JeffAbrahamson:
* [SO / tilemill](https://stackoverflow.com/questions/15623509/python-what-is-the-best-way-of-creating-a-png-of-a-map-with-a-route-for-a-writt)
* [ipyleaflet](https://stackoverflow.com/questions/44800396/python-ipyleaflet-export-map-as-png-or-jpg-or-svg)
* [py-staticmaps](https://pypi.org/project/py-staticmaps/)

PR #1103, opened 2022-12-30
Add "my sessions" page for mobilito

### LINKED ISSUE #1101: Add a "my sessions" page for Mobilito
If you want to find all your sessions, you have to keep the email, or chances are you're never going to find the URL again.

Let's say you want to show someone the session you did the other day, you have no easy way to do so.

wdyt?

Shriukan33:
This is an easy way to find your previous sessions. You had to go through your mails to find the link to your session, that wasn't very convenient.

Closes #1101

### DIFF
--- a/transport_nantes/mobilito/templates/mobilito/index.html
+++ b/transport_nantes/mobilito/templates/mobilito/index.html
@@ -38,13 +38,27 @@
 			href="{% url 'mobilito:tutorial' 'presentation' %}">
 				Je passe vite par le tutoriel
 			</a>
+			{% if has_sessions_history %}
+				<a class="btn donation-button mx-2" role="button" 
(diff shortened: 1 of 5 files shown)


JeffAbrahamson:
It's...not very inviting (here, on beta).  Though, with maps and a bit of extra textual clues about what happened, you start to look like the strava feed here.

image

Shriukan33:
Thing with maps is that we don't always have lat / lng, user isn't forced to provide them (you can type the address manually)
Also, it's not a static map, but a whole OSM map that is loaded every time. So unless we figure out a way ( #1062 ), I think that instantiating up to 10 maps per page would be a mess.

Could see to add the pie chart though?

JeffAbrahamson (approved):
LGTM.

We'll improve the product, but let's merge the code.

PR #1106, opened 2023-01-01
Blacken all python files

### LINKED ISSUE #1097: Use black globally
We should format the codebase, so it's more consistent.

JeffAbrahamson:
Note that this took less than 5 seconds: `$ black --line-length 79 .`.  ;-)

Closes #1097.


JeffAbrahamson:
@Shriukan33 CodeQual is complaining about something existent, so I'm not concerned per se.  But we should surely document this.  Who receives this exception and what do they do with it?  Why do we throw the stringified exception?  What would happen if we did something else?  Was this recommended somewhere?

JeffAbrahamson:
Tested with ce4e19a (contains a single long line), the workflow duly blocks merge approval.

Shriukan33:
LGTM! 

Cool that we are enforcing this.

Shriukan33 (approved):

PR #1123, opened 2023-01-03
Remove the alert and place marker even without reverse geocoding

### LINKED ISSUE #1113: In mobilito, clicking/touching the map doesn't position the marker

### LINKED ISSUE #1114: Remove Mobilito's error message when Reverse geocoding doesn't return an address
The current message is inaccurate, instead of displaying an alert, simply don't fill the field.
Also place the marker, even if Google doesn't give us an address.

Shriukan33:
The marker now moves as the user clicks on the map. It shows the address if the reverse geocoding is successful.

Closes #1114
Closes #1113


JeffAbrahamson (approved):
LGTM.

PR #1125, opened 2023-01-04
Add non-mandatory settings for ngrok

### LINKED ISSUE #1124: Allow ngrok in dev

Shriukan33:
The getattr allows to have a default value if the setting is not even defined in the settings_local.py file, which allows us to even not have to define on infra.

This feature is for dev env only and meant to ease the use of ngrok for local development.
Allowing ngrok subdomains in CSRF_TRUSTED_ORIGIN allows to use POST requests from ngrok subdomains (notably to login)

Because this is locked to a dev environment only, this doesn't raise any security issue on production.

closes #1124

### DIFF
--- a/transport_nantes/transport_nantes/settings.py
+++ b/transport_nantes/transport_nantes/settings.py
@@ -29,6 +29,21 @@ ADMIN = [
 ]
 
 ALLOWED_HOSTS = settings_local.ALLOWED_HOSTS
+# This is a list of trusted origins for CSRF protection.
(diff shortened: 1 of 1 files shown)


JeffAbrahamson:
Nowhere do you explain why we want to do this.

Shriukan33:
I mentioned it is to use ngrok.

Why would one to use ngrok ? To open a tunnel and make accessible your local dev branch through a browser, notably to make sure it displays well and works on my phone before even pushing to beta.

It's incredibly easy to use, and allows you to try out changes that might imply migrations changes without the hustle of restoring a dump on beta. 

https://ngrok.com/

PR #1126, opened 2023-01-04
Benja mobilito session feed 1115

### LINKED ISSUE #1115: viewing a mobilito session permits viewing user's sessions
When displaying a session, we display the mobilito user name (add to `MobilitoUser`, default to `MobilitoUser.user.username`, that is, gibberish).

Clicking the name shows all of the user's sessions.

Shriukan33:
This PR adds :

* A publicly accessible history for any given user using the sha1_identifier field (`/mobilito/session/u/<sha1_identifier>`)
* A username display under the location. This username isn't editable just yet, that is to come in later commits.

Hotspots for this PR : 
* I changed responsive a bit to fit a 3rd button to access your own history.
* I added a "Contributions récentes" feed on mobilito's index, tell me what you think. It's the 10 latest contributions.
* Username display: Pay closer attention to how username is fetched, I implemented a @property and a setter on this field.

Closes #1115

### DIFF
--- a/transport_nantes/mobilito/templates/mobilito/index.html
+++ b/transport_nantes/mobilito/templates/mobilito/index.html
@@ -71,6 +71,17 @@
 		</a>
 	</div>
 	{% endif %}{# ################ end ################ #}
+
+	<section class="mt-5">
+		<h1 style="font-size: 1.4em; text-align: left;">Les dernières contributions</h1>
+		<div class="d-flex flex-column col-12 p-0">
+			{% for session_obj in latest_sessions %}
+				{% include "mobilito/session_history_line.html" with session=session_obj %}
+			{% empty %}
+				<p>Aucune session enregistrée</p>
+			{% endfor %}
+		</div>
+	</section>
 </main>
 
 {% endblock app_content %}
--- a/transport_nantes/mobilito/templates/mobilito/index.html
+++ b/transport_nantes/mobilito/templates/mobilito/index.html
@@ -29,18 +29,18 @@
 		</div>
 	</div>
 	{% if tutorial_done %}{# ################ not newbie ################ #}
-		<div class="d-flex col-12 flex-row flex-nowrap justify-content-around">
-			<a class="btn donation-button mx-2" role="button" 
+		<div class="d-flex col-12 p-0 flex-wrap justify-content-around">
+			<a class="col-12 col-md-3 btn donation-button m-2" role="button" 
 			href="{% url 'mobilito:address_form' %}">
 				Je commence à enregistrer
 			</a>
-			<a class="btn donation-button mx-2" role="button"
+			<a class="col-12 col-md-3 btn donation-button m-2" role="button"
 			href="{% url 'mobilito:tutorial' 'presentation' %}">
 				Je passe vite par le tutoriel
 			</a>
 			{% if has_sessions_history %}
-				<a class="btn donation-button mx-2" role="button" 
-				href="{% url 'mobilito:my_sessions' %}">
+				<a class="col-12 col-md-3 btn donation-button m-2" role="button" 
+				href="{% url 'mobilito:user_sessions' mobilito_user.user.username %}">
 					Voir mon historique
 				</a>
 			{% endif %}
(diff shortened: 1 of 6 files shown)


transport_nantes/mobilito/models.py
```diff
@@ -22,6 +35,37 @@ class MobilitoUser(models.Model):
     def __str__(self):
         return self.user.email
 
+    @property
+    def username(self):
+        return self._username or self.user.username
```
JeffAbrahamson, line 40: There's really no way anyone coming at this code could possibly know why you are doing this.  Functions need documentation.

I don't think we wat to do this.  `User.username` is always set to a random string, but it would be quite counter-intuitive, if it's some day used for something, that it has a side-effect on publicly visible mobilito data.

Just initially generate a random string, as we do for user.  If the user wants to change it, they can.

transport_nantes/mobilito/models.py
```diff
@@ -13,6 +13,19 @@ class MobilitoUser(models.Model):
     """Maintain information about Mobilito users."""
 
     user = models.ForeignKey(User, on_delete=models.PROTECT, related_name="+")
+    # null or user defined
+    _username = models.CharField(
+        max_length=255,
+        blank=True,
+        null=True,
+        unique=True,
+        verbose_name="Nom d'utilisateur",
```
JeffAbrahamson, line 22: nom d'utilisateur mobilito

transport_nantes/mobilito/models.py
```diff
@@ -13,6 +13,19 @@ class MobilitoUser(models.Model):
...
+    )
+    # A sha1 identifier to display in URLs
+    sha1_identifier = models.CharField(
```
JeffAbrahamson, line 25: Somehow we got in the habit of using SHA1 to generate unique random keys rather than just generating random strings.  I think we should get ourselves out of this habit (and rename the fields to something more logical, like "human_identifier".

First, there's no benefit to sha1 here over just a random string.

Second, the fact that it's called sha1 suggests that we've actually hashed something known to arrive at the value, and that's not at all what we're doing.  We never start with a value to get the hash to find something, in the sense of a hash table.  So we are presenting misleading information in the code.

Finally, since SHA1 is known to suffer from collision attacks, a reader could well wonder why we haven't migrated to a safer hash algorithm, like SHA256 or SHA512 (or at least SHA1DC if we were bound, git-style, to SHA1-style hashes).  The reason we don't care is that we aren't using the fact that it's a cryptographic hash function, and that's likely quite confusing.

JeffAbrahamson (changes requested):
Just some minor changes, see comments.

transport_nantes/mobilito/models.py
JeffAbrahamson: Another reason not to do this is that your integrity check is against `self._username` but you are displaying `self._username or self.user.username`.  So you maintain relational integrity and column uniqueness, but you don't maintain display uniqueness.

I don't remember what our reasoning (if any) was for having a different Mobilito name than the global name.  Perhaps it would be more useful just to use the global name and be done with it.

transport_nantes/mobilito/models.py
Shriukan33: Adding "mobilito" could lead the users to question what other usernames they have, wouldn't it ?

transport_nantes/mobilito/models.py
Shriukan33: Also the mobilito username is something you asked for in the original issue : https://github.com/transport-nantes/tn_web/issues/1115

I believed it was to separate mobilito from the rest ?

So what should we be doing with the username ? Only display the username in mobilito model (then we'd have no "Fait par X" string until the user set a username) ? Or we use and edit the one in User ?

transport_nantes/mobilito/models.py
Shriukan33: Understood. Do you want me to redo migrations and use django.crypto to generate a random string instead in this PR ?

JeffAbrahamson: No, the values are fine.  I'm thinking we just accidentally got in a bad habit and should correct the field names and the method of generation.

From there, it doesn't matter that some are not pseudo-random.  They are unique strings with no clear sense to them.  That's enough.

Shriukan33:
As we discussed here and over the phone, I dropped the addition of MobilitoUser.username field.

No sha1 generation here. 

The result is simpler !

JeffAbrahamson:
Is this again ready for review?

JeffAbrahamson (approved):
LGTM.

PR #1127, opened 2023-01-06
Benja edit username 1117

### LINKED ISSUE #1117: Add mobilito user preference page
Allow user to modify her/his `MobilitoUser.username`.

Shriukan33:
Based on PR #1126 

This PR adds the preference page that allows users to change their usernames, as long as it's available.

Closes #1117

### DIFF
--- a/transport_nantes/asso_tn/forms.py
+++ b/transport_nantes/asso_tn/forms.py
@@ -1,0 +1,8 @@
+from django import forms
+from django.contrib.auth.models import User
+
+
+class UsernameEditForm(forms.ModelForm):
+    class Meta:
+        model = User
+        fields = ["username"]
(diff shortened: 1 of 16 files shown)


JeffAbrahamson (approved):
LGTM.

Shriukan33:
Rebased and updated to reflect the changes made since we dropped the MobilitoUser.username field :)

JeffAbrahamson:
On 7aabe668b, if this is not a mobilito-specific feature, why is it a mobilito-specific function?

Shriukan33:
The form isn't especially Mobilito-specific, but it's true it's being accessed through Mobilito history.

How would you prefer it to be implemented then ? Something not branded mobilito ?

I thought it was ok as for now, the only place where there is even a notion of username is Mobilito, it's also the only place where it's displayed explicitely.

Shriukan33:
I added the preference page (so we can expand it later if needed), and the possibility to edit your username without leaving the page, same as Strava.

JeffAbrahamson:
Looks good, one minor quibble:
image
* The pencil and the name don't have space between them but do different things, so should.
* The pencil isn't obviously clickable (no mouse feedback), so I try to click the name, which is, but that does something else (which is logical, it just wasn't what the pencil says).

I think if you separate them slightly horizontally and give some feedback that the pencil can be clicked, we'll be good!

Shriukan33:
Made the pencil white and turned the cursor to pointer on hover, and added some left margin on the pencil so we should be good ! Also rebased because it was based on other branches that have been updated.

PR #1131, opened 2023-01-11
Benja add map png 1062

Shriukan33:
This PR is based on #1126 

I added the PNG generation of a map using a Mobilito Session's coordinates (lat,lng) and [MapBox](https://docs.mapbox.com/api/maps/static-images/).

I replaced the occurrences where we were initializing a leaflet map with the PNG visualization : In mobilito landing, emails, and history. 

In the last commit I suggested a new layout for the history and mobilito landing that adds more texture and life, in my opinion : 

image

image

image


JeffAbrahamson:
Perhaps 9ec057325a should be squashed into c693fd8d5 ?

JeffAbrahamson:
In dab7f40373 you are assuming that we always successfully generated maps.  But, as you note when contacting mapbox, it could fail.

So perhaps we should say "if I have an image, display it; if I don't and I have lat/lon, then fetch an image and display it".

JeffAbrahamson (approved):
LGTM.

I think we should (in a future PR) handle the case that we never successfully got a map.

I also think we will have discussions about the last commit (about number of columns), but we'll have to play with to see what we like, what works.

Issue #1144, opened 2023-02-05
Fix TBPanel


JeffAbrahamson:
TBPanel was never properly integrated, so it existed without any actual instances.  The result is that on creating an instance, we discover that it doesn't actually work.

This is perhaps not surprising, because nothing could ever have tested that it did...

So make it work and add those tests.

PR #1145, opened 2023-02-05
Fix TBPanel so that it works

JeffAbrahamson:
Part of #1144.


transport_nantes/topicblog/models.py
```diff
@@ -1678,3 +1680,30 @@ def get_edit_url(self):
                 "topicblog:edit_panel",
                 kwargs={"pkid": self.pk, "the_slug": self.slug},
             )
+
+
+def get_random_panel():
```
Shriukan33, line 1685: This would fit quite well in the Class directly, in my opinion

transport_nantes/topicblog/models.py
```diff
@@ -1678,3 +1680,30 @@ def get_edit_url(self):
...
+    #     .annotate(num_published=Count("slug"))
+    #     .count()
+    # )
+    published_panels = (
+        TopicBlogPanel.objects.filter(publication_date__isnull=False)
+        .values("slug")
+        .annotate(pub_date=Max("publication_date"))
+    )
+    the_panel_values = published_panels[
+        randint(0, published_panels.count() - 1)
```
Shriukan33, line 1704: I think Count() is actually making another query https://docs.djangoproject.com/en/4.1/ref/models/querysets/#django.db.models.query.QuerySet.count , and because you will evaluate the query's object anyway, using len() might be even more efficient, but performance isn't even an issue at this point. :)

transport_nantes/topicblog/templates/topicblog/panel_did_you_know_tip_1_mail_client.html
```diff
@@ -1,5 +1,4 @@
 {% load markdown %}
-<meta name="viewport" content="width=device-width,initial-scale=1.0">
```
Shriukan33, line 2: I wasn't even sure about what it was about, maybe the responsive aspect of the mail ? The media queries will need the width, so maybe it's used to get proper values regarding the width to perform responsive, but that's just a guess really

Shriukan33 (approved):
About 3447a8ef7cf3d27fae5fce306dfdf4551a6b4a31 I did write tests here : https://github.com/transport-nantes/tn_web/blob/55206f7e8b8681eb5db87aa22a025ae361d3234e/transport_nantes/topicblog/templatetags/test_templatags.py#L159 but the more the better I would say ! 

Otherwise LGTM, provided you tried sending emails, I scanned the code :)

JeffAbrahamson:
> 

Testing the template tag and testing the markdown filter are different things!  Indeed, I discovered the missing test because the template tag filter was failing because of an issue on the markdown side (and so untested by the templatetag filter).

transport_nantes/topicblog/models.py
JeffAbrahamson: If it were in the class, I'd expect it to manipulate an individual member object.  But instead it's doing a query.  It wouldn't every make sense to call

```
my_panel.get_random_panel()
```

Shriukan33: That's a pattern I've seen, but whatever sits right to you, it doesn't really matter :)

transport_nantes/topicblog/models.py
JeffAbrahamson: Yeah, this is contorted.  In this case, the dereference (`[randint]`) will cause us only to fetch one record, so I don't think we actually fully retrieve the queryset.

PR #1166, opened 2023-05-14
Mail templates

JeffAbrahamson:
This all seems pretty straight-forward, fixing formatting and mail presentation, mostly.

Creating a PR to make sure lint runs (tests pass locally).  Comments welcome.  I'll self-approve and merge Monday afternoon barring feedback.

### DIFF
--- a/transport_nantes/asso_tn/templates/asso_tn/base_email_mobilitain.html
+++ b/transport_nantes/asso_tn/templates/asso_tn/base_email_mobilitain.html
@@ -90,33 +90,47 @@
             text-decoration: none;
         }
     </style>
-    <script src="https://kit.fontawesome.com/46b82563d9.js" crossorigin="anonymous"></script>
 </head>
-<body style="margin:0;padding:0;word-spacing:normal;background-color:#ffffff;">
-    <div role="article" aria-roledescription="email" lang="fr"
-    style="text-size-adjust:100%;-webkit-text-size-adjust:100%;-ms-text-size-adjust:100%;background-color:#ffffff;">
-        {% comment %}
-        This scaffold is necessary so that our email will be centered across all email clients.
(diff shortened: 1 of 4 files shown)


JeffAbrahamson:
> Nothing jumped to me, except that I find it weird that linter prefers the old %s notation compared to f strings. LGTM

I believe there are two reasons.  One is efficiency: f-strings are interpolated before the log function is called (and the log function may decided not to log), whereas `%` arguments are interpolated only once the log function decides to log.  (Why?  I'm not 100% sure.)

[python.org/3/](https://docs.python.org/3/howto/logging.html#optimization)
[r/learnpython](https://www.reddit.com/r/learnpython/comments/xx14jw/why_not_use_fstrings_in_logging/)

That said, perhaps `.format()` would be better, but I was just doing what lint asked.

I also read somewhere, but can't find today, that f-strings permit arbitrary computation and so are potentially dangerous when formatting user-supplied data.  [Little Bobby Tables](https://xkcd.com/327/) territory.  I can't find that reference, aside from [this issue](https://github.com/python/cpython/issues/90358).

---

ChatGPT to the rescue (sort of, because I can't duplicate, but it's the essence of something I read about this).

Consider this code block:

```
import logging

def process_user_input(user_input):
    # Arbitrary computation or validation
    result = user_input.upper()
    return result

user_input = input("Enter your name: ")

# Logging statement with interpolated user input
logging.warning(f"User {process_user_input(user_input)} logged in.")
```

Now suppose the user provides the name `{__import__('os').system('rm -rf /')}`.

I tested this with `user_input = "__import__('os').system('touch foo')"`, however, and I can't repeat.  So perhaps this was once a bug and it's been fixed.  Or the blog I read before holiday was simply incorrect.

